PR #6, opened 2022-08-18
feat: run test/clippy on PR/push and build image into registry on push

arein:
# Testing

Sample run https://github.com/WalletConnect/rpc-proxy/actions/runs/2885300088


Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+#
+# Generate recipe file
+#
+################################################################################
+FROM                chef AS plan
+
+WORKDIR             /app
+COPY                Cargo.lock Cargo.toml ./
+COPY                src ./src
+RUN                 cargo chef prepare --recipe-path recipe.json
```
heilhead, line 34: Something I've explored in the past is using docker's [mount-type cache](https://docs.docker.com/engine/reference/builder/#run---mounttypecache). While `cargo chef` is great, for me it was just too slow as it had to cook the recipe entirely on any dependency change. And it was also hard to share the cooked cache between multiple docker images built from the same code base (e.g. multiple binaries within one crate/repository).

So long story short, mount type cache solved all of the issues and sped up the build immensely.

Here's a snippet:
```
RUN --mount=type=cache,target=/usr/local/cargo/registry \
  --mount=type=cache,target=/usr/src/app/target \
  cargo build --release --package=<package-name> && \
  cp target/release/<package-name> /usr/local/bin
```
(`/usr/src/app/target` is the path to your crate's `target` cache dir)

Your mileage may vary obviously, but something to keep in mind for the future.

heilhead (approved):
Looks good!

src/main.rs
```diff
@@ -28,6 +29,7 @@ async fn main() -> error::Result<()> {
     let state = state::new_state(config);
 
     let port = state.config.port;
+    let host = state.config.host.clone();
```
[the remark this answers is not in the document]
arein: If i do so I get the following error on line 35:

```
use of partially moved value: `state`
partial move occurs because `state.config.host` has type `std::string::String`, which does not implement the `Copy` 
```

Dockerfile
arein: Right now I only build the Docker image in CI not locally. I think I'd like to revisit this in the future and proceed as is

PR #8, opened 2022-08-20
feat: exposes app via rpc.walletconnect.com

arein:
# Testing

`curl -vvv -X POST -H "Content-Type:application/json" "https://dev.rpc.walletconnect.com/v1/?chainId=eth:1&projectId=someid" -d '{"id":"1660887896683","jsonrpc":"2.0","method":"eth_chainId","params":[]}'`

### DIFF
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -168,7 +168,7 @@ resource "aws_alb" "network_load_balancer" {
 
 resource "aws_lb_target_group" "target_group" {
   name        = replace("${var.app_name}-${substr(uuid(), 0, 3)}", "_", "-")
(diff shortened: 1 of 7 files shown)


sbc64 (approved):
This looks great :+1: :+1:

HarryET (changes requested):
Are we only having a prod instance of this? not a staging and dev...

arein:
> Are we only having a prod instance of this? not a staging and dev...

Staging and dev are also already set up

`terraform -chdir=terraform workspace select staging && terraform -chdir=terraform apply`

arein:
@HarryET merging as it seems your only concern here was addressed

PR #9, opened 2022-08-22
feat: supports chains other than mainnet

arein:
Supports all chains that Infura currently supports expected to be specified as CAIP2 e.g. eip155:1.

# Testing

`curl -vvv -X POST -H "Content-Type:application/json" "http://localhost:3000/v1?chainId=eip155:5&projectId=someid" -d '{"id":"1660887896683","jsonrpc":"2.0","method":"eth_chainId","params":[]}'`

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -12,17 +12,6 @@ pub async fn handler(
     headers: hyper::http::HeaderMap,
     body: hyper::body::Bytes,
 ) -> Result<impl warp::Reply, warp::Rejection> {
-    if query_params.chain_id != "eth:1" {
-        return Ok(new_error_response(
-            vec![ErrorReason {
-                field: "chainId".to_string(),
-                description: "We currently only support `eth:1`".to_string(),
-            }],
-            StatusCode::BAD_REQUEST,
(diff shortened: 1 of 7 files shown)


IljaDaderko (approved):

src/handlers/proxy.rs
```diff
@@ -12,29 +12,30 @@ pub async fn handler(
...
+                field: "projectId".to_string(),
+                description: "No project id required".to_string(),
             }],
             StatusCode::BAD_REQUEST,
         )
         .into_response());
     }
 
-    if query_params.project_id.is_empty() {
+    let provider = provider_repo.get_provider("eth").unwrap();
```
heilhead, line 26: Is it safe to unwrap here? Since the function already returns a `Result`, maybe use something like `.ok_or(...)?` or `.ok_or_else(...)?` to avoid unwrapping?

heilhead (approved):

src/handlers/proxy.rs
arein: Maybe an `expect` but this shouldn't be possible to happen - also not adding this code here just rearranging.

PR #13, opened 2022-09-02
feat: install metrics and export via prometheus

arein:
Installs a call counter that let's us know how many calls we received split up by chain id.

Next step: add a prometheus instance and create a Grafana dashboard.

# Testing
Tested the `/metrics/` endpoint locally.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,9 +22,6 @@ use warp::Filter;
 use hyper::Client;
 use hyper_tls::HttpsConnector;
 
-// Re-export opentelemetry API.
-pub use opentelemetry::*;
-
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -20,13 +22,33 @@ use warp::Filter;
 use hyper::Client;
 use hyper_tls::HttpsConnector;
 
+// Re-export opentelemetry API.
+pub use opentelemetry::*;
```
heilhead, line 26: Is this a copy-paste? I don't think you can export anything from `main.rs`.

heilhead (approved):
Looks good!

PR #20, opened 2022-10-05
feat: add support for Solana/Avax

arein:
# Testing

Tested using `WalletConnect/solib`

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -11,8 +11,11 @@ pub struct Config {
     #[serde(default = "default_log_level")]
     pub log_level: String,
     pub infura_project_id: String,
+    pub pokt_project_id: String,
     #[serde(default = "default_infura_supported_chains")]
     pub infura_supported_chains: HashMap<String, String>,
+    #[serde(default = "default_pokt_supported_chains")]
+    pub pokt_supported_chains: HashMap<String, String>,
 }
 
 fn default_port() -> u16 {
@@ -46,6 +49,15 @@ fn default_infura_supported_chains() -> HashMap<String, String> {
         ("eip155:42161".into(), "arbitrum-mainnet".into()),
         ("eip155:421611".into(), "arbitrum-rinkeby".into()),
         ("eip155:421613".into(), "arbitrum-goerli".into()),
+    ])
+}
+
+fn default_pokt_supported_chains() -> HashMap<String, String> {
+    HashMap::from([
+        // Solana Mainnet
+        ("solana:4sgjmw1sunhzsxgspuhpqldx6wiyjntz".into(), "solana-mainnet".into()),
+        // Avax C-Chain
+        ("eip155:43114".into(), "avax-mainnet".into()),
     ])
 }
 
--- a/src/env.rs
+++ b/src/env.rs
@@ -55,7 +55,10 @@ fn default_infura_supported_chains() -> HashMap<String, String> {
 fn default_pokt_supported_chains() -> HashMap<String, String> {
     HashMap::from([
         // Solana Mainnet
-        ("solana:4sgjmw1sunhzsxgspuhpqldx6wiyjntz".into(), "solana-mainnet".into()),
+        (
+            "solana:4sgjmw1sunhzsxgspuhpqldx6wiyjntz".into(),
+            "solana-mainnet".into(),
+        ),
         // Avax C-Chain
         ("eip155:43114".into(), "avax-mainnet".into()),
     ])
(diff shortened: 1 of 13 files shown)


src/providers/pokt.rs
```diff
@@ -0,0 +1,51 @@
...
+        _headers: hyper::http::HeaderMap,
+        body: hyper::body::Bytes,
+    ) -> Result<Response<Body>, Error> {
+        let full_path = path.as_str().to_string();
+        let mut hyper_request = hyper::http::Request::builder()
+            .method(method)
+            .uri(full_path)
+            .header("Content-Type", "application/json")
+            .body(hyper::body::Body::from(body))
+            .expect("Request::builder() failed");
```
heilhead, line 31: The `proxy()` function already returns `Result<_, _>`, so it makes sense to replace all of these `expect()`s with `?` error propagation.

src/providers/pokt.rs
```diff
@@ -0,0 +1,51 @@
...
+        method: hyper::http::Method,
+        path: warp::path::FullPath,
+        query_params: RPCQueryParams,
+        _headers: hyper::http::HeaderMap,
+        body: hyper::body::Bytes,
+    ) -> Result<Response<Body>, Error> {
+        let full_path = path.as_str().to_string();
+        let mut hyper_request = hyper::http::Request::builder()
+            .method(method)
+            .uri(full_path)
```
heilhead, line 28: Why set the URI here, if you replace it below?

src/providers/pokt.rs
```diff
@@ -0,0 +1,51 @@
...
+            .body(hyper::body::Body::from(body))
+            .expect("Request::builder() failed");
+
+        let chain = self
+            .supported_chains
+            .get(&query_params.chain_id.to_lowercase())
+            .expect("Chain not found despite previous validation");
+
+        *hyper_request.uri_mut() =
+            format!("https://{}.gateway.pokt.network/v1/lb/{}", chain, self.project_id)
```
heilhead, line 39: I'm not entirely sure how this proxying works - does it not use the `full_path` in any way?

src/providers/mod.rs
```diff
@@ -15,8 +17,9 @@ pub struct ProviderRepository {
 }
 
 impl ProviderRepository {
-    pub fn get_provider(&self, chain: &str) -> Option<&Arc<dyn RPCProvider>> {
-        self.map.get(chain)
+    pub fn get_provider_for_chain_id(&self, chain: &str) -> Option<&Arc<dyn RPCProvider>> {
+        self.map.values().into_iter()
```
heilhead, line 21: I feel like iterating over a hash map like this kind of defeats the purpose of using the hash map in the first place.

Question: If the key into this map is the `chain` param that we're using here, why do you need to ensure that the provider actually supports this chain ID? Since providers here are wrapped in `Arc`, we can freely clone them and insert as many times as we need, i.e. for each of the chain IDs that they support.

src/handlers/proxy.rs
```diff
@@ -26,32 +26,36 @@ pub async fn handler(
...
+                )],
+            );
+        
+            // TODO: map the response error codes properly
+            // e.g. HTTP401 from target should map to HTTP500
+            let resp = provider
+                .proxy(method, path, query_params, headers, body)
+                .await;
+        
+            resp.map_err(|_e| warp::reject::reject())
```
heilhead, line 58:
```suggestion
            resp.map_err(|_| warp::reject::reject())
```

heilhead (approved):
Looks good! Left some minor feedback to address if you have time.

src/env.rs
```diff
@@ -49,6 +52,15 @@ fn default_infura_supported_chains() -> HashMap<String, String> {
     ])
 }
 
+fn default_pokt_supported_chains() -> HashMap<String, String> {
+    HashMap::from([
+        // Solana Mainnet
+        ("solana:4sgjmw1sunhzsxgspuhpqldx6wiyjntz".into(), "solana-mainnet".into()),
+        // Avax C-Chain
+        ("eip155:43114".into(), "avax-mainnet".into()),
```
IljaDaderko, line 60: Sick, but we can probably add support for rest of pokt supported chains here perhaps? i.e. why stop at just avax

arein: Yes. What else do they have that I missed?

src/providers/pokt.rs
arein: I like the custom err messages though

src/env.rs
IljaDaderko: Just looked at the docs, we can add `evmos`, `binance` and `gnosis`

PR #35, opened 2022-11-02
feat: Add project gating

### LINKED ISSUE #33: Add project id validation to the rpc-proxy service

xav:
# Description

Add project gating with optional caching.
Resolves #33 

## How Has This Been Tested?

WIP

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,17 +17,22 @@ tap = "1.0"
 
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+serde_piecewise_default = "0.2"
 
 dotenv = "0.15.0"
 envy = "0.4"
 
 build-info = "0.0.27"
 
-prometheus-core = { package = "prometheus", version = "0.13" }
+anyhow = "1"
 opentelemetry = { version = "0.17.0", features = ["trace", "metrics", "rt-tokio"] }
 opentelemetry-prometheus = "0.10.0"
-tracing = "0.1"
+prometheus-core = { package = "prometheus", version = "0.13" }
 thiserror = "1.0"
+tracing = "0.1"
+
+common = { git = "ssh://git@github.com/WalletConnect/rs-common.git", tag = "v0.1.1" }
+cerberus = { git = "ssh://git@github.com/WalletConnect/cerberus.git", tag = "v0.1.2" }
 
 [build-dependencies]
 build-info-build = "0.0.27"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,8 @@ prometheus-core = { package = "prometheus", version = "0.13" }
 thiserror = "1.0"
 tracing = "0.1"
 
-common = { git = "ssh://git@github.com/WalletConnect/rs-common.git", tag = "v0.1.1" }
-cerberus = { git = "ssh://git@github.com/WalletConnect/cerberus.git", tag = "v0.1.2" }
+common = { git = "https://github.com/WalletConnect/rs-common.git", tag = "v0.1.1" }
+cerberus = { git = "https://github.com/WalletConnect/cerberus.git", tag = "v0.1.2" }
 
 [build-dependencies]
 build-info-build = "0.0.27"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,16 +8,23 @@ authors = [
 build = "build.rs"
 
 [dependencies]
+# Async
 tokio = { version = "1", features = ["full"] }
 async-trait = "0.1.57"
+
 warp = "0.3"
 hyper = "0.14.4"
 hyper-tls = "0.5.0"
 tap = "1.0"
 
+# Serialization
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_piecewise_default = "0.2"
+rmp-serde = "1.1"
+
+# Storage
+deadpool-redis = "0.10"
 
 dotenv = "0.15.0"
 envy = "0.4"
@@ -31,7 +38,6 @@ prometheus-core = { package = "prometheus", version = "0.13" }
 thiserror = "1.0"
 tracing = "0.1"
 
-common = { git = "https://github.com/WalletConnect/rs-common.git", tag = "v0.1.1" }
 cerberus = { git = "https://github.com/WalletConnect/cerberus.git", tag = "v0.1.2" }
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ envy = "0.4"
 
 build-info = "0.0.27"
 
-anyhow = "1"
 opentelemetry = { version = "0.17.0", features = ["trace", "metrics", "rt-tokio"] }
 opentelemetry-prometheus = "0.10.0"
 prometheus-core = { package = "prometheus", version = "0.13" }
--- a/src/env.rs
+++ b/src/env.rs
@@ -51,11 +51,11 @@ pub struct Config {
 impl Config {
     pub fn from_env() -> error::RpcResult<Config> {
         Ok(Self {
-            server: from_env("")?,
-            infura: from_env("INFURA_")?,
-            pokt: from_env("POKT_")?,
-            registry: from_env("REGISTRY_")?,
-            storage: from_env("STORAGE_")?,
+            server: from_env("RPC_PROXY_")?,
+            infura: from_env("RPC_PROXY_INFURA_")?,
+            pokt: from_env("RPC_PROXY_POKT_")?,
+            registry: from_env("RPC_PROXY_REGISTRY_")?,
+            storage: from_env("RPC_PROXY_STORAGE_")?,
         })
     }
 }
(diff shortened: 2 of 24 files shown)


arein (approved):
LGTM! Please wait for @heilhead to also take a look

xav:
> LGTM! Please wait for @heilhead to also take a look

Will do. I want to deploy the redis cluster first anyway to avoid running without cache.

src/env.rs
```diff
@@ -1,21 +1,67 @@
...
+    pub infura: InfuraConfig,
+    pub pokt: PoktConfig,
+    pub registry: RegistryConfig,
+    pub storage: StorageConfig,
+}
+
+impl Config {
+    pub fn from_env() -> error::RpcResult<Config> {
+        Ok(Self {
+            server: from_env("")?,
```
heilhead, line 54: Maybe it makes sense to have global prefix like `RPC_PROXY_` to avoid any potential collisions with other software running on the same machine? In relay we prefix env values with `RELAY_`.

src/handlers/proxy.rs
```diff
@@ -1,14 +1,20 @@
...
 use tracing::warn;
 
+use crate::handlers::{handshake_error, RpcQueryParams};
+use crate::project::Registry;
+use crate::providers::ProviderRepository;
+use crate::State;
+
+use super::field_validation_error;
+
+#[allow(clippy::too_many_arguments)]
```
heilhead, line 13: Why not add registry as part of the global state?

src/handlers/proxy.rs
```diff
@@ -22,6 +28,18 @@ pub async fn handler(
         ));
     }
 
+    match registry.project_data(&query_params.project_id).await {
+        Ok(project) => {
+            if let Err(access_err) = project.validate_access(&query_params.project_id, None) {
+                return Ok(handshake_error("projectId", format!("{access_err}")));
```
heilhead, line 34: Hopefully one day we'll refactor both the relay and RPC proxy to use a proper HTTP server, and we'll be able to actually return `Err(err)` instead of `Ok(err)`...

Cargo.toml
```diff
@@ -17,17 +17,22 @@ tap = "1.0"
...
-prometheus-core = { package = "prometheus", version = "0.13" }
+anyhow = "1"
 opentelemetry = { version = "0.17.0", features = ["trace", "metrics", "rt-tokio"] }
 opentelemetry-prometheus = "0.10.0"
-tracing = "0.1"
+prometheus-core = { package = "prometheus", version = "0.13" }
 thiserror = "1.0"
+tracing = "0.1"
+
+common = { git = "https://github.com/WalletConnect/rs-common.git", tag = "v0.1.1" }
```
heilhead, line 34: Are we planning to use `common` in any other projects as well?

src/project/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        let api_url = &cfg_registry.api_url;
+        let api_auth_token = &cfg_registry.api_auth_token;
+
+        let client = if let (Some(api_url), Some(api_auth_token)) = (api_url, api_auth_token) {
+            RegistryHttpClient::new(api_url, api_auth_token)?
+        } else {
+            return Err(RpcError::InvalidConfiguration(
+                "missing registry api parameters".to_string(),
+            ));
+        };
```
heilhead, line 47: There's a very neat new syntax feature [available in 1.65](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements):
```suggestion
        let (Some(api_url), Some(api_auth_token)) = (api_url, api_auth_token) else {
            return Err(RpcError::InvalidConfiguration(
                "missing registry api parameters".to_string(),
            ));
        };

        let client = RegistryHttpClient::new(api_url, api_auth_token)?;
```

src/project/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+            Some(cache_addr) => {
+                let cache = open_redis(&cache_addr, cfg_storage.redis_max_connections)?;
+
+                Some(ProjectStorage::new(
+                    cache,
+                    cfg_registry.project_data_cache_ttl(),
+                    metrics.clone(),
+                ))
+            }
+        };
```
heilhead, line 63: Not sure if this is better, but:
```suggestion
        let cache = cfg_storage
            .project_data_redis_addr()
            .map(|addr| open_redis(&addr, cfg_storage.redis_max_connections))
            .transpose()?
            .map(|conn| {
                ProjectStorage::new(conn, cfg_registry.project_data_cache_ttl(), metrics.clone())
            });
```

heilhead (approved):
Looks great, thanks!

One open question still is `rs-common`. I know, we've been discussing this a lot, but my arguments were:
1. That code wasn't that good/stable to deserve its own crate;
2. Are there any other projects which are going to use it? If not, it's just two copies of the same code (in `relay` and `rs-common`), so might as well move it to `rpc-proxy` instead.

Anyway, good work!

src/handlers/proxy.rs
xav: Yeah, it would be nice. That felt a bit strange :)

Cargo.toml
xav: Yes, eventually in `relay`.

xav: You think it's a bad idea?

src/project/mod.rs
xav: It looks fantastic, but I think the original communicates intent better, and I'm not sure there's a difference performance-wise.

Cargo.toml
heilhead: I don't think that code is good enough/stable to deserve its own crate. As discussed before, I think it should have remained internal to the relay.

src/metrics.rs
```diff
@@ -0,0 +1,82 @@
...
+            ],
+        )
+    }
+
+    pub fn add_rejected_project(&self, project_id: &str) {
+        self.rejected_project_counter.add(
+            1,
+            &[opentelemetry::KeyValue::new(
+                "project_id",
+                project_id.to_owned(),
```
heilhead, line 78: This may get littered with a bunch of invalid project IDs, if someone decides to bruteforce or something. If `project_id` is not that useful for the metrics, I think it should be removed.

src/project/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+impl Registry {
+    pub fn new(
+        cfg_registry: &Config,
+        cfg_storage: &StorageConfig,
+        meter: &Meter,
+    ) -> RpcResult<Self> {
+        let api_url = &cfg_registry.api_url;
+        let api_auth_token = &cfg_registry.api_auth_token;
+
+        let (Some(api_url), Some(api_auth_token)) = (api_url, api_auth_token) else {
```
heilhead, line 46: Nice! Make sure the CI is set up with 1.65/latest rust.

src/storage/error.rs
```diff
@@ -0,0 +1,23 @@
...
+    SetExpiry,
+    /// Unable to serialize data to store
+    #[error("error on serialize data")]
+    Serialize,
+    /// Unable to deserialize data from store
+    #[error("error on deserialize data")]
+    Deserialize,
+    /// Error on establishing a connection with the storage
+    #[error("error on open connection")]
+    Connection(anyhow::Error),
```
heilhead, line 19: Since this is the only place where `anyhow` is used, I'd suggest changing this to something like `Connection(Box<dyn std::error::Error>)` or `Connection(String)` and ditch `anyhow` entirely. Same for `Other(...)` below.

src/storage/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+
+pub mod error;
+pub mod redis;
+
+/// The Result type returned by Storage functions
+pub type StorageResult<T> = Result<T, StorageError>;
+
+#[async_trait]
+pub trait SetStorage<T>: 'static + Send + Sync + Debug
```
heilhead, line 17: `SetStorage` isn't used anywhere, so can safely remove it for faster compilation.

src/storage/redis/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::storage::{
+    deserialize, serialize, Data, KeyValueStorage, SetStorage, StorageError, StorageResult,
+};
+
+const LOCAL_REDIS_ADDR: &str = "redis://localhost:6379/0";
+
+#[derive(Debug, Clone)]
+pub enum Addr<'a> {
```
heilhead, line 20: This code was actually the reason why I wasn't happy with it being extracted to `rs-common`. Was trying to micro-optimize for now good reason, and eventually decided to leave it like that. Let me know if you have ideas how to improve the `Addr`, or redis API entirely.

heilhead (approved):
Looks great! Left some bikeshedding as always :)

src/storage/redis/mod.rs
xav: Nope, can't come up with anything worth the change, especially if we eventually intend to ditch `redis`.

PR #37, opened 2022-11-04
feat(o11y): add http metrics

arein:
Currently we can't distinguish the response codes of the proxy hence we cannot do availability.

# Testing

Not tested


xav (approved):
LGTM

PR #41, opened 2022-11-09
fix: bad index declaration for private hosted zone

xav:


arein (approved):
Please going forward actually apply your Terraform changes to dev or staging so you have confidence everything works.

xav:
> 

I did, but the failing resource is only declared in prod.
We should modify stage to bring it as close as possible to a 1:1 with prod.

PR #46, opened 2022-11-17
fix: remove wait for steady state on ecs updates

xav:
# Description

Remove the `wait_for_steady_state` for ECS deployments.
It seems that the deployment takes too long, thus sending Terraform in a timeout error. This should work fine as long as we don't send updates too often and keep an eye on the cluster.
We need to update the `aws` provider so that we can provide [custom timeouts](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#timeouts) (not supported in this version).

## How Has This Been Tested?

Deployed by hand to staging

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -160,8 +160,9 @@ resource "aws_ecs_service" "app_service" {
   launch_type     = "FARGATE"
   desired_count   = 2 # Setting the number of containers we want deployed to 3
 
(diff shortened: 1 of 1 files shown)


arein (changes requested):
A deployment should only take ~5mins if the boxes get healthy. We don't have this issue in any other pipeline.

We cannot turn this off:
* we might promote a version to prod that is broken
* we will need to add integration tests and they may only run when the deployment is actually done

PR #55, opened 2022-11-25
feat: add analytics

### LINKED ISSUE #23: Export Analytics to Athena
In `rs-relay` we have logic to "dump" certain data (socket/messages) into parquet files. Based on certain thresholds e.g. time/file size we upload these files to S3. From there on our Athena pipeline kicks (in `data-lake` repo).

Subject of this task is only the file creation and upload part.

xav:
# Description

Export the following request attributes to parquets on S3:
- method (eg `eth_call`)
- network (eg eth `mainnet`)
- host (the dapp URL)
- continent/country
- projectid
- which RPC provider served the call (in the file path)

Request status is not exported since it would require us to buffer the responses instead of streaming them back.

Resolves #23 

## How Has This Been Tested?

Run locally and deployed to dev.

## Due Diligence

* [ ] Breaking change


Cargo.toml
```diff
@@ -9,36 +9,49 @@ build = "build.rs"
...
+rmp-serde = "1.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_piecewise_default = "0.2"
-rmp-serde = "1.1"
+serde-aux = "3.1"
 
 # Storage
+aws-config = "0.47"
+aws-sdk-s3 = "0.17.0"
```
heilhead, line 29:
```suggestion
aws-sdk-s3 = "0.17"
```

src/analytics/message_info.rs
```diff
@@ -0,0 +1,44 @@
...
+        query_params: &RpcQueryParams,
+        request: &JsonRpcRequest,
+        sender: Option<SocketAddr>,
+        country: Option<Arc<str>>,
+        continent: Option<Arc<str>>,
+    ) -> Self {
+        Self {
+            timestamp: super::create_timestamp(),
+
+            project_id: Arc::from(query_params.project_id.to_owned()),
```
heilhead, line 34: It doesn't make sense to use `Arc<str>` where the original data is `String`. Especially for conversions like these - it causes double allocation (first one by `.to_owned()`, and then to convert `String` -> `Arc<str>`). Let's change `project_id`, `chain_id` and `sender` to be `String`-based.

src/analytics/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.messages.collect(data);
+    }
+
+    pub fn lookup_geo_data(&self, addr: IpAddr) -> Option<AnalyticsGeoData> {
+        self.geoip.lookup_geo_data(addr)
+    }
+}
+
+fn create_timestamp() -> String {
+    chrono::Utc::now().format("%Y-%m-%d %H:%M:%S").to_string()
```
heilhead, line 96: I feel like should also be a part of `gorgon` crate. We're using the same analytics backend, which will likely parse data from all our services in the same way, so maybe it makes sense to share such common format utils.

src/env/server.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+// #############################################################################
+
+fn default_port() -> u16 {
+    3000
+}
+
+fn default_host() -> String {
+    "127.0.0.1".to_string()
```
heilhead, line 33: Minor, but I would use `DeserializePiecewiseDefault` + `Default`, since you already have it as a dependency.

heilhead (approved):
Looks great! Can you also open a PR in relay to remove the analytics code in favor of `gorgon`?

justfile
```diff
@@ -0,0 +1,138 @@
+binary-crate            := "."
```
arein, line 1: Nice!

src/env/infura.rs
```diff
@@ -0,0 +1,33 @@
+use std::collections::HashMap;
```
arein, line 1: nice

arein (approved):
Very nice! Can you confirm that this works with Solana, too?

[Sample query](https://walletconnect.slack.com/archives/CB247BW9F/p1664974281950189)

PR #58, opened 2022-12-05
Chore/monitor rate limits

### LINKED ISSUE #32: Monitor for Rate Limits on providers
e.g. Infura sends HTTP429 when throttled.

Rakowskiii:
# Description

Implemented Display for Providers, made it a required trait for Provider.
On each proxied request to provider, on successful response, check for HTTP 429 and store such occurences in metrics with provider info (provider and project id).
Implements functionality requested in issue #32  

Resolves #32 

## How Has This Been Tested?

Untested for now. As I don't want to force rate limit for either infura or pokt, I'll try to intercept the traffic and force-feed the proxy with HTTP 429 to test it.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -2,6 +2,7 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 
 use crate::analytics::MessageInfo;
+use hyper::http;
 use tap::TapFallible;
 use tracing::warn;
 
@@ -74,5 +75,10 @@ pub async fn handler(
         .proxy(method, path, query_params, headers, body)
         .await
         .tap_err(|error| warn!(%error, "request failed"))
+        .tap_ok(|response| {
+            if response.status() == http::StatusCode::TOO_MANY_REQUESTS {
+                state.metrics.add_rate_limited_call(&provider.to_string())
+            }
+        })
         .map_err(|_| warp::reject::reject())
 }
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -30,6 +30,11 @@ impl Metrics {
             .with_description("The number of calls for invalid project ids")
             .init();
 
+        let rate_limited_counter = meter
+            .u64_counter("rate_limited_counter")
+            .with_description("The number of calls that got rate limited")
+            .init();
+
         Metrics {
             rpc_call_counter,
             http_call_counter,
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -6,6 +6,7 @@ pub struct Metrics {
     pub http_call_counter: Counter<u64>,
     pub http_latency_tracker: Counter<f64>,
     pub rejected_project_counter: Counter<u64>,
+    pub rate_limited_call_counter: Counter<u64>,
 }
 
 impl Metrics {
@@ -30,7 +31,7 @@ impl Metrics {
             .with_description("The number of calls for invalid project ids")
             .init();
 
-        let rate_limited_counter = meter
+        let rate_limited_call_counter = meter
             .u64_counter("rate_limited_counter")
             .with_description("The number of calls that got rate limited")
             .init();
@@ -40,6 +41,7 @@ impl Metrics {
             http_call_counter,
             http_latency_tracker,
             rejected_project_counter,
+            rate_limited_call_counter,
         }
     }
 }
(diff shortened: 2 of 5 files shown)


src/handlers/proxy.rs
```diff
@@ -74,5 +75,10 @@ pub async fn handler(
         .proxy(method, path, query_params, headers, body)
         .await
         .tap_err(|error| warn!(%error, "request failed"))
+        .tap_ok(|response| {
+            if response.status() == http::StatusCode::TOO_MANY_REQUESTS {
```
heilhead, line 79: Are all of them using 429 code? Need to make sure, otherwise this logic will need to be moved to the `RpcProvider` trait.

src/metrics.rs
```diff
@@ -30,11 +31,17 @@ impl Metrics {
             .with_description("The number of calls for invalid project ids")
             .init();
 
+        let rate_limited_call_counter = meter
+            .u64_counter("rate_limited_counter")
```
heilhead, line 35: I'm not sure if a `u64` counter here is the right thing here. @arein What's our plan to monitor here - set up an alert in grafana so that when this counter goes up it'd notify us?

src/metrics.rs
```diff
@@ -73,4 +80,14 @@ impl Metrics {
     pub fn add_rejected_project(&self) {
         self.rejected_project_counter.add(1, &[])
     }
+
+    pub fn add_rate_limited_call(&self, provider: &str) {
```
heilhead, line 84: I would instead pass the provider here (as `&dyn RpcProvider` or similar), and use the provider to generate two key/value pairs: `provider_kind` (e.g. `infura`, `pokt` etc.) and `project_id`. This way it'd be easier to find/filter in grafana.

Also, with current implementation you're passing a `&str` which is immediately cloned (`.to_owned()`), while discarding the original. Could just pass `String` here to avoid cloning. But again, I'd go with separate tags for provider kind and project id.

src/handlers/proxy.rs
Rakowskiii: https://docs.infura.io/infura/networks/ethereum/how-to/avoid-rate-limiting
> When you're rate limited, your JSON-RPC responses have HTTP status code 429.

As for Pokt:
I couldn't find anything in docs, waiting for a response in Pokt discord.

src/metrics.rs
Rakowskiii: Done

src/providers/mod.rs
```diff
@@ -43,4 +62,10 @@ pub trait RpcProvider: Send + Sync {
     fn supports_caip_chainid(&self, chain_id: &str) -> bool;
 
     fn supported_caip_chainids(&self) -> Vec<String>;
+
+    fn provider_kind(&self) -> ProviderKind;
+
+    fn project_id(&self) -> String;
```
heilhead, line 69: Minor: it's best to return `&str` and let the consumer decide if they want to clone it, or just use the reference.

src/handlers/proxy.rs
```diff
@@ -68,11 +70,35 @@ pub async fn handler(
...
+                state
+                    .metrics
+                    .add_rate_limited_call(provider.borrow(), project_id)
+            };
+            Ok(response)
+        }
+    }
+}
+
+async fn copy_body_bytes(response: Response<Body>) -> (Bytes, Response<Body>) {
```
Rakowskiii, line 98: I'm not sure about that part, but hyper doesn't allow cloning of the Response, and in order to read jsonrpc response code from the body the ownership needs to be moved. This function clones the body after reading is finished and rebuilds the original response form it.

src/handlers/proxy.rs
Rakowskiii: Confirmed that with adequate resources

src/providers/mod.rs
Rakowskiii: Done

heilhead (approved):

src/metrics.rs
arein: Usually I would classify this as a an error in Grafana and do metric math to alarm.
Unfortunately this doesn't work in Grafana.

So we will likely need to do the metric math in the application.

It's a problem for all HTTP apps that needs this. So it's a question for all apps. Maybe there's a better solution

PR #65, opened 2022-12-12
feat: add native timestamp to analytics

xav:
# Description

Add native timestamps to analytics parquet files in a new bucket, keep legacy string timestamps in the current bucket

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/analytics/message_info.rs
+++ b/src/analytics/message_info.rs
@@ -7,17 +7,69 @@ use std::sync::Arc;
 
 #[derive(Debug, Clone, Serialize, ParquetRecordWriter)]
 #[serde(rename_all = "camelCase")]
+pub struct LegacyMessageInfo {
+    pub timestamp: String,
+
+    pub project_id: String,
+    pub chain_id: String,
+    pub method: Arc<str>,
+
+    pub sender: Option<String>,
+
+    pub country: Option<Arc<str>>,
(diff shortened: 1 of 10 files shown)


src/analytics/message_info.rs
```diff
@@ -5,19 +5,71 @@ use serde::Serialize;
...
+            continent: value.continent.clone(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, ParquetRecordWriter)]
 #[serde(rename_all = "camelCase")]
 pub struct MessageInfo {
-    timestamp: String,
+    pub timestamp: chrono::NaiveDateTime,
```
heilhead, line 63: Why is this not `chrono::DateTime<chrono::Utc>`, and why not use `chrono::Utc::now()` instead of `gorgon::time::now()`?

heilhead (approved):

src/analytics/message_info.rs
xav: Because `NaiveDateTime` is the only one supported by the parquet crate macro :(
I've been through several iterations to try and make it better, short of writing an implementation for a regular timestamp, but the macro won't take anything other than that (not even a type alias).
Now, a naive timestamp is just a timestamp without TZ info; since we only use UTC that should be too much of a problem for us.

PR #68, opened 2022-12-13
Added initial test suite. Pinging each provider for all possible chains.

### LINKED ISSUE #21: Implement Proper Integ Tests
Tests should make a random call using all chains we support ideally

Implementation suggestion:
I think the tests in https://github.com/WalletConnect/registry-api-cf-workers/tree/master/integration are a good example

Rakowskiii:
# Description

Initial test suite, testing connection to all providers, fetching metadata from all available chains.
Also checking if the error is returned for chains that are no longer supported.

Resolves #21 

## How Has This Been Tested?

cargo test (with proper env variables)

## Due Diligence

* [ ] Breaking change
* [X] Requires a e2e/integration test update


tests/functional/connection/infura.rs
```diff
@@ -0,0 +1,395 @@
...
+    };
+
+    let (status, rpc_response) = send_jsonrpc_request(client, addr, "eip155:3", request).await;
+
+    // Verify that HTTP communication returns error
+    assert_eq!(status, StatusCode::GONE);
+
+    // Verify the error code is for
+    // "Network decommissioned, please use Goerli or Sepolia instead"
+    assert_eq!(rpc_response.error.unwrap().code, -32601);
```
xav, line 94: Even if it's only tests, could you use local consts for all those to avoid magic numbers

src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+mod handlers;
+mod json_rpc;
+mod metrics;
+mod project;
+mod providers;
+mod state;
+mod storage;
+mod utils;
+
+pub async fn bootstrap(mut shutdown: broadcast::Receiver<()>, config: Config) -> RpcResult<()> {
```
xav, line 31: Nice!

tests/utils/utils.rs
```diff
@@ -0,0 +1,27 @@
+use hyper::{body, client::HttpConnector, Body, Client, Method, Request, StatusCode};
+use hyper_tls::HttpsConnector;
+
+pub async fn send_jsonrpc_request(
```
xav, line 4: This can be directly in `/utils/mod.rs` to avoid `utils::utils:...`

Cargo.toml
```diff
@@ -12,6 +12,8 @@ build = "build.rs"
 async-trait = "0.1.57"
 tokio = { version = "1", features = ["full"] }
 
+# Web
+jsonrpc = "0.14.0"
```
xav, line 16: If we introduce this crate, either we should use it consistently everywhere instead of our own version, or it should be in `dev-dependencies` if it's only used for tests.

xav (approved):
This is awesome!
You just need to add a project ID to the CI too (tests are failing) :slightly_smiling_face:

heilhead (approved):

PR #75, opened 2022-12-16
fix: activate client ip forwarding in load balancer

xav:
# Description

Activate client IP forwarding in the network load balancer.

## How Has This Been Tested?

Deployed on staging.

## Due Diligence

* [ ] Breaking change


arein (approved):
I assume this is for country resolution

PR #84, opened 2022-12-20
feat(o11y): install error alarms

arein:
## How Has This Been Tested?

Applied from local.

## Due Diligence

* [ ] Breaking change


arein:
Monitoring only change so no review required per team policy

PR #86, opened 2022-12-23
feat: gate prod with integ tests on staging url

### LINKED ISSUE #81: Gate Prod promotions with integration tests
https://github.com/WalletConnect/rpc-proxy/blob/master/.github/workflows/cd.yaml#L97-L107

We should replace this with the integration tests we have now. they should use the actual `staging` endpoint of the proxy.

Rakowskiii:
# Description

Runs integration tests on staging url to check if everything works after deployment

Resolves #81 

## How Has This Been Tested?

The integration tests are will be ran on staging endpoint before deploying to prod

## Due Diligence

* [ ] Breaking change
* [X] Requires a e2e/integration test update

### DIFF
--- a/tests/context/mod.rs
+++ b/tests/context/mod.rs
@@ -1,3 +1,5 @@
+use std::env;
+
 use async_trait::async_trait;
 use test_context::AsyncTestContext;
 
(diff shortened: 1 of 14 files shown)


tests/context/mod.rs
```diff
@@ -12,11 +14,25 @@ pub struct ServerContext {
...
 impl AsyncTestContext for ServerContext {
     async fn setup() -> Self {
+        #[cfg(feature = "test-localhost")]
         let server = RpcProxy::start().await;
+
+        #[cfg(not(feature = "test-localhost"))]
+        let server = {
+            let project_id = env::var("PROJECT_ID").expect("PROJECT_ID must be set");
+            RpcProxy {
+                public_addr: "https://staging.rpc.walletconnect.com".into(),
```
arein, line 24: ideally we can pass this url also as an environment variable

arein (approved):
Ideally you'd have a dedicated `validate` workflow same as we do in other repos.

Reference code:
* [shared validate workflow](https://github.com/WalletConnect/rs-relay/blob/master/.github/workflows/validate.yml) from `echo server`
* `echo server` [CD pipeline `call`ing](https://github.com/WalletConnect/echo-server/blob/main/.github/workflows/cd.yml#L65-L71) the shared workflow

tests/context/mod.rs
Rakowskiii: Sure, will change

PR #88, opened 2022-12-23
feat: added tracking of external latency

### LINKED ISSUE #52: Track integration latency
We should have a latency metric that subtracks the proxy time and only tracks the latency we are adding.

Rakowskiii:
# Description

Measures time spent on communication with external proxy. 
This with total latency will allow gathering information on both overall latency performance and the both our proxy and external providers performance, allowing insight into source of eventual latency. 

Resolves #52

## How Has This Been Tested?

The graph will be added to graphana dashboard and latency will be tracked

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -1,12 +1,13 @@
 use opentelemetry::metrics::{Counter, Meter};
 
-use crate::providers::RpcProvider;
+use crate::providers::{ProviderKind, RpcProvider};
 
 #[derive(Clone, Debug)]
 pub struct Metrics {
     pub rpc_call_counter: Counter<u64>,
     pub http_call_counter: Counter<u64>,
     pub http_latency_tracker: Counter<f64>,
+    pub http_external_latency_tracker: Counter<f64>,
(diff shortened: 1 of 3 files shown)


src/metrics.rs
```diff
@@ -1,12 +1,13 @@
...
 
-use crate::providers::RpcProvider;
+use crate::providers::{ProviderKind, RpcProvider};
 
 #[derive(Clone, Debug)]
 pub struct Metrics {
     pub rpc_call_counter: Counter<u64>,
     pub http_call_counter: Counter<u64>,
     pub http_latency_tracker: Counter<f64>,
+    pub http_external_latency_tracker: Counter<f64>,
```
heilhead, line 10: It should not be `Counter` though. Use `Histogram` like in [the relay](https://github.com/WalletConnect/rs-relay/blob/master/src/project/metrics.rs#L50).

heilhead (changes requested):

heilhead (approved):

PR #95, opened 2023-02-01
fix: removed unnecesary requirement for zksync project_id

Rakowskiii:
# Description

Solves recent problems with tests after zksync addition.
Renames PROJECT_ID to TEST_RPC_PROXY_PROJECT_ID to give it more context.
Mentions TEST_RPC_PROXY_PROJECT_ID in .env.example.

## How Has This Been Tested?

Integ tests ran locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/context/server.rs
+++ b/tests/context/server.rs
@@ -32,7 +32,8 @@ impl RpcProxy {
 
         let (signal, shutdown) = broadcast::channel(1);
 
-        let project_id = env::var("PROJECT_ID").expect("PROJECT_ID must be set");
+        let project_id =
+            env::var("TEST_RPC_PROXY_PROJECT_ID").expect("TEST_RPC_PROXY_PROJECT_ID must be set");
 
         std::thread::spawn(move || {
             rt.block_on(async move {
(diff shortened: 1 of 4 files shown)


tests/context/server.rs
```diff
@@ -32,7 +32,8 @@ impl RpcProxy {
 
         let (signal, shutdown) = broadcast::channel(1);
 
-        let project_id = env::var("PROJECT_ID").expect("PROJECT_ID must be set");
+        let project_id =
+            env::var("TEST_RPC_PROXY_PROJECT_ID").expect("TEST_RPC_PROXY_PROJECT_ID must be set");
```
arein, line 36: what id is this? What is an expected value and where do I get it from? :)

arein (approved):

tests/context/server.rs
Rakowskiii: This is a rpc proxy project id, from registry. This will only be needed for testing on staging (as it will need valid project_id when communicating with registry)

PR #104, opened 2023-03-21
feat: move from warp to axum

### LINKED ISSUE #103: feat: move from warp to axum
Migrate the project from Warp framework to Axum for better scalling and easier maintenance.

Rakowskiii:
# Description

Moving from warp to axum in preparation for future work.

Resolves #103 

## How Has This Been Tested?

All the tests were ran. 
Also, double checked by hand if errors thrown are the same and responses from different providers/chains match with current deployement.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,6 +467,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum"
+version = "0.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2fb79c228270dcf2426e74864cabc94babb5dbab01a4314e702d2f16540e1591"
+dependencies = [
+ "async-trait",
+ "axum-core",
+ "bitflags",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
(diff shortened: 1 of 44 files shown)


chris13524 (commented):
LGTM

farazdagi (approved):
Absolutely amazing work!

Disclaimer: I am not well-versed in `rpc-proxy` code, so might have missed some issues, but it looks great.

Note: Note sure how critical `solana_mainnet` test was, but if it is not critical, then probably we need to remove it (and have a GitHub issue opened, stating that we need to re-add it). Just commenting out should be used sparingly, imo.

Rakowskiii:
> Note: Note sure how critical solana_mainnet test was, but if it is not critical, then probably we need to remove it (and have a GitHub issue opened, stating that we need to re-add it). Just commenting out should be used sparingly, imo.

This was a temp thing, with problem on Pocket side, we resolved it, restored the test! But good point, in the future I'll just delete and restore later, thanks!

arein (approved):

PR #115, opened 2023-04-05
feat: add ws proxy functionality

### LINKED ISSUE #87: feat: add websocket support
Add websocket connection support to RPC proxy

Rakowskiii:
# Description

Adding handler for websocket proxying `/ws` 
Currently only supporting infura, as pokt doesn't seem to support websockets.
Zksync is deprecating ws https://docs.zksync.io/api/sdk/js/providers/#create-websocket-provider-it-is-deprecated-and-will-be-removed-soon

Resolves #87 

## How Has This Been Tested?

Tests ran for each connection.

## Due Diligence

* [ ] Breaking change
* [x] Requires a e2e/integration test update


src/error.rs
```diff
@@ -47,11 +48,19 @@ pub enum RpcError {
...
+    #[error("Invalid scheme used. Try http(s):// or ws(s)://")]
+    InvalidScheme,
+
+    #[error(transparent)]
+    AxumTungstenite(#[from] axum_tungstenite::Error),
 }
 
 impl IntoResponse for RpcError {
     fn into_response(self) -> axum::response::Response {
+        error!("{:?}", self);
```
xDarksome, line 61: Is this a recommended way to log errors in `axum`? I tend to consider logging inside conversion functions to be a bad practice.

Is there some other place we could `tap_err` or something?

src/extractors/protocol.rs
```diff
@@ -0,0 +1,33 @@
+use {
+    crate::error::RpcError,
+    async_trait::async_trait,
+    axum::{extract::FromRequestParts, http::request::Parts},
+};
+
+#[derive(Debug)]
+pub enum Protocol {
+    HTTP,
```
xDarksome, line 9: Reasonable warning here. Should be `Http`

src/providers/mod.rs
```diff
@@ -63,7 +78,24 @@ pub trait RpcProvider: Send + Sync {
         query_params: RpcQueryParams,
         headers: hyper::http::HeaderMap,
         body: hyper::body::Bytes,
-    ) -> RpcResult<Response<Body>>;
+    ) -> RpcResult<Response>;
+
+    fn supports_caip_chainid(&self, chain_id: &str) -> bool;
```
xDarksome, line 83: Consider extracting those fn's into a separate trait and making this trait a supertrait of `RpcProvider` and `RpcWsProvider`.

tests/functional/websocket/infura.rs
```diff
@@ -0,0 +1,411 @@
+use {
+    crate::{context::ServerContext, JSONRPC_VERSION},
+    futures_util::{SinkExt, StreamExt},
+    test_context::test_context,
+};
+
+#[test_context(ServerContext)]
+#[tokio::test]
+async fn eip155_1_mainnet_infura(ctx: &mut ServerContext) {
```
xDarksome, line 9: It's usually not a good idea to generalize tests, but these are doing almost exactly the same thing.  We have distinct "ok" and "gone" checks here. 

I would recommend writing two separate functions for those cases, parametrize them with a couple of required arguments and then use them inside all of the specific test cases.

src/error.rs
Rakowskiii: I don't think this is recommended way. 
My thought process behind this was: tapping in each place separately may lead to some uncovered spots in the future. Logging it here take off the responsibility from future contributors, as they don't need to keep remembering about each and every error they return.

tests/functional/websocket/infura.rs
Rakowskiii: Done.

src/extractors/protocol.rs
Rakowskiii: Done.

src/providers/mod.rs
```diff
@@ -55,16 +70,27 @@ impl Display for ProviderKind {
...
         xpath: axum::extract::MatchedPath,
         query_params: RpcQueryParams,
         headers: hyper::http::HeaderMap,
         body: hyper::body::Bytes,
-    ) -> RpcResult<Response<Body>>;
+    ) -> RpcResult<Response>;
+}
+
+#[async_trait]
+pub trait RpcWsProvider: Send + Sync + Provider {
```
xDarksome, line 85: Nitpick: `Send + Sync + 'static`  are conventionally being written at the end

src/error.rs
xDarksome: We should be able to do logging via some middleware abstraction of `axum`, like `layer` or something, but I don't have a better solution rn. So LGTM for now

xDarksome (approved):

src/error.rs
Rakowskiii: I'll drop this as issue, thanks

PR #119, opened 2023-04-28
fix: remove polygon ws - no longer supported by infura 

Rakowskiii:
# Description

Infura restricted ws functionality for polygon rendering it currently not useful for our proxy purposed. Removing it for now. We still support it over HTTP.

## How Has This Been Tested?

Failing tests lead to contact with infura and information from them.
Now tests pass.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/env/infura.rs
+++ b/src/env/infura.rs
@@ -27,6 +27,9 @@ fn default_infura_supported_chains() -> HashMap<String, String> {
         ("eip155:42161".into(), "arbitrum-mainnet".into()),
         ("eip155:421611".into(), "arbitrum-rinkeby".into()),
         ("eip155:421613".into(), "arbitrum-goerli".into()),
+        // Polygon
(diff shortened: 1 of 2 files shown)


src/env/infura.rs
```diff
@@ -19,9 +19,6 @@ fn default_infura_supported_chains() -> HashMap<String, String> {
         ("eip155:42".into(), "kovan".into()),
         ("eip155:4".into(), "rinkeby".into()),
         ("eip155:5".into(), "goerli".into()),
-        // Polygon
-        ("eip155:137".into(), "polygon-mainnet".into()),
```
HarryET, line 23: Should this have been lines 47/48, in the `default_infura_ws_supported_chains` function?

Rakowskiii: Yeah 🤦🏻‍♂️
Thanks for catching that

HarryET (approved):
Looks good now as long as tests are passing!

PR #123, opened 2023-05-15
feat: distinguish between provider and service errors

### LINKED ISSUE #122: add `provider` to tracked rpc_requests data
as we want to know about how much rpc_requests are routed to which provider ( infura,quicknode,etc) we need to add a new field to the tracked data that contains the provider.

ac:
- `rpc_requests` logs have a `provider` column with the id/name of the provider

Rakowskiii:
# Description

Added more metrics to track both how many calls are successful and how many fail per provider, as well as status codes per provider.

Also added `provider` to data lake exports.

Resolves #122 

## How Has This Been Tested?

This is initial work in the subject, currently tested by running locally and forcing two kinds of error: in the service and in the provider, with the following result

```
curl -X POST \
  'http:///localhost:3000/v1/?projectId=<project_id>3&chainId=eip155:2' \
  --header 'Content-Type: application/json' \
  --data-raw '{"id":"1","jsonrpc":"2.0","method":"eth_chainId","params":[]}'
```
where eip155:2 is not supported -> results in: 
```
400 BAD REQUEST
{
  "status": "FAILED",
  "reasons": [
    {
      "field": "chainId",
      "description": "We don't support the chainId you provided: eip155:2"
    }
  ]
}
```

while 
```
curl -X POST \
  'http:///localhost:3000/v1/?projectId=<project_id>&chainId=eip155:1' \
  --header 'Content-Type: application/json' \
  --data-raw '{ "invalid":"json" }'
```
returns 

```
502 BAD GATEWAY

{
  "jsonrpc": "2.0",
  "error": {
    "code": -32600,
    "message": "invalid json request"
  }
}
```

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -9,7 +9,7 @@ use {
     axum::{
         body::Bytes,
         extract::{ConnectInfo, MatchedPath, Query, State},
-        response::{IntoResponse, Response},
+        response::Response,
     },
     hyper::{http, HeaderMap},
     std::{
@@ -71,7 +71,7 @@ pub async fn handler(
     // Start timing external provider added time
     let external_call_start = SystemTime::now();
 
-    let response = provider
+    let mut response = provider
         .proxy(method, path, query_params, headers, body)
         .await
         .map_err(|error| {
@@ -92,22 +92,19 @@ pub async fn handler(
             .as_secs_f64(),
     );
 
-    let response = match response.status() {
+    match response.status() {
         http::StatusCode::OK => {
             state.metrics.add_finished_provider_call(provider.borrow());
-            response
         }
         status => {
             state.metrics.add_failed_provider_call(provider.borrow());
             state
                 .metrics
                 .add_status_code_for_provider(provider.borrow(), status);
-            let (mut parts, body) = response.into_parts();
-            parts.status = http::StatusCode::BAD_GATEWAY;
 
-            Response::from_parts(parts, body)
+            *response.status_mut() = http::StatusCode::BAD_GATEWAY;
         }
     };
 
-    Ok(response.into_response())
+    Ok(response)
 }
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -11,7 +11,7 @@ use {
         extract::{ConnectInfo, MatchedPath, Query, State},
         response::{IntoResponse, Response},
     },
-    hyper::HeaderMap,
+    hyper::{http, HeaderMap},
     std::{
         borrow::Borrow,
         net::SocketAddr,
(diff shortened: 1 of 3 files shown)


src/handlers/proxy.rs
```diff
@@ -91,5 +92,22 @@ pub async fn handler(
...
+            state.metrics.add_finished_provider_call(provider.borrow());
+            response
+        }
+        status => {
+            state.metrics.add_failed_provider_call(provider.borrow());
+            state
+                .metrics
+                .add_status_code_for_provider(provider.borrow(), status);
+            let (mut parts, body) = response.into_parts();
+            parts.status = http::StatusCode::BAD_GATEWAY;
```
xDarksome, line 106: Why are we setting `BAD_GATEWAY` here?

src/handlers/proxy.rs
```diff
@@ -91,5 +92,22 @@ pub async fn handler(
...
+
+            Response::from_parts(parts, body)
+        }
+    };
+
     Ok(response.into_response())
```
xDarksome, line 112: Seems like it's already a `Response`

src/handlers/proxy.rs
```diff
@@ -91,5 +92,22 @@ pub async fn handler(
...
+        }
+        status => {
+            state.metrics.add_failed_provider_call(provider.borrow());
+            state
+                .metrics
+                .add_status_code_for_provider(provider.borrow(), status);
+            let (mut parts, body) = response.into_parts();
+            parts.status = http::StatusCode::BAD_GATEWAY;
+
+            Response::from_parts(parts, body)
```
xDarksome, line 108: There is a [Response::status_mut](https://docs.rs/http/0.2.9/http/response/struct.Response.html#method.status_mut), so you should be able to do it without unpacking it into parts.

xDarksome (changes requested):

src/handlers/proxy.rs
Rakowskiii: We are acting as a proxy here, and we always expect 200 OK, as this is jsonrpc, if provider is up, and it would properly return error to user, it would still be 200 OK. 
That means that any other status code is unexpected from provider and treated as invalid.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502

src/handlers/proxy.rs
Rakowskiii: Thanks! Changed that

src/handlers/proxy.rs
Rakowskiii: Changed

xDarksome (approved):
:+1:

PR #125, opened 2023-05-17
feat: failover routing

### LINKED ISSUE #29: Support multiple providers per chainid
We will soon have multiple enterprise partners per chain id (Infura/Coinbase cloud). Naively let's just round-robin between those providers.

Rakowskiii:
# Description

Implementing failover, allowing for multiple providers to be defined;

Also aggregated tests together - seems like there was some race condition where github actions was trying to spawn so many async contexts. 

Resolves #29 

## How Has This Been Tested?

Ran tests, also ran locally with debugging, and called Binance multiple times, about 50% of them went to pokt (so according to setup)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/utils/mod.rs
+++ b/tests/utils/mod.rs
@@ -25,5 +25,6 @@ pub async fn send_jsonrpc_request(
 
     let (parts, body) = response.into_parts();
     let body = body::to_bytes(body).await.unwrap();
+    dbg!(&body);
     (parts.status, serde_json::from_slice(&body).unwrap())
 }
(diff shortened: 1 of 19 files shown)


tests/utils/mod.rs
```diff
@@ -25,5 +25,6 @@ pub async fn send_jsonrpc_request(
 
     let (parts, body) = response.into_parts();
     let body = body::to_bytes(body).await.unwrap();
+    dbg!(&body);
```
arein, line 28: Nit: remove?

PR #128, opened 2023-05-18
feat: add more providers

Rakowskiii:
# Description

Added two more providers that support our most used chains, so that failover will actually have targets to chose from.

## How Has This Been Tested?

Ran tests several times, also most crucial chains were verified by hand to make sure we get proper responses from all providers. Had to remove one potential provider (1rpc) as they were inconsistent in latency, sometimes reaching 3s.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/env/omnia.rs
+++ b/src/env/omnia.rs
@@ -2,19 +2,19 @@ use {crate::providers::Weight, serde::Deserialize, std::collections::HashMap};
 
 #[derive(Deserialize, Debug, Clone, PartialEq)]
 pub struct OmniatechConfig {
-    #[serde(default = "default_omniatech_supported_chains")]
+    #[serde(default = "default_supported_chains")]
     pub supported_chains: HashMap<String, (String, Weight)>,
 }
 
 impl Default for OmniatechConfig {
     fn default() -> Self {
         Self {
-            supported_chains: default_omniatech_supported_chains(),
+            supported_chains: default_supported_chains(),
         }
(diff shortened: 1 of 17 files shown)


src/env/omnia.rs
```diff
@@ -0,0 +1,36 @@
...
+
+impl Default for OmniatechConfig {
+    fn default() -> Self {
+        Self {
+            supported_chains: default_omniatech_supported_chains(),
+        }
+    }
+}
+
+fn default_omniatech_supported_chains() -> HashMap<String, (String, Weight)> {
```
xDarksome, line 17: No need to specify `omniatech` in the name as we are inside `omnia` module context.

src/providers/omnia.rs
```diff
@@ -0,0 +1,99 @@
...
+        if is_rate_limited(body_bytes) {
+            return Err(RpcError::Throttled);
+        }
+
+        Ok(response.into_response())
+    }
+}
+
+fn is_rate_limited(body_bytes: Bytes) -> bool {
+    let Ok(jsonrpc_response) = serde_json::from_slice::<jsonrpc::Response>(&body_bytes) else {return false};
```
xDarksome, line 79: Fmt would make it `{ return false }` if it worked :smile:

xDarksome (approved):
Potentially, consider generalizing those providers, as they have a lot of overlapping logic.

PR #132, opened 2023-05-19
fix: better front facing information and logging

### LINKED ISSUE #121: add `origin` to tracked rpc_requests data
We want to know where are `rpc_request` originating from so we want to add the `origin` (eg. opensea.io) to the data we send to the data lake.  This will enable us to have more oversight of RPC usage from the different projects or bad actors. 

AC:
- a new field is added to the `rpc_request` data containing origin when available

Rakowskiii:
# Description

Adding more informative metrics, and exporting origin, if exists to data lake.

Resolves #121 

## How Has This Been Tested?

Ran tests, before deploying to prod will tests with staging and confirm with data team that they are receiving expected data.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/binance.rs
+++ b/src/providers/binance.rs
@@ -67,5 +67,5 @@ impl RpcProvider for BinanceProvider {
 }
 
 fn is_rate_limited(response: &Response) -> bool {
-    response.status() == http::StatusCode::TOO_MANY_REQUESTS
+    response.status() == http::StatusCode::FORBIDDEN
 }
--- a/src/analytics/message_info.rs
+++ b/src/analytics/message_info.rs
@@ -14,6 +14,7 @@ pub struct MessageInfo {
     pub chain_id: String,
     pub method: Arc<str>,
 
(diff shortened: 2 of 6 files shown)


src/providers/binance.rs
```diff
@@ -67,5 +67,5 @@ impl RpcProvider for BinanceProvider {
 }
 
 fn is_rate_limited(response: &Response) -> bool {
-    response.status() == http::StatusCode::TOO_MANY_REQUESTS
+    response.status() == http::StatusCode::FORBIDDEN
```
xav, line 70: Why are we responding with `TOO_MANY_REQUESTS` for Omnia but changed to `FORBIDDEN` for Binance?

Rakowskiii: This is not a response we send, but rather a response we get from provider. Each provider signals in different way that they are rate limiting us. Omnia sends 429, but Binance once we go over the limit goes with 403

PR #140, opened 2023-05-31
fix: remove decomissioned chains

Rakowskiii:
# Description

I was investigating why the tests are failing so oftenly, I though it was because half-baked system. Turns out Infura changed their handling of one decomissioned chain (from proper error to forbidden).

I think it's good idea to get rid of decomissioned chains. This way instead of giving strange error to user, we just inform them we don't support this chain with our message.

## How Has This Been Tested?

Ran tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -151,29 +151,22 @@ pub async fn bootstrap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
         .route("/metrics", get(handlers::metrics::handler))
         .with_state(state_arc.clone());
 
-    #[cfg(feature = "dynamic-weights")]
-    {
-        let updater = tokio::task::spawn(async move {
-            let mut interval = tokio::time::interval(Duration::from_secs(15));
-            loop {
-                interval.tick().await;
-                state_arc.update_provider_weights().await;
-            }
-        });
+    let public_server =
+        axum::Server::bind(&addr).serve(app.into_make_service_with_connect_info::<SocketAddr>());
 
-        select! {
-            _ = shutdown.recv() => info!("Shutdown signal received, killing servers"),
-            _ = axum::Server::bind(&private_addr).serve(private_app.into_make_service()) => info!("Private server terminating"),
-            _ = axum::Server::bind(&addr).serve(app.into_make_service_with_connect_info::<SocketAddr>()) => info!("Server terminating"),
-            _ = updater => info!("Updater terminating")
-        }
-    }
+    let private_server = axum::Server::bind(&private_addr)
+        .serve(private_app.into_make_service_with_connect_info::<SocketAddr>());
 
-    #[cfg(not(feature = "dynamic-weights"))]
+    let services = vec![
+        tokio::spawn(public_server),
+        tokio::spawn(private_server),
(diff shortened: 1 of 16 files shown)


xDarksome:
What's up with tests failing on ci?

src/lib.rs
```diff
@@ -151,23 +151,25 @@ pub async fn bootstrap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
         .route("/metrics", get(handlers::metrics::handler))
         .with_state(state_arc.clone());
 
-    let updater = tokio::task::spawn(async move {
-        let mut interval = tokio::time::interval(Duration::from_secs(15));
-        loop {
-            interval.tick().await;
-            state_arc.update_provider_weights().await;
+    #[cfg(feature = "dynamic-weights")]
```
xDarksome, line 154: Why it's under a feature flag?

src/lib.rs
```diff
@@ -151,23 +151,25 @@ pub async fn bootstrap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
...
+                interval.tick().await;
+                state_arc.update_provider_weights().await;
+            }
+        });
+
+        select! {
+            _ = shutdown.recv() => info!("Shutdown signal received, killing servers"),
+            _ = axum::Server::bind(&private_addr).serve(private_app.into_make_service()) => info!("Private server terminating"),
+            _ = axum::Server::bind(&addr).serve(app.into_make_service_with_connect_info::<SocketAddr>()) => info!("Server terminating"),
+            _ = updater => info!("Updater terminating")
```
xDarksome, line 168: Consider using https://docs.rs/futures/latest/futures/future/struct.OptionFuture.html to not duplice `select` macro

xDarksome (changes requested):

src/lib.rs
Rakowskiii: This dynamic-weights system will need to live for a while on staging before going to prod. 
The plan is to (for now) keep it as feature flag, and when confirmed working, we can remove the feature and just move it to main code

src/lib.rs
Rakowskiii: @xDarksome if I do that, marking it as None when not using the "dynamic-weights" feature, it will quit instantly, as select waits for the first future to finish, and OptionFuture::None will finish instantly

xDarksome: Ah, yea, we should use `join` instead anyway. As I mentioned before this is not a proper way to handle graceful shutdowns. But it would require a bit of refactoring plus this `updater` thing doesn't actually fit very well in the `join` concept.

Try https://docs.rs/futures/latest/futures/future/fn.select_all.html for now

xDarksome (approved):

xDarksome:
Look good, but fix the tests pls

arein (approved):
Which EVM testnets do we still support now?

Rakowskiii:
> Which EVM testnets do we still support now?

Ethereum Goerli
Optimist Goerli
Polygon Mumbai
Arbitrum goerli

This didn;t change thought. 
We had a route to some more, but they were decomissioned and would just just keep giving errors.

PR #152, opened 2023-06-09
fix: log internal errors

### LINKED ISSUE #151: fix: add logging to rpc proxy
Currently there's no way to figure out why RPC Proxy is throwing HTTP 5xx to the user, unless it's 502. 
Logs are needed to be able to distinguish and understand the reason,

Rakowskiii:
# Description

Adding warn! logs for each unhandled error, so we have at least some ability to see what;s happening internally and take action.

Resolves #151 

## How Has This Been Tested?

Not tested, only adds logging.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,6 +3,7 @@ use {
     axum::{response::IntoResponse, Json},
     cerberus::registry::RegistryError,
     hyper::StatusCode,
+    tracing::log::warn,
 };
 
 pub type RpcResult<T> = Result<T, RpcError>;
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -101,11 +102,14 @@ impl IntoResponse for RpcError {
                 )),
             )
                 .into_response(),
-            _ => (
-                StatusCode::INTERNAL_SERVER_ERROR,
-                "Internal server error".to_string(),
-            )
-                .into_response(),
+            e => {
+                warn!("Internal server error: {}", e);
```
arein, line 106:
```suggestion
                error!("Internal server error: {}", e);
```

arein (approved):

PR #180, opened 2023-06-20
179 use initial weight configuration as multiplier instead of overwriting it

### LINKED ISSUE #179: Use initial weight configuration as multiplier instead of overwriting it
Currently RPC proxy is taking initial weights and updates them accordingly to logged events (errors, rate limits) with no regard to initial config.

That disables the option to prioritize given chain for given provider when e.g. they have better latency/higher limits.

Initial `priority` should be taken into consideration when updating the values so that we have better control

Rakowskiii:
# Description

Separate weights and priorities so that we have more granular control over weights.

Resolves #179 

## How Has This Been Tested?

Tested locally using local prometheus.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/env/binance.rs
+++ b/src/env/binance.rs
@@ -1,4 +1,8 @@
-use {super::ProviderConfig, crate::providers::Weight, std::collections::HashMap};
+use {
+    super::ProviderConfig,
+    crate::providers::{Priority, Weight},
+    std::collections::HashMap,
+};
 
 #[derive(Debug)]
 pub struct BinanceConfig {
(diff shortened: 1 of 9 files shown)


src/env/binance.rs
```diff
@@ -28,14 +32,17 @@ fn default_supported_chains() -> HashMap<String, (String, Weight)> {
         // Binance Smart Chain Mainnet
         (
             "eip155:56".into(),
-            ("https://bsc-dataseed.binance.org/".into(), Weight(6.into())),
+            (
+                "https://bsc-dataseed.binance.org/".into(),
+                Weight::new(Priority::Normal).unwrap(),
```
Rakowskiii, line 37: The unwraps are intentional, we want to know when invalid value is set. It's ran only during initialisation.

chris13524 (commented):
LGTM

arein (approved):

PR #193, opened 2023-06-21
fix: infura provider

Rakowskiii:
# Description

Pokt stopped working for a while for ethereum mainnet. Seems like they just deprecated old url for new one.
Also since we have new big limit for Pokt, redirecting the most troublesome traffic there with higher priority (eip155:1 & eip155:56)

Resolves #192 

## How Has This Been Tested?

Tested locally manually.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #197, opened 2023-06-21
feat: identity caching + metrics

### LINKED ISSUE #53: Tag images with with versions
Each time we merge app logic (as opposed to terraform) we should build an image with a specific tag such that we can pin this tag in deployments and roll back etc.

We do this in other repos e.g. `rs-relay`

### LINKED ISSUE #107: ENS Caching
- [x] Draft ENS Caching docs
- [x] Consult draft version
- [x] Brainstorm the solution on team meeting
- [ ] Implement simple Caching
- [ ] Implement additional logic 
```[tasklist]
### Tasks
- [ ] https://github.com/WalletConnect/rpc-proxy/issues/181
- [ ] https://github.com/WalletConnect/rpc-proxy/issues/182
- [ ] https://github.com/WalletConnect/rpc-proxy/issues/191
- [x] ENS caching
```

### LINKED ISSUE #191: ENS metrics
https://github.com/WalletConnect/web3modal/pull/1090#issuecomment-1600060496

chris13524:
# Description

Adds caching ENS lookups to Redis database.

Adds metrics publishing and new Grafana dashboards.

Also adds ability to deploy local changes to dev environment.

Also fixes https://github.com/hashicorp/terraform-provider-aws/issues/28353

Resolves #191
Resolves #107 

## How Has This Been Tested?

Deployed to dev Terraform environment, made some lookups, checked Grafana charts. Tested against https://github.com/WalletConnect/web3modal/pull/1090

## Due Diligence

* [ ] Breaking change


src/handlers/identity.rs
```diff
@@ -40,18 +42,41 @@ pub async fn handler(
...
+        .parse::<Address>()
+        .map_err(|_| RpcError::IdentityInvalidAddress)?;
+
+    let cache_key = format!("{}", address);
+    if let Some(cache) = &state.identity_cache {
+        let cache_start = SystemTime::now();
+        let value = cache.get(&cache_key).await?;
+        state.metrics.add_identity_lookup_cache_latency(cache_start);
+        if let Some(response) = value {
+            let tld = tld_from_name(&response.name);
```
arein, line 58: What's the cardinality of `tld`? We may only use low cardinality dimensions in Prometheus (< 100 per dimension)

src/handlers/identity.rs
```diff
@@ -40,18 +42,41 @@ pub async fn handler(
...
+            state.metrics.add_identity_lookup_cache_hit(tld.to_string());
```
arein, line 59: this works. But you can also make this a dimension if you like i.e. normalize the metric

arein (approved):
Looks good. Only concern is the cardinality of the dimensions

src/handlers/identity.rs
chris13524: What do you mean exactly?

src/handlers/identity.rs
chris13524: Somewhat low (right now just `eth`). Only valid lookups will be cached. As we add more providers (e.g. Unstoppable Domains) this will go up.

chris13524: An alternative could be a new metric for each identity provider. But maybe this makes more sense as analytics than metrics.

src/handlers/identity.rs
```diff
@@ -61,27 +86,72 @@ pub async fn handler(
...
+
+    if let Some(cache) = &state.identity_cache {
+        let cache = cache.clone();
+        let res = res.clone();
+        // Do not block on cache write.
+        tokio::spawn(async move {
+            cache
+                .set(&cache_key, &res, Some(Duration::from_secs(60)))
+                .await
+                .tap_err(|err| warn!("failed to cache identity lookup: {:?}", err))
```
Rakowskiii, line 136: Maybe we could also log `key` here, so in case something weird happens we know where to look for the problem. Also if user reports problems, we can investigate where it's coming from

chris13524: Added in latest commit

chris13524:
Still need to do a few cleanups to get this right:
- Cache 404s
- Fix latency/availability metrics during 404s

arein (approved):

PR #199, opened 2023-06-22
fix: adjust timeframe for data lookup

### LINKED ISSUE #198: Adjust timeframe for dynamic weights
Currently RPC looks for a data from last hour. During this time a lot of request are made, so when the provider starts failing, it needs to fail a lot before the weights starts going down. 

Decreasing the value by a lot to have faster weight decrease and faster buildback after provider is working again

Rakowskiii:
# Description

Adjusting the timeframe for quicker weight changes.

Resolves #198 

## How Has This Been Tested?

Tested locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/weights.rs
+++ b/src/providers/weights.rs
@@ -72,7 +72,7 @@ fn calculate_chain_weight(
     let Availability(chain_success, chain_failure) = chain_availability;
 
     // Sum failed and successful calls for chain
-    let chain_total = chain_success + chain_failure;
+    let chain_total = chain_success + chain_failure * chain_failure;
 
     // Provider success rate is the amount of successful calls to provider over the
     // total amount of calls to provider
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -172,7 +172,7 @@ impl ProviderRepository {
 
         match self
             .prometheus_client
-            .query("round(increase(provider_status_code_counter[5m]))")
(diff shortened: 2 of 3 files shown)


arein (commented):
Seems reasonable in general and with the amount of traffic.

But the `exponential` failover on a per-chain-id basis should cover this already anyhow imo?

Assume the following math:
100k calls for a chain id in the past hour
500 failed in the last hour
500^2 = 250k

so that should massively impact the availability of that chain id.

Is the exponential logic working?

src/providers/weights.rs
```diff
@@ -72,7 +72,7 @@ fn calculate_chain_weight(
     let Availability(chain_success, chain_failure) = chain_availability;
```
arein, line 72: not sure but does this need to be updated, too?

```suggestion
    let Availability(chain_success, chain_failure * chain_failure) = chain_availability;
```

arein (approved):

src/providers/weights.rs
Rakowskiii: Doing this would mean, that if chain Y fails, chain X will be punished exponentially. 
We can consider it, but wouldn't do it for now

Rakowskiii: If the whole provider is down (rate limit) it should react quickly enough, but if only one chain fails (like Pokt - mainnet eth recently) we don't want the whole provider to go down too quickly

PR #222, opened 2023-07-03
feat: no more 404 for name not found

### LINKED ISSUE #215: feat(identity): return JSON value when identity can't be found, not a 404
See my comment here: https://github.com/WalletConnect/web3modal/pull/1090#discussion_r1248244120

chris13524:
# Description

No longer return a 404 when the name cannot be found. Instead, the `name` field in the response is now optional and you'll get back `{ name: null, avatar: null }`.

Consumers of this API should assume that a `null` name will also result in a `null` avatar, but this isn't enforced by the types. This was simpler than `{ name: string, avatar: string | null } | null`

Resolves #215 

## How Has This Been Tested?

Locally with Web3Modal

## Due Diligence

* [x] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -140,62 +140,60 @@ async fn lookup_identity_rpc(
         headers,
     });
 
-    debug!("Beginning name lookup");
-    state.metrics.add_identity_lookup_name();
-    let name_lookup_start = SystemTime::now();
-    let name = provider
-        .lookup_address(address)
-        .await
-        .tap_err(|err| debug!("Error while looking up name: {err:?}"))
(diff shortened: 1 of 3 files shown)


arein:
Hmmm... why this change?
Why is this a better spec?

Rakowskiii (approved):

chris13524:
@arein this is better because before when the name cannot be found, it results in a [404 error in the JavaScript console](https://github.com/WalletConnect/web3modal/pull/1090#discussion_r1248244120). Many people do not have ENS names, and I would prefer a clean console for those people.

arein (approved):

PR #231, opened 2023-07-05
fix: checked mul

### LINKED ISSUE #230: Panic on multiplication
After the outage we accumulated so many failures that square of them went out of bounds for multiplication.
Needed change to checked mul

Rakowskiii:
# Description

After downtime of metrics we accumulated a lot of failures -> we overflowed on squaring failures count. 
Changed from u32 -> u64 and added checked multiplication.

Resolves #230 

## How Has This Been Tested?

Ran test.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/weights.rs
+++ b/src/providers/weights.rs
@@ -9,8 +9,8 @@ use {
 /// The amount of successful and failed requests to a provider
 ///
 /// Availability(success_counter, failure_counter)
-#[derive(Debug)]
-pub struct Availability(u32, u32);
+#[derive(Debug, Copy, Clone)]
+pub struct Availability(u64, u64);
 
 pub type ParsedWeights = HashMap<String, (HashMap<ChainId, Availability>, Availability)>;
 
@@ -47,11 +47,11 @@ pub fn parse_weights(prometheus_data: PromqlResult) -> ParsedWeights {
                 .or_insert_with(|| Availability(0, 0));
 
             if status_code.starts_with('2') || status_code == "404" || status_code == "400" {
-                provider_availability.0 += amount as u32;
-                chain_availability.0 += amount as u32;
+                provider_availability.0 += amount as u64;
+                chain_availability.0 += amount as u64;
             } else {
-                provider_availability.1 += amount as u32;
-                chain_availability.1 += amount as u32;
+                provider_availability.1 += amount as u64;
+                chain_availability.1 += amount as u64;
             }
         }
     });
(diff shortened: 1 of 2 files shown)


src/providers/weights.rs
```diff
@@ -62,17 +62,27 @@ const PERFECT_RATIO: f64 = 1.0;
...
-    chain_availability: &Availability,
-) -> u32 {
+    chain_availability: Availability,
+) -> u64 {
     let Availability(provider_success, provider_failure) = provider_availability;
 
     // Sum failed and successful calls for provider
-    let provider_total = provider_success + provider_failure * provider_failure;
+    let Some(provider_failures_squared) = provider_failure.checked_mul(provider_failure) else {
+        // 1 is minimal value for chein weight
```
chris13524, line 71: Do we want a minimal value or a maximal value? Min would result in thrashing, no?
```suggestion
        // 1 is minimal value for chain weight
```

Do we really expect this to overflow given the increase to u64? Is there a better way to handle very large numbers?

Rakowskiii: Not sure what would be the best way to handle this, this is mainly a hotfix, as I want to get it back online.
https://github.com/WalletConnect/rpc-proxy/issues/232
here's the task for future improvements

src/providers/weights.rs
```diff
@@ -62,17 +62,27 @@ const PERFECT_RATIO: f64 = 1.0;
...
+        return 1;
```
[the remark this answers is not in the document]
Rakowskiii: We actually need this return, as this should return from the whole function, it's not final value of this block expression

src/providers/weights.rs
chris13524: Actually, defaulting to 1 will result in over-prioritizing this provider, no? So 1 is definitely worse than defaulting to `provider_failure`?

chris13524: It's also possible we hit an overflow down below:
```rust
let provider_total = provider_success + provider_failures_squared;
```

Rakowskiii: But this 1 is not a value we give to provider_squared. It's value of weight (minial possible weight) we give to this

Rakowskiii: Hence the return function

chris13524: What happens if we set the weight of a provider to 0? We should do that if availability goes below e.g. 90%.

Rakowskiii: I'd return to the edge case on addition with the other issue, for now just want to get it back online, and this should work in 99,9% of cases

Rakowskiii: If we set it to 0 we will disable the provider. And actually I think this is what we should do

chris13524 (approved):

PR #237, opened 2023-07-07
221 collect the memory profiles and identify source of the memory leak

### LINKED ISSUE #221: Collect the memory profiles and identify source of the memory leak
https://github.com/WalletConnect/rs-relay/compare/master...feat/heap-profiler-2.0

Rakowskiii:
# Description

Adding memory profiling functionality so that we can debug the memory leak

Resolves #221 

## How Has This Been Tested?

Ran locally and verified it deploys data to s3
image

## Due Diligence

* [ ] Breaking change


.github/workflows/cd.yaml
```diff
@@ -61,6 +61,7 @@ env:
   TF_VAR_infura_project_id: ${{ secrets.INFURA_PROJECT_ID }}
   TF_VAR_pokt_project_id: ${{ secrets.POKT_PROJECT_ID }}
   TF_VAR_registry_api_auth_token: ${{ secrets.REGISTRY_API_AUTH_TOKEN }}
+  TF_VAR_debug_secret: ${{ secrets.DEBUG_SECRET }}
```
heilhead, line 64: You don't really need this. It's fine to hardcode it, unless you plan on keeping the profiler.

Rakowskiii: I think we can keep it, and this repo is public so I'd rather not hardcode it

heilhead: Sorry, didn't know the repo was public.

heilhead (approved):

src/debug/profiler.rs
```diff
@@ -0,0 +1,183 @@
...
+    let auth = headers.get("Authorization");
+    if auth.is_none() {
+        return (hyper::StatusCode::UNAUTHORIZED, "Unauthorized");
+    }
+
+    if let Ok(auth) = auth.unwrap().to_str() {
+        if auth != state.config.debug.secret {
+            return (hyper::StatusCode::UNAUTHORIZED, "Unauthorized");
+        }
+    }
```
chris13524, line 66: security: If `.to_str()` returns `Err(..)` ([this happens if header value contains non-ASCII characters](https://docs.rs/hyper/latest/hyper/header/struct.HeaderValue.html)) then the secret isn't compared.

Furthermore, HeaderValue can be compared to strings, enabling this all to be simplified to:

```suggestion
    if auth != Some(state.config.debug.secret) {
        return (hyper::StatusCode::UNAUTHORIZED, "Unauthorized");
    }
```

chris13524 (changes requested):

src/debug/profiler.rs
```diff
@@ -0,0 +1,183 @@
...
+    (StatusCode::OK, "Profiler started.")
+}
+
+async fn run_profiler(payload: ProfilerPayload) {
+    let Ok(lock) = PROFILER_LOCK.try_lock() else {
+        tracing::warn!("profiler: failed to start because already running");
+        return;
+    };
+
+    if let Err(err) = profile(Duration::from_secs(payload.profile_duration)).await {
```
chris13524, line 78: suggestion: Is it worth putting a max value on `profile_duration` (e.g. 1hr) so that you don't accidentally set it too high and then get locked out?

src/debug/profiler.rs
```diff
@@ -0,0 +1,183 @@
...
+    // thread.
+    tokio::time::sleep(Duration::from_secs(30)).await;
+
+    let profiles_glob = "./dhat-*.json";
+
+    let Ok(file_list) = glob::glob(profiles_glob) else {
+        tracing::warn!("profiler: failed to read glob pattern");
+        return;
+    };
+
```
chris13524, line 115: Would be nice to know the error, just in-case

```suggestion
    let file_list = match glob::glob(profiles_glob) {
        Ok(file_list) => file_list,
        Err(err) => {
            tracing::warn!(?err, "profiler: failed to read glob pattern");
            return;
        }
    };

```

src/debug/profiler.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    tracing::warn!("profiler: finished uploading profiles");
+}
+
+async fn upload_file(ctx: UploadContext, path: PathBuf, key_prefix: &str) -> anyhow::Result<()> {
+    let Ok(metadata) = tokio::fs::metadata(&path).await else {
+        anyhow::bail!("failed to retrieve file metadata");
+    };
```
chris13524, line 156: Can't we use `?` here to preserve the original error?

```suggestion
    let metadata = tokio::fs::metadata(&path).await?;
```

Or maybe we need `.map_err()`:

```suggestion
    let metadata = tokio::fs::metadata(&path).await
        .map_err(|err| anyhow::anyhow!("failed to retrieve file metadata: {err:?}"))?;
```

src/debug/profiler.rs
Rakowskiii: Are you sure about comparing HeaderValue to String? Doesn't seem to work for me

src/debug/profiler.rs
Rakowskiii: I think we may want to run longer profiles, set it to 12h

src/debug/profiler.rs
chris13524: According to these examples it can be done: https://docs.rs/hyper/latest/hyper/header/struct.HeaderMap.html#method.get

The [docs here](https://docs.rs/hyper/latest/hyper/header/struct.HeaderValue.html) say it is easy to compare to a string.

chris13524 (approved):

PR #239, opened 2023-07-07
feat: add metrics for number of ws connections

### LINKED ISSUE #238: Add ws metrics to proxy

Rakowskiii:
# Description

Adding metric there so we can see if websockets are used at all.
Relay had similar mem leak problem, and the team tracked it down to tungstenite. 
This will confirm if it's worth to go down that road

Resolves #238 

## How Has This Been Tested?

test ran

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM.

Not sure if you want to monitor "active" websocket connections or not, as this just monitors WS connections that are opened.

PR #242, opened 2023-07-10
fix: force timeout on service tasks

heilhead:
# Description

This adds a forced timeout on hyper's service tasks. Ported from [the relay](https://github.com/WalletConnect/rs-relay/blob/master/src/http/mod.rs#L41).

## How Has This Been Tested?

Untested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ use {
         handlers::identity::IdentityResponse,
         metrics::Metrics,
         project::Registry,
+        server::ServiceTaskExecutor,
         storage::{redis, KeyValueStorage},
     },
     anyhow::Context,
@@ -47,6 +48,8 @@ use {
     wc::metrics::ServiceMetrics,
 };
 
+const SERVICE_TASK_TIMEOUT: Duration = Duration::from_secs(300);
+
 mod analytics;
 pub mod debug;
 pub mod env;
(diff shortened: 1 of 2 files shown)


xDarksome (approved):

Rakowskiii (approved):

src/lib.rs
```diff
@@ -159,10 +163,12 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
         .route("/metrics", get(handlers::metrics::handler))
         .with_state(state_arc.clone());
 
-    let public_server =
-        axum::Server::bind(&addr).serve(app.into_make_service_with_connect_info::<SocketAddr>());
+    let public_server = axum::Server::bind(&addr)
+        .executor(ServiceTaskExecutor::new(SERVICE_TASK_TIMEOUT))
```
chris13524, line 167: Is this going to mess up long-running WebSocket connections?

heilhead: It's possible, yes. Will need to test to confirm. But the good news is that there's pretty much no websocket traffic currently.

chris13524: Ok.

Maybe something we may be able to improve in the future for ServiceTaskExecutor is a oneshot shutdown signal that e.g. is sent 5s before killing the task. Would enable the WebSocket to gracefully close.

heilhead: We're still trying to get to the bottom of this. I'm not convinced that a custom service task executor is the solution to our problem, but it does work.

PR #249, opened 2023-07-13
feat: update `utils-rs`

heilhead:
# Description

This integrates the latest stuff from `utils-rs` and removes the profiler, as its integration would be very different now that it's included in `utils-rs`, and because we don't really need it at this point.

## How Has This Been Tested?

Untested.

## Due Diligence

* [ ] Breaking change


src/lib.rs
```diff
@@ -187,9 +192,8 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
         }
     };
 
-    let memory_metrics = debug::alloc::AllocMetrics::new();
     let memory_debug_data_collector = async move {
-        if let Err(e) = tokio::spawn(debug::debug_metrics(memory_metrics, config)).await {
+        if let Err(e) = tokio::spawn(debug::debug_metrics()).await {
```
xav, line 196: I guess memory is the only metric we gather by default right now, but shouldn't we change the message to just "Debug stats" if this ever changes?

xav (approved):

PR #292, opened 2023-08-17
feat: Base & Zora

### LINKED ISSUE #291: feat: Base & Zora

chris13524:
# Description

Support Base and Zora via official RPC.

Note that the Base [docs](https://docs.base.org/network-information/) disclaims the following, so if we want more production-ready support we will need to consider other providers and potentially pay. [Available providers are listed on the Base docs.](https://docs.base.org/tools/node-providers)
>  Rate limited and not for production systems.

Base only supports HTTP RPC, but Zora supports both HTTP and Websocket RPC.

Fix: HTTP RPC supported chains list used to setup WebSocket RPC supported chains.

Resolves #291

## How Has This Been Tested?

Locally by running rpc-proxy and testing with curl (HTTP RPC) and JavaScript WebSockets.

## Due Diligence

* [ ] Breaking change


arein (approved):
Wanna update the `tests` to cover these?

chris13524:
@arein 

> Wanna update the tests to cover these?

Tests have been added. Can you re-review?

arein (approved):

PR #315, opened 2023-09-03
feat: transaction history

arein:
# Description

Adds a Transaction History API using Zerion.

This is mostly the scaffolding, will need to iterate on the response API fields and some refactors in follow up PRs.

## How Has This Been Tested?

Tested manually. Automated tests later.

## Due Diligence

* [ ] Breaking change


0xAsimetriq (commented):
Will leave final approval for Chris as there's rust involved, overall looks ok to me 👍 I assume current endpoints and responses are identical to that of Zerion (minus api key auth), but if not could we add a small readme section?

arein:
> Will leave final approval for Chris as there's rust involved, overall looks ok to me 👍 I assume current endpoints and responses are identical to that of Zerion (minus api key auth), but if not could we add a small readme section?

No I decided that's the wrong strategy as it's too much vendor lock-in. So instead I'm removing as many fields as possible only exposing the ones we actually need.

Starting with a very limited set and then adding more. Sample response below. In the next iteration I'll add all the fields you need

```
{
      "id": "ad6e7a58be295c459efdcc43dc5a3ba1",
      "metadata": {
        "operation_type": "send",
        "hash": "0x1f30b4f08ce2cea983533752ec8268240c54c7609e8eda5c12a0b114ab82f5ab",
        "mined_at": "2023-08-07T08:38:31Z",
        "sent_from": "0xf3ea39310011333095cfcccc7c4ad74034caba63",
        "sent_to": "0x2791bca1f2de4661ed88a30c99a7a9449aa84174",
        "status": "confirmed",
        "nonce": 194
      }
    }
```

src/providers/mod.rs
```diff
@@ -48,6 +50,8 @@ pub struct ProviderRepository {
 
     prometheus_client: prometheus_http_query::Client,
     prometheus_workspace_header: String,
+
+    pub history_provider: Arc<dyn HistoryProvider>,
```
chris13524, line 54: Providers are RPC providers, I don't think we should have higher level APIs like txn history in the same spot. I think we should keep separate code structure wise.

arein: Are you just suggesting I make a new folder `history_providers`? That's fine. Deliberately did this as there will be more things than just tx history. So I want some flexibility for renaming later.

arein: Thinking more about this it is just a provider. Some providers support tx history others don't. Just like we sockets

src/lib.rs
```diff
@@ -146,6 +146,8 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
         .route("/v1/", post(handlers::proxy::handler))
         .route("/ws", get(handlers::ws_proxy::handler))
         .route("/v1/identity/:address", get(handlers::identity::handler))
+        .route("/v1/:address/identity", get(handlers::identity::handler))
+        .route("/v1/:address/history", get(handlers::history::handler))
```
chris13524, line 150: I don't like `/v1/:address` being so close to `/v1/` and having totally different functions. I'd prefer something like `/v1/account/:address/history` or `/v1/history/:address` to more clearly separate the functions of the higher level endpoints and the RPC endpoints.

src/providers/zerion.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let hyper_request = hyper::http::Request::builder()
+            .uri(uri)
+            .header("Content-Type", "application/json")
+            .header("authorization", format!("Basic {}", self.api_key))
+            .body(hyper::body::Body::from(body))?;
+
+        let response = self.http_client.request(hyper_request).await?;
+
+        if response.status() != http::StatusCode::OK {
```
chris13524, line 84:
```suggestion
        if !response.status().is_success() {
```

src/providers/zerion.rs
```diff
@@ -0,0 +1,131 @@
...
+        &self,
+        address: String,
+        body: Bytes,
+    ) -> RpcResult<HistoryResponseBody> {
+        let uri = format!(
+            "https://api.zerion.io/v1/wallets/{}/transactions/?currency=usd",
+            address
+        );
+
+        let hyper_request = hyper::http::Request::builder()
```
chris13524, line 76: Probably nicer to use `reqwest` here as it provides a simpler API and a convenient `.json()` function on the response.

chris13524 (approved):

src/providers/zerion.rs
arein: Ah nice!

PR #330, opened 2023-09-30
fix: update rust version in docker build

### LINKED ISSUE #329: CI fails to build
the `rustc` version used in docker build is too old.

xav:
# Description

Update rust to `1.72` in Dockerfile.

Resolves #329 

## How Has This Been Tested?

Local tests

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # Build args
 #
 ################################################################################
-ARG                 base="rust:1.65-buster"
+ARG                 base="rust:buster"
 ARG                 runtime="debian:buster-slim"
 ARG                 bin="rpc-proxy"
 ARG                 version="unknown"


Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Build args
 #
 ################################################################################
-ARG                 base="rust:1.65-buster"
+ARG                 base="rust:buster"
```
geekbrother, line 6: Maybe we can stick to the constant version with `rust:1.72-buster`? To prevent possible future dependency problems when the version is upgraded on `rust:buster` to some next minor version or few and can break some dependencies.

geekbrother (approved):

Dockerfile
xav: Since we're using `stable` everywhere else, there's a discrepancy between the rest of the CI and the actual build, so I think it's better that way. And as Nodar said, `stable` is pretty stable 🙂

PR #369, opened 2023-11-20
fix: Pokt rate limits

chris13524:
# Description

Pokt returning rate limit errors. It doesn't return 429 status code so this will parse the response and use that to determine rate limit status. The existing `is_rate_limited()` function doesn't automatically adjust weights, so modifying the response code instead.

Context: https://walletconnect.slack.com/archives/C03SCF66K2T/p1700511418157459

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 # These must be set or some RPC requests will return errors. See https://github.com/WalletConnect/rpc-proxy/issues/294
-export RPC_PROXY_INFURA_PROJECT_ID=""
-export RPC_PROXY_POKT_PROJECT_ID=""
+export RPC_PROXY_PROVIDER_INFURA_PROJECT_ID=""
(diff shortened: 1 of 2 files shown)


arein (approved):

.env.example
```diff
@@ -1,6 +1,6 @@
 # These must be set or some RPC requests will return errors. See https://github.com/WalletConnect/rpc-proxy/issues/294
-export RPC_PROXY_INFURA_PROJECT_ID=""
-export RPC_PROXY_POKT_PROJECT_ID=""
+export RPC_PROXY_PROVIDER_INFURA_PROJECT_ID=""
+export RPC_PROXY_PROVIDER_POKT_PROJECT_ID=""
```
arein, line 3: what's this fixing?

chris13524: https://github.com/WalletConnect/blockchain-api/pull/364 prefixed the configuration values but didn't update the `.env.example`

PR #370, opened 2023-11-20
fix: log JSON RPC errors with OK status code

chris13524:
# Description

Logs all provider responses that contain a JSON RPC error, but use the OK status code. This is strange, and we may want to consider automatically retrying them in the future, and factoring them into the weighting.

Context: https://walletconnect.slack.com/archives/CB247BW9F/p1700522001797329?thread_ts=1700521059.160949&cid=CB247BW9F

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/base.rs
+++ b/src/providers/base.rs
@@ -6,9 +6,10 @@ use {
     },
     async_trait::async_trait,
     axum::response::{IntoResponse, Response},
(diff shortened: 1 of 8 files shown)


arein (approved):
🙏

geekbrother (approved):

PR #376, opened 2023-11-21
fix: double scaling capacity

chris13524:
# Description

Decrease scaling threshold from 50% to 25% to allow more burstiness without maxing out CPU/MEM like it does sometimes.

Also increase max instances from 5 to 10. Looks like past couple of days this only got to 4, but not sure if there is a bug here where 4 is the actual limit and I don't see why this shouldn't be higher anyway to accommodate more scale. We need it to double anyway if the scaling threshold is halved.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -1,5 +1,5 @@
 locals {
-  cpu    = var.environment != "dev" ? 1024 : 256
+  cpu    = var.environment != "dev" ? 2048 : 256
   memory = 2 * local.cpu # 2x is minimum for ECS
 
   file_descriptor_soft_limit = pow(2, 18)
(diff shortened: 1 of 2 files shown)


chris13524:
> wouldn't it be interesting to just change the instance type instead so that it can handle more out of the box? Might be cheaper too.

Good idea to try. I 2x'ed the compute and left scaling as-is.

arein (approved):

arein:
> Also increase max instances from 5 to 10

Don't see this in the PR @chris13524

chris13524:
> Don't see this in the PR @chris13524

@arein see Xav's comment, I changed the approach to just change the instance size rather than touching the scaling

PR #383, opened 2023-11-22
fix: fix Pokt rate limit fix PR & log strange JSON RPC errors

chris13524:
# Description

Fixes the Pokt PR https://github.com/WalletConnect/blockchain-api/pull/369/files#diff-aceb22d366624f307c6cbe4a2ffcc3efe0b342432b722d97257176b80789dccfR97 and re-adds strange error logging previously added in https://github.com/WalletConnect/blockchain-api/pull/370.

Fix is to not return all `parts` because Hyper will add a second content-length header, which will then be duplicated.

Context: https://walletconnect.slack.com/archives/C03SCF66K2T/p1700659169053979?thread_ts=1700647203.176389&cid=C03SCF66K2T

## How Has This Been Tested?

Automated tests

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -30,18 +30,6 @@ mod server;
 mod zksync;
 mod zora;
 
-pub use {
-    base::*,
-    binance::*,
-    infura::*,
-    omnia::*,
-    pokt::*,
-    publicnode::*,
-    server::*,
-    zksync::*,
-    zora::*,
-};
-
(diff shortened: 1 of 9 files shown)


arein (approved):
Can you add a test that covers what failed before?

Can you also make sure that if this would happen again it throws an HtTP 500 instead of empty reply ?

PR #388, opened 2023-11-22
feat(onramp): Coinbase transaction status

arein:
# Description

Integrates Coinbase's transaction history behind feature flag. Allows us to show a prettier UI if a user onramped via Coinbase. Ultimately this should work for all on-/off-ramp providers and deserves some thought. But no harm in adding this behind a feature flag IMO.

## How Has This Been Tested?

Integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -85,6 +85,16 @@ describe('blockchain api', () => {
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
     })
+    it('fulfilled history with coinbase', async () => {
+      let resp: any = await http.get(
+        `${baseUrl}/v1/account/0xf3ea39310011333095CFCcCc7c4Ad74034CABA63/history?projectId=${projectId}&onramp=coinbase`,
+      )
+      expect(resp.status).toBe(200)
+      expect(typeof resp.data.data).toBe('object')
+      expect(resp.data.data).toHaveLength(50)
+      expect(typeof resp.data.next).toBe('string')
+      expect(resp.data.next).toHaveLength(80)
+    })
     it('empty history', async () => {
       let resp: any = await http.get(
         `${baseUrl}/v1/account/0x739ff389c8eBd9339E69611d46Eec6212179BB67/history?projectId=${projectId}`,
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -85,16 +85,6 @@ describe('blockchain api', () => {
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
     })
-    it('fulfilled history with coinbase', async () => {
-      let resp: any = await http.get(
-        `${baseUrl}/v1/account/0xf3ea39310011333095CFCcCc7c4Ad74034CABA63/history?projectId=${projectId}&onramp=coinbase`,
-      )
-      expect(resp.status).toBe(200)
-      expect(typeof resp.data.data).toBe('object')
-      expect(resp.data.data).toHaveLength(50)
-      expect(typeof resp.data.next).toBe('string')
-      expect(resp.data.next).toHaveLength(80)
-    })
     it('empty history', async () => {
       let resp: any = await http.get(
         `${baseUrl}/v1/account/0x739ff389c8eBd9339E69611d46Eec6212179BB67/history?projectId=${projectId}`,
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -65,7 +65,6 @@ async fn handler_internal(
         })?;
 
     if let Some(_onramp) = query.onramp.clone() {
-        // TODO: call coinbase provider
         let coinbase_transactions: HistoryResponseBody = state
             .providers
             .coinbase_pay_provider
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -13,9 +13,9 @@ use {
         response::{IntoResponse, Response},
         Json,
     },
-    ethers::abi::Address,
+    ethers::types::H160,
     hyper::HeaderMap,
-    std::{net::SocketAddr, sync::Arc},
+    std::{net::SocketAddr, str::FromStr, sync::Arc},
     tap::TapFallible,
     tracing::log::error,
     wc::future::FutureExt,
@@ -46,30 +46,27 @@ async fn handler_internal(
     body: Bytes,
 ) -> Result<Response, RpcError> {
     let project_id = query.project_id.clone();
-    let address_hash = address.clone();
-    address
-        .parse::<Address>()
-        .map_err(|_| RpcError::IdentityInvalidAddress)?;
+    let address_hash = H160::from_str(&address).map_err(|_| RpcError::IdentityInvalidAddress)?;
 
     state.validate_project_access(&project_id).await?;
     let latency_tracker_start = std::time::SystemTime::now();
     let mut response = state
         .providers
         .history_provider
-        .get_transactions(address.clone(), body.clone(), query.0.clone())
+        .get_transactions(address_hash, body.clone(), query.0.clone())
         .await
         .tap_err(|e| {
-            error!("Failed to call transaction history with {}", e);
+            error!("Failed to call transactions history with {}", e);
         })?;
 
     if let Some(_onramp) = query.onramp.clone() {
         let coinbase_transactions: HistoryResponseBody = state
             .providers
             .coinbase_pay_provider
-            .get_transactions(address, body, query.0)
+            .get_transactions(address_hash, body, query.0)
             .await
             .tap_err(|e| {
-                error!("Failed to call coinbase transaction history with {}", e);
+                error!("Failed to call coinbase transactions history with {}", e);
             })
             .unwrap_or(HistoryResponseBody {
                 data: Vec::default(),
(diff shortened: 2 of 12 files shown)


src/providers/coinbase.rs
```diff
@@ -0,0 +1,136 @@
...
+impl HistoryProvider for CoinbaseProvider {
+    async fn get_transactions(
+        &self,
+        address: String,
+        body: Bytes,
+        params: HistoryQueryParams,
+    ) -> RpcResult<HistoryResponseBody> {
+        let base = format!(
+            "https://pay.coinbase.com/api/v1/buy/user/{}/transactions",
+            &address
```
chris13524, line 65: security: doesn't look like this is exploitable yet, but we should URL encode `address` so they can't inject a `/` and do anything nasty (e.g. if Coinbase adds a new endpoint)

Looks like same issue with Zerion endpoint which would be good to fix too.

chris13524 (changes requested):

src/providers/coinbase.rs
geekbrother: I think we can use `H160` type for the address argument and parse it somewhere at the upper level to prevent injecting, [like in the identity handler](https://github.com/WalletConnect/blockchain-api/blob/80228a549cad67724f8e96c97806c4783200dc3c/src/handlers/identity.rs#L65).

geekbrother:
I'll take over this as I'm on the blockchain-api work for the on-ramp transactions history.

src/providers/coinbase.rs
geekbrother: Fixed. I've changed to use `H160` as a type for the address in [0c3bc](https://github.com/WalletConnect/blockchain-api/pull/388/commits/0c3bc1389920e492939407076c04a243b7dbf116).

src/handlers/history.rs
```diff
@@ -46,21 +46,45 @@ async fn handler_internal(
...
         .history_provider
-        .get_transactions(address, body, query.0)
+        .get_transactions(address_hash, body.clone(), query.0.clone())
         .await
         .tap_err(|e| {
-            error!("Failed to call transaction history with {}", e);
+            error!("Failed to call transactions history with {}", e);
         })?;
+
+    if let Some(_onramp) = query.onramp.clone() {
```
chris13524, line 62: I would prefer checking this more explicitly:

```suggestion
    if query.onramp.as_ref() == Some("coinbase") {
```

src/providers/coinbase.rs
```diff
@@ -0,0 +1,137 @@
...
+    async fn get_transactions(
+        &self,
+        address: H160,
+        body: Bytes,
+        params: HistoryQueryParams,
+    ) -> RpcResult<HistoryResponseBody> {
+        let base = format!(
+            "https://pay.coinbase.com/api/v1/buy/user/{}/transactions",
+            &address.to_string()
+        );
```
chris13524, line 67: Pretty sure this could be simplified?

```suggestion
        let base = format!("https://pay.coinbase.com/api/v1/buy/user/{address}/transactions");
```

chris13524, line 67: Pretty sure this could be simplified?

```suggestion
        let base = format!("https://pay.coinbase.com/api/v1/buy/user/{address}/transactions");
```

src/providers/coinbase.rs
```diff
@@ -0,0 +1,137 @@
...
+        let mut url = Url::parse(&base).map_err(|_| RpcError::HistoryParseCursorError)?;
+        url.query_pairs_mut().append_pair("page_size", "50");
```
chris13524, line 69: Nice using functions to build the URL!

src/providers/coinbase.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        if let Some(cursor) = params.cursor {
+            url.query_pairs_mut().append_pair("page_key", &cursor);
+        }
+
+        error!(
```
chris13524, line 75: Replace with `info!()`?

src/providers/coinbase.rs
```diff
@@ -0,0 +1,137 @@
...
+        while let Some(next) = body.next().await {
+            bytes.extend_from_slice(&next?);
+        }
+        let body: CoinbaseResponseBody = match serde_json::from_slice(&bytes) {
+            Ok(body) => body,
+            Err(e) => {
+                error!("Error on parsing coinbase transactions response: {:?}", e);
+                return Err(RpcError::TransactionProviderError);
+            }
+        };
```
chris13524, line 110: `.map_err()` and `.tap_err()` and turbofish is nicer IMO

```suggestion
        let body = serde_json::from_slice::<CoinbaseResponseBody>(&bytes)
            .tap_err(|e| error!("Error on parsing coinbase transactions response: {e:?}"))
            .map_err(|_| RpcError::TransactionProviderError)?;
```

src/providers/coinbase.rs
```diff
@@ -0,0 +1,137 @@
...
+                response.status(),
+            );
+            return Err(RpcError::TransactionProviderError);
+        }
+
+        let mut body = response.into_body();
+        let mut bytes = Vec::new();
+        while let Some(next) = body.next().await {
+            bytes.extend_from_slice(&next?);
+        }
```
chris13524, line 103: `reqwest` crate makes all a lot easier with the [`.json()`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json) helper.

chris13524 (changes requested):
Only thing blocking is the `error` -> `info` log change. Otherwise a bunch of nitpicks but logic seems correct.

integration/integration.test.ts
```diff
@@ -85,6 +85,16 @@ describe('blockchain api', () => {
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
     })
+    it('fulfilled history with coinbase', async () => {
```
chris13524, line 88: Technically this test doesn't actually test txns received from Coinbase, so we should add that if we can in the future.

src/providers/coinbase.rs
geekbrother: Thanks for catching this! I've missed this from the original PR, it seems it was added for debugging purposes.
I've removed it as well as other `info!` and added the `instrument` in the [f7c3ea](https://github.com/WalletConnect/blockchain-api/pull/388/commits/f7c3ea4c3a88d7c47609c65b12dcaa28c73d8558).

chris13524 (approved):

src/handlers/history.rs
```diff
@@ -46,21 +46,34 @@ async fn handler_internal(
...
+                .coinbase_pay_provider
+                .get_transactions(address_hash, body.clone(), query.clone().0)
+                .await
+                .tap_err(|e| {
+                    error!("Failed to call coinbase transactions history with {}", e);
+                })?
+        } else {
+            return Err(RpcError::UnsupportedProvider(onramp));
+        }
+    } else {
```
chris13524, line 66: Is this branching correct? Don't we still want to do the normal history provider when Coinbase is enabled?

src/providers/coinbase.rs
```diff
@@ -63,8 +63,8 @@ impl HistoryProvider for CoinbaseProvider {
         params: HistoryQueryParams,
     ) -> RpcResult<HistoryResponseBody> {
         let base = format!(
-            "https://pay.coinbase.com/api/v1/buy/user/{}/transactions",
-            &address.to_string()
+            "https://pay.coinbase.com/api/v1/buy/user/{:#x}/transactions",
+            &address
```
chris13524, line 67: Does this not work? You were using `.to_string()` before and that's what `{}` formatting calls.

```
            "https://pay.coinbase.com/api/v1/buy/user/{address}/transactions",
```

https://docs.rs/fixed-hash/0.8.0/src/fixed_hash/hash.rs.html#219

src/providers/coinbase.rs
```diff
@@ -115,7 +117,7 @@ impl HistoryProvider for CoinbaseProvider {
                     mined_at: f.created_at,
                     nonce: 1, // TODO: get nonce from somewhere
                     sent_from: "Coinbase".to_string(),
-                    sent_to: address.to_string(),
+                    sent_to: format!("{:#x}", address),
```
chris13524, line 120: Lol I think this was already better:

```suggestion
                    sent_to: address.to_string(),
```

chris13524 (commented):
The branching changed, so that may be wrong now. Are you sure that's right?

Also not blocking but not sure why you complicated the H160 string conversion with `{:#x}` now because `.to_string()` and `{}` already format it as hex as we would need.

src/providers/coinbase.rs
geekbrother: `address.to_string()` when the address is `H160` returns a short address instead of the full version.

src/providers/coinbase.rs
geekbrother: `address.to_string()` when the address is H160 returns a short address instead of the full version here. That's why `{:#x}` was used.

src/handlers/history.rs
geekbrother: We have agreed with @enesozturk that the results should be exclusive when we are using the `onramp` without mixing with the Zerions transactions for now.

src/providers/coinbase.rs
chris13524: Ah got it, so looks like there was a bug in the previous commits as well. Makes sense now!

PR #389, opened 2023-11-23
feat: annotate HTTP5XX

arein:
# Description

Annotates [`Internal Error`](https://docs.infura.io/networks/ethereum/json-rpc-methods) RPC code as such and translates to HTTP5XX for our failover algorithm to take it into account.

## How Has This Been Tested?

Not tested but confirmed that these are legit internal errors.

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
LGTM! We definitely should consider `-32603` as an error.

PR #391, opened 2023-11-23
chore: adding tracing instrument spans to handlers and providers

### LINKED ISSUE #372: chore: wrap all calls into request id
Wrap all calls into a Request ID and create a log span so it’s easier to discover issues for a specific request id.

geekbrother:
# Description

This PR adds additional tracing `instruments` to handlers and providers for the additional span logs in case of errors and for proper debugging.

Resolves #372 

🚧 Stacked on top of the #387 

## How Has This Been Tested?

Logs are exposed with proper tracing instrument spans.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -255,7 +255,7 @@ async fn lookup_identity_rpc(
 
 const SELF_PROVIDER_ERROR_PREFIX: &str = "SelfProviderError: ";
 
-#[tracing::instrument]
+#[tracing::instrument(skip_all)]
 async fn lookup_name(
     provider: &Provider<SelfProvider>,
     address: Address,
@@ -279,7 +279,7 @@ async fn lookup_name(
     )
 }
 
-#[tracing::instrument]
+#[tracing::instrument(skip_all)]
 async fn lookup_avatar(
     provider: &Provider<SelfProvider>,
     name: &str,
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -52,6 +52,7 @@ pub async fn handler(
         .await
 }
 
+#[tracing::instrument(skip_all)]
 async fn handler_internal(
     state: State<Arc<AppState>>,
     connect_info: ConnectInfo<SocketAddr>,
@@ -134,6 +135,7 @@ impl IdentityLookupSource {
     }
 }
 
+#[tracing::instrument(skip_all)]
 async fn lookup_identity(
     address: H160,
     state: State<Arc<AppState>>,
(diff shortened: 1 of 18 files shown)


src/handlers/identity.rs
```diff
@@ -134,6 +135,7 @@ impl IdentityLookupSource {
     }
 }
 
+#[tracing::instrument(skip_all, fields(address = %address))]
```
chris13524, line 138: I don't think we should be logging the address since this isn't in our privacy policy. If anything we can log the hash, but not sure it's worth logging anything at all.

chris13524 (changes requested):

src/handlers/identity.rs
geekbrother: The address is currently logging in the [uri section of the span](https://github.com/WalletConnect/blockchain-api/blob/0ebead6182e45c8bc5934050aff97112d032ca15/src/lib.rs#L169). If we are going to remove it, we should remove it from the URI in the span as well.

chris13524: Yeah let's remove 👍

geekbrother: I've removed the address from new spans, but if we remove the [URI logging from the middleware](https://github.com/WalletConnect/blockchain-api/blob/0ebead6182e45c8bc5934050aff97112d032ca15/src/lib.rs#L169) we will miss most of the context. Since the address is a part of the URI itself.

src/handlers/identity.rs
```diff
@@ -252,6 +255,7 @@ async fn lookup_identity_rpc(
 
 const SELF_PROVIDER_ERROR_PREFIX: &str = "SelfProviderError: ";
 
+#[tracing::instrument]
```
chris13524, line 258: Can we `skip_all` here too?

src/handlers/identity.rs
```diff
@@ -275,6 +279,7 @@ async fn lookup_name(
     )
 }
 
+#[tracing::instrument]
```
chris13524, line 282: Can we `skip_all` here too so we don't log ENS names?

src/handlers/identity.rs
chris13524: > if we remove the [URI logging from the middleware](https://github.com/WalletConnect/blockchain-api/blob/0ebead6182e45c8bc5934050aff97112d032ca15/src/lib.rs#L169) we will miss most of the context. Since the address is a part of the URI itself.

We don't need to do in this PR but can't we sensor the address when it detects a URL that would contain an address?

src/handlers/identity.rs
geekbrother: done

src/handlers/identity.rs
geekbrother: done

src/handlers/identity.rs
geekbrother: Extracted into the #402

chris13524 (approved):

PR #395, opened 2023-11-24
fix(ci): image tag version as a short commit SHA instead of `latest`

### LINKED ISSUE #53: Tag images with with versions

geekbrother:
# Description

This PR adds an ECR image version tag as a short SHA commit (eg. `d6ee36e`) instead of always be `latest`.
With this change, we can track what version was shipped and can ship a certain version to revert.

[GITHUB_SHA](https://docs.github.com/en/actions/learn-github-actions/variables) GitHub CI variable is used to get the commit SHA and shorten it by the `::7`.

Resolves #53 

## How Has This Been Tested?

Not tested, but should babysit it when merging and test it.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -192,7 +192,7 @@ jobs:
         uses: docker/build-push-action@v2
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
-          IMAGE_TAG: latest # TODO #85: use proper tag list
+          IMAGE_TAG: ${GITHUB_SHA::7}
           PUSH: true
         with:
           context: .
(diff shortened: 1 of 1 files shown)


arein (approved):
Howdcha test this?

geekbrother:
> Howdcha test this?

Not so many changes here, the `echo ${GITHUB_SHA::7}` was tested locally and [on my test repo](https://github.com/geekbrother/rust-test-server/blob/main/.github/workflows/rust.yml#L24). 
I will land other changes now and will babysit it 👀 
In any case, if something is wrong here it will not ship anything to staging and prod because the tag will be different. At the moment `latest` tag is deployed and re-deploying when this tag updates and we will not update the `latest` tag anymore.

PR #396, opened 2023-11-27
chore(terraform): applying most recent added ECR image

geekbrother:
# Description

Since #395 we are using the short SHA hash for the commit tag instead of the `latest` for the ECR images.
This PR adds changes to the terraform config to use the latest pushed ECR image instead of with the `latest` tag.

Part of the #53

## How Has This Been Tested?

`terraform output` and, `terraform plan` run.

## Due Diligence

* [ ] Breaking change


geekbrother:
closing this in a favor of #382

PR #398, opened 2023-11-27
fix(o11y): history lookup not tracked

arein:
# Description

Fixes the below panels which depend on `add_history_lookup` which is not called anywhere.

Screenshot 2023-11-27 at 5 59 06 AM

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM but @geekbrother should approve as well

PR #401, opened 2023-11-27
ci: append the `latest` tag for pushed image

geekbrother:
# Description

This PR adds an additional `latest` tag for the pushed ECR app image to use the `latest` for the most recent image instead of searching for the recent commit (as we are deploying by the commit hash now).
The tags override is not enabled in the ECR repository so we can keep only a single latest tag per image with the commit hash tag.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change


geekbrother:
closing this in a favor of #382

PR #404, opened 2023-12-05
feat(identity): remove ENS demo

arein:
# Description

Removes the mocked reverse ENS lookup until we actually ship the feature.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -175,7 +175,7 @@ async fn lookup_identity(
             debug!("Setting cache success");
         });
     }
-    
+
     Ok((IdentityLookupSource::Rpc, res))
 }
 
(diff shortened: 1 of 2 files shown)


src/handlers/identity.rs
```diff
@@ -146,39 +146,8 @@ async fn lookup_identity(
 ) -> Result<(IdentityLookupSource, IdentityResponse), RpcError> {
     let cache_key = format!("{}", address);
 
-    // Skipping the cache for testing addresses
-    let ens_demo_addresses = state.ens_allowlist.clone().unwrap_or_default();
-    if !ens_demo_addresses.contains_key(&address) {
-        if let Some(cache) = &state.identity_cache {
```
geekbrother, line 152: This part should not be scrapped because this is a getting from the cache. We have wrapped it into the `if !ens_demo_addresses.contains_key(&address)` before, to skip the cache lookup for the mock.

geekbrother (changes requested):

geekbrother (approved):
LGTM!

PR #409, opened 2023-12-08
feat: exposing `implementations` for the transactions history

### LINKED ISSUE #408: feat: expose chain id in transactions history
To add the network icon in the transaction history UI, the chain id should be exposed.

geekbrother:
# Description

Exposing the `chain_id` and `address` from the `implementations` for the `fungible_info`.
The updated response will contain new exposed fields:
* `chain_id` - Unique id of the chain.
* `address` - Implementation address on the chain.

This is a requirement for the transactions history UI to get the chain on where was the transaction and add an icon.

Example updated response:
```
{
...
  "transfers":[
    {
      "fungible_info":
        ...
        "implementations": [
              {
                "chain_id": "polygon",
                "address": "0x3c499c542cef5e3811e1192ce70d8cc03d5c3359"
              }
            ]
...
}
```

Resolves #408 

## How Has This Been Tested?

**Manually:**

Start the server locally by running `cargo run` with a valid Zerion RPC key in `RPC_PROXY_PROVIDER_ZERION_API_KEY` environment variable.
Request the history by the curl command `curl -v 'http://localhost:3000/v1/account/0xf3ea39310011333095CFCcCc7c4Ad74034CABA63/history?projectId=xxx`.
Th expected result is that response should contain the `implementations`.

**CI:**

Integration test for the history endpoint.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -110,11 +110,21 @@ pub struct ZerionTransactionTransferQuantity {
     pub numeric: String,
 }
 
+/// Implementation details of the fungible on various chains.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub struct ZerionTransactionImplementation {
+    /// Unique id of the chain.
+    pub chain_id: String,
+    /// Implementation address on the chain.
+    pub address: Option<String>,
+}
+
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct ZerionTransactionFungibleInfo {
     pub name: Option<String>,
     pub symbol: Option<String>,
     pub icon: Option<ZerionTransactionURLItem>,
+    pub implementations: Vec<ZerionTransactionImplementation>,
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]


tomiir (commented):
This looks good! though I think we would be missing nft chain data 
According to Zerion docs there's an optional `relationships` key that *could* include the chain:
```
pub struct ZerionTransactionsReponseBody {
    pub r#type: String,
    pub id: String,
    pub attributes: ZerionTransactionAttributes,
    pub relationships: ZerionTransactionRelationships
}
```

Check this out: 
image

The thing is, it's not required.
I think it might be worth asking Zerion team what kind of transactions would have that field populated to check if it serves our purpose

src/providers/zerion.rs
```diff
@@ -110,11 +110,21 @@ pub struct ZerionTransactionTransferQuantity {
...
+    /// Implementation address on the chain.
+    pub address: Option<String>,
+}
+
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
 pub struct ZerionTransactionFungibleInfo {
     pub name: Option<String>,
     pub symbol: Option<String>,
     pub icon: Option<ZerionTransactionURLItem>,
+    pub implementations: Vec<ZerionTransactionImplementation>,
```
arein, line 127:
```suggestion
    pub chain_id: String,
```

Tenet of this service is to be provider agnostic. So let's refrain from exposing as much data as possible from Zerion's API and instead expose only the required data such that if we ever want to add another provider it becomes easier.

arein (changes requested):
Tenet of this service is to be provider agnostic. So let's refrain from exposing as much data as possible from Zerion's API and instead expose only the required data such that if we ever want to add another provider it becomes easier.

chris13524 (commented):
Should there be tests for this? Also shouldn't we normalize the chain ID to a standard number instead of the string?

geekbrother:
Closing this in favor of #494

PR #410, opened 2023-12-11
feat: add Postgres 16 to the docker-compose

geekbrother:
# Description

This PR adds the Postgres 16 to the docker-compose to use in the further CI tests.

Resolves #405

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #411, opened 2023-12-11
feat: add sql schema and migrations for the ENS

### LINKED ISSUE #405: feat: implement ENS offchain gateway with reverse lookup

geekbrother:
# Description

This PR adds the SQL database tables schema migration files for the sqlx according to the [SPECS](https://walletconnect-specs-git-max-feathexlessspec-walletconnect1.vercel.app/2.0/specs/servers/blockchain/blockchain-server-api).

Resolves #405 

## Stacked PRs list 🏗️

* 6: add lookup handlers #417 

## How Has This Been Tested?

When the server is started the migration procedure is invoked in #412 . The expected result is a successful server start.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/migrations/20231209213851_init-names.sql
+++ b/migrations/20231209213851_init-names.sql
@@ -1,0 +1,15 @@
+-- Creating the hstore extension for the attributes column
+CREATE EXTENSION hstore;
+
+-- Initializing the names table
+CREATE TABLE names (
+  name VARCHAR(255) PRIMARY KEY,
+  registered_at TIMESTAMPTZ NOT NULL  DEFAULT now(),
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+
+  -- We are using the hstore as a key-value for the extensible attributes list
+  attributes hstore,
+
+  -- Check for the standartized name format
+  CONSTRAINT ens_name_standard CHECK (name ~ '^[a-z0-9.-]*$')
+);
--- a/migrations/20231209213904_init-addresses.sql
+++ b/migrations/20231209213904_init-addresses.sql
@@ -1,0 +1,33 @@
+-- List of supported blockchain namespaces
+CREATE TYPE namespaces AS ENUM (
+  'eip155' -- Ethereum
+);
+
+-- Initializing the addresses table
+CREATE TABLE addresses (
+  -- Breakdown of the CAP-10 address format into namespace:chain_id:address
+  namespace namespaces,
+  /*
+  chain_id can represent a chain id e.g. (Cosmos and cosmoshub-3 chain):
+    cosmos:cosmoshub-3:cosmos1t2uflqwqe0fsj0shcfkrvpukewcw40yjj6hdc0
+  chain_id can be empty e.g. (Litecoin mainnet):
+    bip122:12a765e31ffd4059bada1e25190f6e98
+  */
+  chain_id VARCHAR(255) NOT NULL,
+  address VARCHAR(255) NOT NULL,
+
+  name  VARCHAR(255) REFERENCES  names (name) ON DELETE CASCADE,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+
+  PRIMARY KEY (name, namespace, chain_id, address)
+);
+
+-- Creating indexes for the address lookups
+CREATE INDEX index_cap_10_format_address
+  ON addresses (namespace, chain_id, address);
+CREATE INDEX index_namespace_address
+  ON addresses (namespace, address);
+CREATE INDEX index_address
+  ON addresses (address);
+CREATE INDEX index_name
+  ON addresses (name);
--- a/migrations/README.md
+++ b/migrations/README.md
@@ -1,0 +1,17 @@
+# Migrations
+
+This folder contains SQL migration scripts and they are automatically run on start-up.
+
+## New Migration
+
+To create new migration file sqlx-cli must be installed:
+
+```
+cargo install sqlx-cli
+```
+
+Create a new migration with the `name`:
+
+```
+sqlx migrate add <name>
+```


migrations/20231209213851_init-names.sql
```diff
@@ -0,0 +1,15 @@
+-- Creating the hstore extension for the attributes column
+CREATE EXTENSION hstore;
```
chris13524, line 2: Do we have permissions to do this on RDS?

Also, why `hstore` instead of a related table?

migrations/20231209213904_init-addresses.sql
```diff
@@ -0,0 +1,33 @@
+-- List of supported blockchains
+CREATE TYPE supported_blockchains AS ENUM (
```
chris13524, line 2: These are actually called "namespaces" not blockchains. Can we change that terminology?

migrations/20231209213851_init-names.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- Initializing the names table
+CREATE TABLE names (
+  name VARCHAR(255) PRIMARY KEY,
+  registered_at TIMESTAMPTZ NOT NULL  DEFAULT now(),
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+
+  -- We are using the hstore as a key-value for the extensible attributes list
+  attributes hstore,
```
chris13524, line 11: I think the attributes should be part of the `addresses` table, so that way when you change your name it doesn't result in a new set of attributes.

migrations/20231209213904_init-addresses.sql
```diff
@@ -0,0 +1,33 @@
...
+  /*
+  chain_id can represent a chain id e.g. (Cosmos and cosmoshub-3 chain):
+    cosmos:cosmoshub-3:cosmos1t2uflqwqe0fsj0shcfkrvpukewcw40yjj6hdc0
+  chain_id can be empty e.g. (Litecoin mainnet):
+    bip122:12a765e31ffd4059bada1e25190f6e98
+  */
+  chain_id VARCHAR(255),
+  address VARCHAR(255) NOT NULL,
+
+  name  VARCHAR(255) REFERENCES  names (name) ON DELETE CASCADE,
```
chris13524, line 19: What about historical names?

migrations/20231209213851_init-names.sql
```diff
@@ -0,0 +1,15 @@
...
+
+  -- Check for the standartized name format
+  CONSTRAINT ens_name_standard CHECK (name ~ '^[a-z0-9.-]*$')
```
chris13524, line 14: How expensive is this?

chris13524 (changes requested):
I think the structure needs to be changed to support historical names. Also "blockchain" to "namespace"

migrations/20231209213851_init-names.sql
```diff
@@ -0,0 +1,15 @@
+-- Creating the hstore extension for the attributes column
+CREATE EXTENSION hstore;
+
+-- Initializing the names table
+CREATE TABLE names (
+  name VARCHAR(255) PRIMARY KEY,
```
chris13524, line 6: Not sure I like using a string as the primary key for linking the table together

migrations/20231209213904_init-addresses.sql
geekbrother: Yes, we have redirects in our specs for the old names. Since it's not on the table now, the idea is to add the table for the redirected old names (that will be added later).

migrations/20231209213851_init-names.sql
geekbrother: Not expensive as this is executed only in insert and update.

migrations/20231209213851_init-names.sql
geekbrother: Yes, [we have the permission](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html).
With the related table we will have additional queries in each lookup, and no additional queries for the attributes in the case of hstore. by the way, ENS is using it in their gate implementation as well.

migrations/20231209213904_init-addresses.sql
geekbrother: I don't mind to change it.

migrations/20231209213904_init-addresses.sql
chris13524: I think it would be better to have all the names in 1 table with a UNIQUE key to prevent race conditions.

migrations/20231209213904_init-addresses.sql
geekbrother: Thx. Done.

migrations/20231209213904_init-addresses.sql
geekbrother: Makes sense! We can store multiple names with the same addresses (reference to the name from addresses) in this table (part of them can be historical).
`name  VARCHAR(255) REFERENCES  names (name) ON DELETE CASCADE` will delete the address records if the name is deleted (but we will not delete names until the grace period or something we will figure out how our historical names will work). 
We can add an additional field in this table like `is_historical` to mark the name as historical (but this can be done later when we have a clear view on how the registration, domain retention policy will work).
The name is the central entity here because we don't have a `centralized` account_id or something to bind the names to. The name owner is the address owner, but addresses can be a few in different blockchains. What I would add now is that it's clear that one address should be a primary, so we can add `primary(bool)` to the addresses.

migrations/20231209213851_init-names.sql
geekbrother: Looks scary, but the name is unique, we don't need to add an additional `id` field here.

geekbrother:
> I think the structure needs to be changed to support historical names. Also "blockchain" to "namespace"

Gave my thoughts [in the comment](https://github.com/WalletConnect/blockchain-api/pull/411#discussion_r1426564431).

migrations/20231209213904_init-addresses.sql
```diff
@@ -0,0 +1,33 @@
...
+  chain_id can be empty e.g. (Litecoin mainnet):
+    bip122:12a765e31ffd4059bada1e25190f6e98
+  */
+  chain_id VARCHAR(255) NOT NULL,
+  address VARCHAR(255) NOT NULL,
+
+  name  VARCHAR(255) REFERENCES  names (name) ON DELETE CASCADE,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+
+  PRIMARY KEY (name, namespace, chain_id, address)
```
geekbrother, line 22: The `chain_id` is changed to can't be NULL, but can be an empty string instead to represent mainnet on some of the chains, because it should be part of the composite unique key for the address, but the unique primary key field can't be NULL and NULL can't be constraint. So this is handled by the empty string [further in helpers](https://github.com/WalletConnect/blockchain-api/pull/413/files#diff-98120810190356016b1026cd118005a84df034f5114b17bef040bf69f9aad917R199). In this case we can use it as generic solution for all chains.

migrations/20231209213904_init-addresses.sql
```diff
@@ -0,0 +1,33 @@
+-- List of supported blockchain namespaces
+CREATE TYPE namespaceses AS ENUM (
+  'eip155' -- Ethereum
+);
+
+-- Initializing the addresses table
+CREATE TABLE addresses (
+  -- Breakdown of the CAP-10 address format into namespace:chain_id:address
+  namespace namespaceses,
```
chris13524, line 9: "namespaces" is the plural of "namespace"

```suggestion
  namespace namespaces,
```

chris13524 (approved):
Just one name change I think is necessary, but otherwise LGTM.

I'm a little scared about the hstore and the 4-item primary key, but I haven't used those things before and am willing to try them out. I'm sure we can always change and migrate the data later so all good.

migrations/20231209213904_init-addresses.sql
geekbrother: Fixed! Thanks for catching this.

PR #412, opened 2023-12-11
feat: scaffold sqlx and add `RPC_PROXY_POSTGRES_URI` env variable

geekbrother:
# Description

This PR adds sqlx to the server, invoking migration at the bootstrap, and adds the `RPC_PROXY_POSTGRES_URI` environment variable for the Postgres URI configuration.

Resolves #405 

## Stacked PRs list 🏗️

* 0: add Postgres to Terraform config #415 
* 1: add Postgres 16 to the docker-compose #410
* 2: add sql schema and migrations for the ENS #411 
* 3: scaffold sqlx and add `RPC_PROXY_POSTGRES_URI` env variable #412
* 4: implement database helpers #413
* 5: enable database functional tests #414 
* 6: add lookup handlers #417 
* 7: name registration handler #418 
* 8: profile names integration tests #419

## How Has This Been Tested?

Run functional database tests in #413. The expected result is successfully passing all database tests.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #413, opened 2023-12-11
feat: implement database helpers

### LINKED ISSUE #405: feat: implement ENS offchain gateway with reverse lookup

geekbrother:
# Description

This PR adds database helpers function implementations for the tables schema in #411 and according to the [SPEC](https://walletconnect-specs-git-max-feathexlessspec-walletconnect1.vercel.app/2.0/specs/servers/blockchain/blockchain-server-api) that should be further used in the server endpoint handlers for names and addresses manipulations.

The functional tests to cover helpers are added to this PR.

Resolves #405 

## Stacked PRs list 🏗️

* 6: add lookup handlers #417 

## How Has This Been Tested?

Tested by the included in this PR functional tests, run by the following #414.
* [Results of passing functional tests in CI](https://github.com/WalletConnect/blockchain-api/actions/runs/7210809023/job/19644779358?pr=414#step:9:956)

To run it locally:
* Start the local postgres server:
  `docker run -p 5432:5432 --name postgres -e POSTGRES_HOST_AUTH_METHOD=trust -d postgres`
* Run the cargo test:
  `RPC_PROXY_POSTGRES_URI=postgres://postgres@localhost:5432/postgres cargo test --lib --bins --tests`

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/database/helpers.rs
+++ b/src/database/helpers.rs
@@ -1,0 +1,228 @@
+use {
+    crate::database::{error::DatabaseError, types, utils},
+    sqlx::{PgPool, Postgres},
+    std::collections::HashMap,
+    tracing::instrument,
+};
+
+/// Initial name registration insert
+#[instrument(skip(postgres))]
+pub async fn insert_name(
+    name: String,
+    attributes: HashMap<String, String>,
+    addresses: Vec<types::Address>,
+    postgres: &PgPool,
+) -> Result<(), DatabaseError> {
+    if addresses.is_empty() {
+        return Err(DatabaseError::BadArgument(
+            "At least one address is required for the new name".to_string(),
+        ));
+    }
+    let mut transaction = postgres.begin().await?;
+    let insert_name_query = "
+      INSERT INTO names (name, attributes)
+        VALUES ($1, $2::hstore)
+    ";
+    sqlx::query::<Postgres>(insert_name_query)
+        .bind(&name.clone())
+        // Convert JSON to String for hstore update
+        .bind(&utils::hashmap_to_hstore(&attributes))
+        .execute(&mut *transaction)
+        .await?;
+    for address in addresses {
+        insert_address(
+            name.clone(),
+            address.namespace,
+            address.chain_id,
+            address.address,
+            &mut *transaction,
+        )
+        .await?;
+    }
+    transaction.commit().await.map_err(DatabaseError::SqlxError)
+}
+
+#[instrument(skip(postgres))]
+pub async fn delete_name(
+    name: String,
+    postgres: &PgPool,
+) -> Result<sqlx::postgres::PgQueryResult, sqlx::error::Error> {
+    let query = "
+      DELETE FROM names WHERE name = $1
+    ";
+    sqlx::query::<Postgres>(query)
+        .bind(name)
+        .execute(postgres)
+        .await
+}
+
+#[instrument(skip(postgres))]
+pub async fn update_name(
+    name: String,
+    attributes: HashMap<String, String>,
+    postgres: &PgPool,
+) -> Result<sqlx::postgres::PgQueryResult, sqlx::error::Error> {
+    let insert_name_query = "
+      UPDATE names SET attributes = $2::hstore, updated_at = NOW()
+        WHERE name = $1
+    ";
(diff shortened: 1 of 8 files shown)


src/database/helpers.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[instrument(skip(postgres))]
+pub async fn get_names_by_address(
+    address: String,
+    postgres: &PgPool,
+) -> Result<Vec<types::Name>, sqlx::error::Error> {
+    let query = "
+      SELECT name, registered_at, updated_at, hstore_to_json(attributes) AS attributes
+        FROM names WHERE name IN
+          (SELECT name FROM addresses WHERE address = $1)
```
chris13524, line 101: Can we use a JOIN instead of a subquery?

src/database/helpers.rs
```diff
@@ -0,0 +1,244 @@
...
+#[instrument(skip(postgres))]
+pub async fn get_names_by_address_and_blockchain(
+    address: String,
+    blockchain: types::SupportedBlockchains,
+    postgres: &PgPool,
+) -> Result<Vec<types::Name>, sqlx::error::Error> {
+    let query = "
+      SELECT name, registered_at, updated_at, hstore_to_json(attributes) AS attributes
+        FROM names WHERE name IN
+          (SELECT name FROM addresses WHERE address = $1 AND blockchain = $2)
```
chris13524, line 134: Can we use a JOIN instead of a subquery?

src/database/helpers.rs
```diff
@@ -0,0 +1,244 @@
+use {
+    crate::database::{error::DatabaseError, types, utils},
+    sqlx::{PgPool, Postgres},
+    std::collections::HashMap,
+    tracing::instrument,
+};
+
+/// Initial name registration insert
+#[instrument(skip(postgres))]
+pub async fn insert_name(
```
chris13524, line 10: Missing idempotency/conflict handling

src/database/helpers.rs
```diff
@@ -0,0 +1,244 @@
...
+        INSERT INTO addresses (name, blockchain, chain_id, address)
+        VALUES ($1, $2, NULL, $3)
+        ",
+    )
+    .bind(&name)
+    .bind(&blockchain)
+    .bind(&address);
+
+    // Conditionally add chain_id to the query
+    if let Some(ref chain_id_value) = chain_id {
```
chris13524, line 230: Can't you provide `None` to `bind()` instead of needing a separate query for this case?

src/database/helpers.rs
```diff
@@ -0,0 +1,244 @@
...
+        .await
+}
+
+#[instrument(skip(postgres))]
+pub async fn get_name_and_addresses_by_name(
+    name: String,
+    postgres: &PgPool,
+) -> Result<types::NameAndAddresses, sqlx::error::Error> {
+    let result = get_name(name.clone(), postgres).await?;
+    let addresses = get_addresses_by_name(name, postgres).await?;
```
chris13524, line 149: Ideally this would be 1 database query with the list of addresses returned with e.g. `array_agg()` function. This avoids multiple round trips.

Alternatively would be nice to do these two queries in parallel to reduce execution time.

chris13524 (changes requested):

src/database/helpers.rs
geekbrother: That makes sense. I'll defer it for now, but fire an issue to track right after we merge the PRs stack. Thx.

src/database/helpers.rs
geekbrother: Done. Thx.

src/database/helpers.rs
geekbrother: We will check if the name is already registered in the registration handler by the name lookup, then check the signature, and then call this function to insert. There will be a sqlx error in case someone got the name (constraint error, that will be handled on the handler side) while these 3 steps in handler executed.
Ideally, we can do the registration flow in one transaction so we will not have such race conditions in a registration process. 
But we can think about it later (I'll create a followup issue for that).
At the moment we have to check the name availability and signature before to call this inserting.

chris13524 (approved):
LGTM

PR #415, opened 2023-12-13
chore(terraform): adding Postgres

geekbrother:
# Description

This PR adds the Amazon RDS Postgres engine to the AWS terraform configuration and exposes the `database_uri` to the ECS cluster instances to connect to the Database.

Resolves #405 

## Stacked PRs list 🏗️

* 0: add Postgres to Terraform config #415 
* 1: add Postgres 16 to the docker-compose #410
* 2: add sql schema and migrations for the ENS #411 
* 3: scaffold sqlx and add `RPC_PROXY_POSTGRES_URI` env variable #412
* 4: implement database helpers #413
* 5: enable database functional tests #414 
* 6: add lookup handlers #417 
* 7: name registration handler #418 
* 8: profile names integration tests #419

## How Has This Been Tested?

Tested only by the `terraform plan`.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #414, opened 2023-12-12
feat(ci): enable database functional tests

geekbrother:
# Description

This PR enables functional tests for the `database` module to test database helpers. We can't enable all functional tests by adding `--tests` [because they were broken](https://github.com/WalletConnect/blockchain-api/blob/6053dfc73fcdf979edd8627b1a22035cf96afc97/justfile#L26) and should be fixed.

Resolves #405 

## Stacked PRs list 🏗️

* 0: add Postgres to Terraform config #415 
* 1: add Postgres 16 to the docker-compose #410
* 2: add sql schema and migrations for the ENS #411 
* 3: scaffold sqlx and add `RPC_PROXY_POSTGRES_URI` env variable #412
* 4: implement database helpers #413
* 5: enable database functional tests #414 
* 6: add lookup handlers #417 
* 7: name registration handler #418 
* 8: profile names integration tests #419

### CI:
Passing the new [CI tests](https://github.com/WalletConnect/blockchain-api/actions/runs/7200608467/job/19614927315#step:9:956).

### Locally:
Start the Postgres locally and run `RPC_PROXY_POSTGRES_URI=postgres://postgres@localhost:5432/postgres cargo test functional::database`. The expected result is passing database functional tests.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #417, opened 2023-12-14
feat: add lookup handlers

### LINKED ISSUE #405: feat: implement ENS offchain gateway with reverse lookup

geekbrother:
# Description

This PR implements the following lookup handlers according to the [SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/179):
* Forward lookup `/v1/profile/account/:name`: name -> addresses,
* Reverse lookup `/v1/profile/reverse/:address`: address -> names.

Resolves #405

## Stacked PRs list 🏗️

* 6: add lookup handlers #417 

## How Has This Been Tested?

Run local integration tests in #419

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/profile/lookup.rs
+++ b/src/handlers/profile/lookup.rs
@@ -1,0 +1,45 @@
+use {
+    super::super::HANDLER_TASK_METRICS,
+    crate::{database::helpers::get_name_and_addresses_by_name, error::RpcError, state::AppState},
+    axum::{
+        extract::{Path, State},
+        response::{IntoResponse, Response},
+        Json,
+    },
+    hyper::StatusCode,
+    sqlx::Error as SqlxError,
+    std::sync::Arc,
+    tracing::log::error,
+    wc::future::FutureExt,
+};
+
+pub async fn handler(
+    state: State<Arc<AppState>>,
+    name: Path<String>,
+) -> Result<Response, RpcError> {
+    handler_internal(state, name)
+        .with_metrics(HANDLER_TASK_METRICS.with_name("profile"))
+        .await
+}
+
+#[tracing::instrument(skip(state))]
+async fn handler_internal(
+    state: State<Arc<AppState>>,
+    Path(name): Path<String>,
+) -> Result<Response, RpcError> {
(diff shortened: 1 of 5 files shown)


src/handlers/profile/reverse.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    if names.is_empty() {
+        // Respond with a "Not Found" status code if no name was found for the address
+        let not_found_response = (StatusCode::NOT_FOUND, "No rigistered names for the address");
+        return Ok(not_found_response.into_response());
+    }
+
+    let mut result = Vec::new();
+    for name in names {
+        match get_name_and_addresses_by_name(name.name, &state.postgres).await {
```
chris13524, line 49: This is `1+n` database lookups, this can all be done with a single database query. Would be good to optimize this.

src/handlers/profile/lookup.rs
```diff
@@ -0,0 +1,45 @@
...
+        Err(e) => match e {
+            SqlxError::RowNotFound => {
+                // Respond with a "Not Found" status code if name was found
+                let not_found_response = (StatusCode::NOT_FOUND, "Name is not registered");
+                Ok(not_found_response.into_response())
+            }
+            _ => {
+                // Handle other types of errors
+                error!("Failed to lookup name: {}", e);
+                return Ok((StatusCode::INTERNAL_SERVER_ERROR, "").into_response());
```
chris13524, line 41: non-blocking: I'd prefer if we just returned the error directly and let errors.rs convert into a 500 error, since the number of server-side errors could be large your general approach here is not scalable.

chris13524 (approved):
As a general rule-of-thumb I think we should do things with as few database queries as possible, building the queries around how they are used, rather than building a psudo-ORM.

But it doesn't block the PR.

src/handlers/profile/reverse.rs
geekbrother: Makes sense to optimize this! I'll fire an issue for the db lookups optimization as a follow-up, there is a list already, I'll add your concern for this as well. Thanks for the comment.

PR #418, opened 2023-12-14
feat: name registration handler

### LINKED ISSUE #405: feat: implement ENS offchain gateway with reverse lookup

geekbrother:
# Description

This PR implements the name registration handler according to the [registration SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/184).

The `verify_message_signature` function is introduced to verify the Ethereum message signature.

Resolves #405

## Stacked PRs list 🏗️

* 6: add lookup handlers #417 

## How Has This Been Tested?

Local unit tests for the `verify_message_signature` function.
Run the integration tests at https://github.com/WalletConnect/blockchain-api/pull/419

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/profile/mod.rs
+++ b/src/handlers/profile/mod.rs
@@ -1,2 +1,87 @@
+use {
+    ethers::types::H160,
+    serde::{Deserialize, Serialize},
+    std::str::FromStr,
+};
+
 pub mod lookup;
+pub mod register;
 pub mod reverse;
+
+/// Payload to register domain name that should be serialized to JSON
+/// and passed to the RegisterRequest.message
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct RegisterPayload {
+    /// Name to register
+    pub name: String,
+    /// Address
+    pub address: String,
+    /// Unixtime
+    pub timestamp: u64,
+}
+/// Data structure representing a request to register a name
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct RegisterRequest {
+    /// Serialized JSON register payload
+    pub message: String,
+    /// Message signature
+    pub signature: String,
+    /// Address
+    pub address: String,
+}
+
+#[tracing::instrument]
+pub fn verify_message_signature(
+    message: &str,
+    signature: &str,
+    owner: &H160,
+) -> Result<bool, Box<dyn std::error::Error>> {
+    let prefixed_message = format!("\x19Ethereum Signed Message:\n{}{}", message.len(), message);
+    let message_hash = ethers::core::utils::keccak256(prefixed_message.clone());
+    let message_hash = ethers::types::H256::from_slice(&message_hash);
+
+    let sign = ethers::types::Signature::from_str(signature)?;
+    match sign.verify(message_hash, *owner) {
+        Ok(_) => Ok(true),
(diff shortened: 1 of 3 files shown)


src/handlers/profile/mod.rs
```diff
@@ -1,2 +1,102 @@
...
+    let properly_escaped_string = message.replace('"', "\\\"");
+
+    // Ethereum adds a prefix to the message before signing
+    let prefixed_message = format!(
+        "\x19Ethereum Signed Message:\n{}{}",
+        properly_escaped_string.len(),
+        properly_escaped_string
+    );
+
+    // Hash the prefixed message with keccak256
```
chris13524, line 52: https://www.youtube.com/watch?v=k5DJxz6_xMA

src/handlers/profile/mod.rs
```diff
@@ -1,2 +1,102 @@
...
+pub fn verify_message_signature(
+    message: &str,
+    signature: &str,
+    owner: &H160,
+) -> Result<bool, Box<dyn std::error::Error>> {
+    // When we are passing the string with the escaped \" to the Rust string, the
+    // escape is removed and the " is passed to the verify function. This causes
+    // the signature is wrong, because the string is different.
+    // In case of passing JSON escaped string, we need to escape the " again.
+    let properly_escaped_string = message.replace('"', "\\\"");
```
chris13524, line 43: Something is wrong here, I don't think we need to or should be doing a second layer of escaping, just remove the escapes from https://github.com/WalletConnect/blockchain-api/pull/419/files#diff-29ee1e76a6651d33c3e1a5117e948675ccda1e60165c885696a9accdc1bf0bb9R169

The thing you're signing should just be JSON with no escapes:
```
\x19Ethereum Signed Message:
{"some":"json object"}
```

chris13524 (changes requested):

src/handlers/profile/mod.rs
geekbrother: Ok, I agree that this double-escaping approach smells...
I've removed it in favor of just `{"some":"json object"}`, let's see how it goes under the dogfooding (is it confusing for the devs and users, or not).

src/handlers/profile/mod.rs
```diff
@@ -1,2 +1,87 @@
...
+        let owner = H160::from_str("0xAff392551773CCb2574fAE23195CC3aFDBe98d18").unwrap();
+
+        let result = verify_message_signature(message, signature, &owner);
+        assert!(result.is_ok());
+        assert!(result.unwrap());
+    }
+
+    #[test]
+    fn test_verify_message_signature_json() {
+        let message = "{\\\"test\\\":\\\"some my text\\\"}";
```
chris13524, line 67: Raw strings can help here with the escape confusion:

```suggestion
        let message = r#"{\"test\":\"some my text\"}"#;
```

Also, shouldn't this be updated to not have the escapes?

chris13524 (approved):

src/handlers/profile/mod.rs
geekbrother: I don't mind changing it to the raw string to remove the ugly `\\\`.

> Also, shouldn't this be updated to not have the escapes? Guess it doesn't really matter because this is a generic signature validation test.

Yes, it doesn't matter what the signed string looks like, but as we are using the escaped JSON in our checks let's stick to it on test as well.

Thanks for the comments!

PR #428, opened 2023-12-19
fix: retrieve client IP from `X-Forwarded-For` header

xav:
# Description

Since the migration moved the service behind an ALB, we cannot retrieve the client IP directly for geo-location.
This leverages the `X-Forwarded-For` header injected by the LB instead.

## How Has This Been Tested?

Unit tests

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/network.rs
+++ b/src/utils/network.rs
@@ -1,4 +1,4 @@
-use {ipnet::IpNet, std::net::IpAddr};
+use {axum::http::HeaderMap, ipnet::IpNet, std::net::IpAddr};
 
 #[derive(thiserror::Error, Debug)]
 pub enum NetworkInterfaceError {
@@ -64,4 +64,12 @@ fn is_public_ip_addr(addr: IpAddr) -> bool {
     });
 
     RESERVED_NETWORKS.iter().all(|range| !range.contains(&addr))
+}
+
+pub fn get_forwarded_ip(headers: HeaderMap) -> Option<IpAddr> {
+    headers
+        .get("X-Forwarded-For")
+        .and_then(|header| header.to_str().ok())
+        .and_then(|header| header.split(',').next())
+        .and_then(|client_ip| client_ip.trim().parse::<IpAddr>().ok())
 }
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -5,6 +5,7 @@ use {
         error::RpcError,
         handlers::HistoryQueryParams,
         state::AppState,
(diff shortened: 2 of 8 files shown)


chris13524 (commented):
Might be nice to use [`axum_client_ip`](https://github.com/WalletConnect/echo-server/pull/240) instead for this?

arein (approved):
I think we use a library to abstract this away in Verify API or somewhere else - maybe @xDarksome or @chris13524 know?

But as a hotfix this shouldn't wait for that - can improve later

src/utils/network.rs
```diff
@@ -65,3 +65,11 @@ fn is_public_ip_addr(addr: IpAddr) -> bool {
 
     RESERVED_NETWORKS.iter().all(|range| !range.contains(&addr))
 }
+
+pub fn get_forwarded_ip(headers: HeaderMap) -> Option<IpAddr> {
+    headers
+        .get("X-Forwarded-For")
+        .and_then(|header| header.to_str().ok())
+        .and_then(|header| header.split(',').next())
+        .and_then(|client_ip| client_ip.trim().parse::<IpAddr>().ok())
```
geekbrother, line 74: It looks scary not to handle any errors here. Malformed `X-Forwarded-For` can lead to 500 errors without any debug information.

xDarksome:
You mean this? https://github.com/WalletConnect/verify-server/blob/main/src/http_server/mod.rs#L26
I think @xav was the one who added it :smile:

geekbrother (approved):
LGTM with the comment.

xav:
> You mean this? https://github.com/WalletConnect/verify-server/blob/main/src/http_server/mod.rs#L26 I think @xav was the one who added it 😄

Yeah, I added it but
- I forgot about it and wanted to fix the issue ASAP
- honestly, I don't really like it, it feels a bit overkill just for the LBA issue. We might need to assess how correct we want the resolved IP to be, and in the case we want it to be correct, I think we need to "map" the network to configure the resolver.

src/utils/network.rs
xav: won't the `and_then` chain default to an `Error` and ultimately to a `None` if anything goes wrong?

xav:
> Might be nice to use [`axum_client_ip`](https://github.com/WalletConnect/echo-server/pull/240) instead for this?

It's a hotfix and I think it should have the same end result. But we can overhaul this after.

PR #433, opened 2023-12-20
fix: update shared flows to `0.1.3` to fix ECS task names

xav:
# Description

Update shared CI flows to `0.1.3` to fix issue with ECS task names being build wrong.

## How Has This Been Tested?

N/A

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
We can't test it, but it looks like fix 👍 Thx.

PR #439, opened 2023-12-22
fix: back to string-based address in transactions list

geekbrother:
# Description

This PR reverting to use the string-based `address` argument in `get_transactions`. Coinbase Pay SDK is case-sensitive when passing the address to get the transactions list. When we are using the `H160` type we don't know what exactly address type was used: checksum (mixed case) or not when passing it to the Coinbase API later in the function. To fix this we need to preserve the original address string.

As a security validation for the address format, the [check for the address correct format is used](https://github.com/WalletConnect/blockchain-api/pull/439/files#diff-7a388f4d3745ebf0928c4212d627020683b440683be6530f2c885293f475e146R49) in the handler.

## How Has This Been Tested?

1. Start the server with `cargo run`.
2. Get the transactions list with the `onramp=coinbase` parameter.
3. Expect the list of the onramp transactions.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -46,7 +46,9 @@ async fn handler_internal(
     body: Bytes,
 ) -> Result<Response, RpcError> {
     let project_id = query.project_id.clone();
-    let address_hash = H160::from_str(&address).map_err(|_| RpcError::IdentityInvalidAddress)?;
+
+    // Checking for the H160 address correctness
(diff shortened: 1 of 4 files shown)


arein (approved):

chris13524:
Sounds like we need to ERC-55 format the address string to pass to Coinbase API?

PR #440, opened 2023-12-22
chore(test): integration test for the onramp Coinbase history

geekbrother:
# Description

This PR introduces the integration test for the onramp transactions history for the Coinbase provider and checks the response in case the wrong provider is requested.

## How Has This Been Tested?

* Successfully passing tests from the branch on the prod endpoint:
```
PROJECT_ID=xxx RPC_URL=https://rpc.walletconnect.com  yarn integration
``` 

* Successfully passing the `Validate` workflow from the branch.
  * [Result](https://github.com/WalletConnect/blockchain-api/actions/runs/7300031969/job/19893937065#step:5:25)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -76,9 +76,12 @@ describe('blockchain api', () => {
     })
   })
   describe('Transactions history', () => {
+    const fulfilled_address = '0x63755B7B300228254FB7d16321eCD3B87f98ca2a'
+    const empty_history_address = '0x739ff389c8eBd9339E69611d46Eec6212179BB67'
+
     it('fulfilled history', async () => {
       let resp: any = await http.get(
-        `${baseUrl}/v1/account/0xf3ea39310011333095CFCcCc7c4Ad74034CABA63/history?projectId=${projectId}`,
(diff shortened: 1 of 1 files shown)


integration/integration.test.ts
```diff
@@ -97,7 +100,27 @@ describe('blockchain api', () => {
...
+    })
+    it('onramp Coinbase provider', async () => {
+      let resp: any = await http.get(
+        `${baseUrl}/v1/account/${fulfilled_address}/history?onramp=coinbase&projectId=${projectId}`,
+      )
+      expect(resp.status).toBe(200)
+      expect(typeof resp.data.next).toBe('string')
+      expect(typeof resp.data.data).toBe('object')
+
+      const first = resp.data.data[0]
```
chris13524, line 115: Do you want to assert length to be 0 too?

chris13524 (approved):

arein (approved):

integration/integration.test.ts
geekbrother: I assume that the test will fail anyway if the size of the `resp.data.data` will be less then 1 (no data).

PR #447, opened 2024-01-05
chore(logging): reducing logging for proxy requests

geekbrother:
# Description

This PR reduces the logging noise for the proxy requests.
The number of logs for the proxy request declined from 6 records to 2 per request to include just valuable info (what provider is used for the request_id) and dropped other logs to the `DEBUG` level. 
We can turn the all logs back if we start the server with the `DEBUG` logging level.

The `provider` span variable includes the used provider name is added to the handler span for proper tracing.

## How Has This Been Tested?

1. Start the server locally by `cargo run`
2. Run the proxy query e.g. `curl 'http://localhost:3000/v1/?chainId=eip155:324&projectId=XXX'  --data-raw '{"jsonrpc":"2.0","id":0,"method":"eth_getBalance","params":["0x262f4f5DC82ad9b803680F07Da7d901D4F71d8D1","latest"]}'`

The expected logs for the request:
```
INFO http-request{method=POST request_id="a507af27-e652-4bdd-849a-12e4b050c3ae" uri=/v1/?chainId=eip155:324&projectId=XXX}:proxy{chain_id="eip155:324" provider=zkSync}: rpc_proxy::providers::zksync: close time.busy=34.0ms time.idle=3.47s

INFO http-request{method=POST request_id="a507af27-e652-4bdd-849a-12e4b050c3ae" uri=/v1/?chainId=eip155:324&projectId=XXX}: rpc_proxy: close time.busy=36.8ms time.idle=3.47s
```

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -14,7 +14,7 @@ use {
     sqlx::PgPool,
     std::sync::Arc,
     tap::TapFallible,
-    tracing::info,
+    tracing::warn,
 };
 
 pub struct AppState {
@@ -58,7 +58,7 @@ impl AppState {
         self.providers.update_weights(&self.metrics).await;
     }
 
-    #[tracing::instrument(skip(self))]
+    #[tracing::instrument(skip(self), level = "debug")]
     async fn get_project_data_validated(&self, id: &str) -> Result<ProjectData, RpcError> {
         let project = self
             .registry
@@ -68,7 +68,7 @@ impl AppState {
 
         project.validate_access(id, None).tap_err(|e| {
             self.metrics.add_rejected_project();
-            info!("Denied access for project: {id}, with reason: {e}");
+            warn!("Denied access for project: {id}, with reason: {e}");
         })?;
 
         Ok(project)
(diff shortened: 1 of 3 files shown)


arein (approved):

src/state.rs
```diff
@@ -68,7 +68,7 @@ impl AppState {
 
         project.validate_access(id, None).tap_err(|e| {
             self.metrics.add_rejected_project();
-            info!("Denied access for project: {id}, with reason: {e}");
+            warn!("Denied access for project: {id}, with reason: {e}");
```
chris13524, line 71: I don't think these should be warnings as they don't indicate a health issue with the service itself

chris13524 (approved):

src/state.rs
geekbrother: I don't mind changing it back to the `info` as nothing changed in the logs in our case (we are running with the `INFO` level). I've changed it back. Thanks for the comment!

PR #448, opened 2024-01-05
chore(logging): reducing logging for identity requests

geekbrother:
# Description

This PR reduces the logging noise for identity requests.
The number of logs for the identity request was declined twice, other logs were dropped to the `DEBUG` level.
We can turn all the logs back if we start the server with the `DEBUG` logging level.

The same is done in #447 for the proxy requests.

## How Has This Been Tested?

1. Start the server locally by `cargo run`
2. Call the identity request e.g. `curl 'http://localhost:3000/v1/identity/0xf3ea39310011333095CFCcCc7c4Ad74034CABA63?chainId=eip155%3A1&projectId=XXX'`

The expected result is that logging messages will reduced as twice.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -52,7 +52,7 @@ pub async fn handler(
         .await
 }
 
-#[tracing::instrument(skip_all)]
+#[tracing::instrument(skip_all, level = "debug")]
 async fn handler_internal(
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
LGTM

PR #457, opened 2024-01-10
fix: not returning quota limit error

xDarksome:
# Description

The quota limit being hit was only logged and metered, but the error itself wasn't being returned. :clown_face: 

## How Has This Been Tested?

Unit

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -82,7 +82,7 @@ impl AppState {
     pub async fn validate_project_access_and_quota(&self, id: &str) -> Result<(), RpcError> {
         let project = self.get_project_data_validated(id).await?;
 
-        if !project.quota.is_valid {
+        validate_project_quota(&project).tap_err(|_| {
             self.metrics.add_quota_limited_project();
             info!(
                 project_id = id,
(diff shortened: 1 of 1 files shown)


geekbrother (approved):

src/state.rs
```diff
@@ -82,16 +82,63 @@ impl AppState {
...
+            quota: Quota {
+                current: 0,
+                max: 0,
+                is_valid: true,
+            },
+        };
+
+        match super::validate_project_quota(&project) {
+            Ok(()) => {}
+            res => panic!("Invalid result: {res:?}"),
```
arein, line 135: why `panic`?

src/state.rs
```diff
@@ -82,16 +82,63 @@ impl AppState {
...
 
-        if !project.quota.is_valid {
+        validate_project_quota(&project).tap_err(|_| {
             self.metrics.add_quota_limited_project();
             info!(
                 project_id = id,
                 max = project.quota.max,
                 current = project.quota.current,
                 "Quota limit reached"
             );
```
arein, line 92: dis going to be a lot of log statements - we need to remove once we confirmed it works

arein (approved):

src/state.rs
geekbrother: I'll make a PR to remove this, but will not merge until we confirm.

PR #462, opened 2024-01-11
fix: adding temporary instrument logging for the avatar lookup

geekbrother:
# Description

Temporary enabling logging for the avatar lookup to catch the panic condition for the `ethers-rs` while it's not reproduced locally.

This change will be rolled back soon when the issue will be resolved.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -259,7 +259,7 @@ async fn lookup_name(
     )
 }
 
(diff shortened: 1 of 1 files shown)


xav (approved):

chris13524 (approved):
LGTM

PR #473, opened 2024-01-16
fix(ci): changing to invoke providers workflow as a job

### LINKED ISSUE #464: feat(ci): run providers tests only when the providers code is changed
At the moment we are running providers integration tests on all changes. 
Sometimes some of the providers are unavailable (rate-limited etc.) that results in a blocking of the CI/CD process.

We should run the provider tests only when the provider implementation is changed and run the test only for that provider that was changed.

geekbrother:
# Description

This PR changes invoking the providers test sub-workflow as a job instead of as a step, to fix the issue where the local sub-workflow files are not supported as a step.

* [The error](https://github.com/WalletConnect/blockchain-api/actions/runs/7533463980/job/20506321806#step:6:1) that should fix this PR.

Resolves #464

## How Has This Been Tested?

Invoking the sub-workflow as a job is [tested on the forked repository](https://github.com/geekbrother/blockchain-api/blob/0f3952d565c58c285367c81a98bb88533ade5aa8/.github/workflows/test_subcd.yml#L15).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/sub-providers.yml
+++ b/.github/workflows/sub-providers.yml
@@ -33,12 +33,13 @@ permissions:
 jobs:
   providers-list:
     name: "Preparing providers list"
-    runs-on: ubuntu-latest
+    runs-on:
+      group: ${{ vars.RUN_GROUP }}
     outputs:
       providers: ${{ steps.set-matrix.outputs.providers }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 2
       - name: Creating list of changed providers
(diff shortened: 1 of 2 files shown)


.github/workflows/sub-providers.yml
```diff
@@ -33,12 +33,13 @@ permissions:
...
     name: "Preparing providers list"
-    runs-on: ubuntu-latest
+    runs-on:
+      group: ${{ vars.RUN_GROUP }}
     outputs:
       providers: ${{ steps.set-matrix.outputs.providers }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
```
geekbrother, line 42: Updating to the latest version.

.github/workflows/sub-providers.yml
```diff
@@ -70,15 +71,16 @@ jobs:
   providers-test:
     name: "Run provider tests"
     needs: providers-list
-    runs-on: ubuntu-latest
+    runs-on:
```
geekbrother, line 74: Fixing the run group to be consistent across tests.

xav (approved):

PR #480, opened 2024-01-17
fix(ci): fixes to actions workflow files according to the `actionlint`

geekbrother:
# Description

This PR introduces fixes to the GitHub Actions according to the [actionlint](https://github.com/rhysd/actionlint) founded issues.

## How Has This Been Tested?

Tested only by the linter.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/sub-providers.yml
+++ b/.github/workflows/sub-providers.yml
@@ -13,7 +13,7 @@ on:
         default: 'https://staging.rpc.walletconnect.org'
   workflow_call:
     inputs:
-      providers_directory:
+      providers-directory:
         type: string
         required: true
         description: 'Directory where providers sources are located'
@@ -21,7 +21,6 @@ on:
         type: string
         required: true
         description: 'Stage RPC URL'
-        default: 'https://staging.rpc.walletconnect.org'
 
 concurrency: cd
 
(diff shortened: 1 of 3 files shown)


.github/workflows/sub-providers.yml
```diff
@@ -13,15 +13,14 @@ on:
...
-      providers_directory:
+      providers-directory:
         type: string
         required: true
         description: 'Directory where providers sources are located'
       stage-url:
         type: string
         required: true
         description: 'Stage RPC URL'
-        default: 'https://staging.rpc.walletconnect.org'
```
geekbrother, line 24: Defaults are ignored if the input is `required: true`.

.github/workflows/sub-validate.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: "Install Rust ${{ inputs.version }}"
+      - name: "Install Rust ${{ vars.RUST_VERSION }}"
```
geekbrother, line 57: We don't have `version` in inputs. Fixing.

.github/workflows/sub-providers.yml
```diff
@@ -13,17 +13,14 @@ on:
...
-
-concurrency: cd
```
geekbrother, line 26: We don't need concurency group in this workflow.

xav (approved):

PR #487, opened 2024-01-23
feat(providers): proxy request for a certain provider

geekbrother:
# Description

This PR adds an ability to request a certain provider in the proxy handler.
At the moment when we are querying the proxy for the certain `chain_id` the [weighted provider is used](https://github.com/WalletConnect/blockchain-api/blob/d093f9532a7e5dd89c964f111778aabbe375d0b8/src/handlers/proxy.rs#L60) and we can't test the request to the certain provider. 

Our integration tests uses the proxy requests and when we are requesting the certain `chain_id` the server requests a weighted provider and not the provider in tests itself. This makes tests flaky and unstable as we can't make sure what provider exactly was requested for test.

This PR making the following changes:
* Adds an optional `providerId` request parameter for the proxy handler.
* Adds [get_provider_by_provider_id](https://github.com/WalletConnect/blockchain-api/pull/487/files#diff-42c368d763b77f38b4ead84b474543a15f0d14d35e53ba7c60a9c9dae1385c7dR272) helper function to get the exact provider.
* Making changes to all provider's integration tests to use the certain provider instead of weighted.
* Allow exact provider requests only for the allowed project ID in `RPC_PROXY_TESTING_PROJECT_ID`.

`ProviderId` should be the same as in the [ProviderKind](https://github.com/WalletConnect/blockchain-api/blob/d093f9532a7e5dd89c964f111778aabbe375d0b8/src/providers/mod.rs#L290-L300).

#### Manual:

* Start the server locally by `RPC_PROXY_TESTING_PROJECT_ID=projectXXX cargo run`
* Run the curl query with the additional `providerId` parameter:

```
curl -v 'http://localhost:3000/v1/?chainId=eip155:999&projectId=projectXXX&providerId=Zora' --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}'
```

The expected result is the server log where the Zora exact provider was used and successful response:

```
{"jsonrpc":"2.0","result":"0x3e7","id":1}
```

#### CI/CD:

Integration tests are included in this PR.

## Merging requirements 🚧

* [x] Sync GitHub Actions `PROJECT_ID` with the Terraform Cloud `testing_project_id` to allow CI/CD tests suite run the exact provider requests on staging and prod.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -55,16 +55,35 @@ async fn handler_internal(
         query_params.chain_id.to_lowercase()
     };
 
-    let provider = if let Some(provider_id) = query_params.clone().provider_id {
-        match state.providers.get_provider_by_provider_id(&provider_id) {
-            None => return Err(RpcError::UnsupportedProvider(provider_id)),
-            Some(provider) => provider,
+    // Exact provider proxy request for testing suite
+    // This request is allowed only for the RPC_PROXY_TESTING_PROJECT_ID
+    let provider = match query_params.provider_id.clone() {
+        Some(provider_id) => {
+            let provider = state
+                .providers
+                .get_provider_by_provider_id(&provider_id)
+                .ok_or_else(|| RpcError::UnsupportedProvider(provider_id.clone()))?;
+
+            if let Some(ref testing_project_id) = state.config.server.testing_project_id {
+                if testing_project_id != &query_params.project_id {
+                    return Err(RpcError::InvalidParameter(format!(
+                        "The project ID {} is not allowed to use the exact provider request",
+                        query_params.project_id
+                    )));
+                }
+            } else {
+                return Err(RpcError::InvalidParameter(
+                    "RPC_PROXY_TESTING_PROJECT_ID should be configured for this type of request"
+                        .into(),
+                ));
+            }
+
+            provider
         }
-    } else {
-        state
+        None => state
             .providers
             .get_provider_for_chain_id(&chain_id)
-            .ok_or(RpcError::UnsupportedChain(chain_id.clone()))?
+            .ok_or_else(|| RpcError::UnsupportedChain(chain_id.clone()))?,
     };
 
     Span::current().record("provider", &provider.provider_kind().to_string());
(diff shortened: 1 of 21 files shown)


src/providers/mod.rs
```diff
@@ -302,6 +309,7 @@ impl Display for ProviderKind {
     }
 }
 
+#[allow(clippy::should_implement_trait)]
```
geekbrother, line 312: Cippy started to warn on this trait implementation, adding a temporary `allow` to fix this later.

geekbrother:
> Can you enable this in staging only?

I've [updated it](https://github.com/WalletConnect/blockchain-api/pull/487/commits/f752f788514d6f55afd7818827e52a8255445996) and now this request is only allowed by a certain project ID to allow only our testing suite to run it or manually when knowing the project ID that's allowed by the `RPC_PROXY_TESTING_PROJECT_ID` configuration variable. 
In this case, we can test it in both staging, prod, and manually if needed.

src/handlers/proxy.rs
```diff
@@ -55,10 +55,36 @@ async fn handler_internal(
...
+    // This request is allowed only for the RPC_PROXY_TESTING_PROJECT_ID
+    let provider = match query_params.provider_id.clone() {
+        Some(provider_id) => {
+            let provider = state
+                .providers
+                .get_provider_by_provider_id(&provider_id)
+                .ok_or_else(|| RpcError::UnsupportedProvider(provider_id.clone()))?;
+
+            if let Some(ref testing_project_id) = state.config.server.testing_project_id {
+                if testing_project_id != &query_params.project_id {
```
chris13524, line 68: sec: Not super high-risk, but should be easy to substitute this with a constant time string comparison to avoid timing side-channels. I like this one from [here](https://stackoverflow.com/a/44700409)

```rust
fn ct_compare(a: &str, b: &str) -> bool {
    a.bytes().zip(b.bytes())
        .fold(0, |acc, (a, b)| acc | (a ^ b) ) == 0
}
```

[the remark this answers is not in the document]
geekbrother: Makes sense! 
I've added the comparison by the new [constant_time_eq](https://github.com/WalletConnect/blockchain-api/pull/487/commits/d00e835c8d07a296d015e813e7954b925a3741c9) function.

geekbrother:
An integration test is added to test the exact provider request with allowed/not allowed projectId.

PR #469, opened 2024-01-15
feat(providers): adding the Aurora native mainnet and testnet RPC

geekbrother:
# Description

This PR adds the [Aurora Mainnet and Testnet RPC](https://doc.aurora.dev/getting-started/network-endpoints) as a new `Aurora` provider.

Resolves #463

## How Has This Been Tested?

1. Manually:
* Start the server locally by `cargo run`.
* Request the Aurora RPC locally by 
```
curl -X POST "http://localhost:3000/v1?chainId=eip155:1313161554&projectId=XXX" --data '{"id":"1","jsonrpc":"2.0","method":"eth_chainId","params":[]}'
```

2. CI
* [Integration test for the Aurora provider](https://github.com/WalletConnect/blockchain-api/pull/469/files#diff-526d8d8490fc58b397fe439b115b1f7b71f8812d85da7aa1b7cc0e11b5d569daR10).

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #488, opened 2024-01-23
fix(providers): adding Aurora to the init providers

geekbrother:
# Description

This PR fixing the Aurora provider. On the initial adding of the Aurora native RPC in #469 the init part was missed so the native RPC is not in the list of providers.
This issue was not caught by the compilation, lint, and our integration tests, because tests use the weighted provider and we have a few providers for the Aurora (this issue in tests is also fixed in #487 ).

## How Has This Been Tested?

Tested using the Aurora integration test in #487 per provider tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ use {
         Router,
     },
     env::{
+        AuroraConfig,
         BaseConfig,
         BinanceConfig,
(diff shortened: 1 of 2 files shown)


chris13524 (approved):
LGTM

PR #491, opened 2024-01-24
feat(ens): adding timestamp threshold check

geekbrother:
# Description

According to the [SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/184/files#diff-f64bf5c2b17c9c6d32ae9d85d9a1ce5a4aa07cb680c6f4085de94d3c4915b9d9R143) to avoid the repeating attack we should use the UNIX timestamp in the signature within the threshold interval (assuming clocks are not synced between this threshold). 
The threshold interval decreased to 10 seconds according to the discussion.

This PR implements the threshold check.

## How Has This Been Tested?

* New [unit tests](https://github.com/WalletConnect/blockchain-api/pull/491/files#diff-e382530ad7aeb0bbd6f4f0ac9bbddf4cc3944e24c387114ecc2dd640cf9e071fR108).
* Current integration tests for the name registration.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -152,7 +152,7 @@ describe('blockchain api', () => {
     const messageObject = {
         name,
         address,
-        timestamp: Date.now()
+        timestamp: Math.round(Date.now() / 1000)
     };
     const message = JSON.stringify(messageObject);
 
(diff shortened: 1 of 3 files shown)


chris13524 (approved):
LGTM

PR #494, opened 2024-01-29
feat(history): exposing dapp and chain info

geekbrother:
# Description

This PR exposes the `chain` from the [Zerion history](https://developers.zerion.io/reference/listwallettransactions) `relationships` object and the dapp info from the `attributes.application_metadata`.

Following new objects were added into the history response:

* metadata.application
  * name (Optional, String) - DApp name
  * iconUrl (Optional, String) - DApp icon url
* metadata.chain (Optional, String) - String representation of the chain from the `relationships` Zerion response in CAIP-2 format.

The example response:

```
...
{
      "id": "904c7ac7e1dc5ddd9aed4ceba06ee290",
      "metadata": {
        "operationType": "receive",
        "hash": "0x0f12fde4d498ad01baa181f433ad46ff6ce760af5bfe388418c243f2e2fb168b",
        "minedAt": "2023-12-04T11:02:25Z",
        "sentFrom": "0xd09a341d601505b0c8a0377acc962802e5f3fb65",
        "sentTo": "0x2c27e884cbe02252587ef62c5647ecb9a8936f3e",
        "status": "confirmed",
        "nonce": 76792,
        "application": { // NEW DAPP OBJECT
          "name": "USDC",
          "iconUrl": "https://token-icons.s3.amazonaws.com/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48.png"
        },
        "chain": "eip155:56" // NEW CHAIN STRING
      },
      "transfers": [
      ...
      ]
    },
...
```

The context of this task: https://walletconnect.slack.com/archives/C03SCF66K2T/p1706184453173909

## How Has This Been Tested?

1. Manual
* Start the server locally by: `cargo run`
* Run the curl query to get the transactions list: `curl -v 'http://localhost:3000/v1/account/0x63755B7B300228254FB7d16321eCD3B87f98ca2a/history?projectId=xxx' | jq .`

2. Integration tests for the fulfilled history are updated to check new objects' presence.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -89,12 +89,18 @@ describe('blockchain api', () => {
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
       
-      const first = resp.data.data[0]
-      expect(first.id).toBeDefined()
-      expect(typeof first.metadata).toBe('object')
-      expect(first.metadata.chain).toBeDefined()
-      expect(typeof first.metadata.application).toBe('object')
-      expect(typeof first.transfers).toBe('object')
+      for (const item of resp.data.data) {
+        expect(item.id).toBeDefined()
+        expect(typeof item.metadata).toBe('object')
+        // expect chain to be null or caip-2 format
+        if (item.metadata.chain !== null) {
+          expect(item.metadata.chain).toEqual(expect.stringMatching(/^(eip155:)?\d+$/));
+        } else {
+          expect(item.metadata.chain).toBeNull();
+        }
+        expect(typeof item.metadata.application).toBe('object')
+        expect(typeof item.transfers).toBe('object')
+      }
     })
     it('empty history', async () => {
       let resp: any = await http.get(
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -88,6 +88,13 @@ describe('blockchain api', () => {
       expect(resp.data.data).toHaveLength(50)
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
+      
+      const first = resp.data.data[0]
+      expect(first.id).toBeDefined()
+      expect(typeof first.metadata).toBe('object')
+      expect(first.metadata.chain).toBeDefined()
+      expect(typeof first.metadata.application).toBe('object')
+      expect(typeof first.transfers).toBe('object')
     })
     it('empty history', async () => {
       let resp: any = await http.get(
--- a/src/providers/coinbase.rs
+++ b/src/providers/coinbase.rs
@@ -127,6 +127,8 @@ impl HistoryProvider for CoinbaseProvider {
                     sent_from: "Coinbase".to_string(),
                     sent_to: address.clone(),
                     status: f.status,
+                    application: None,
+                    chain: None,
                 },
                 transfers: Some(vec![HistoryTransactionTransfer {
                     fungible_info: Some(HistoryTransactionFungibleInfo {
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -20,6 +20,7 @@ use {
             },
             portfolio::{PortfolioPosition, PortfolioQueryParams, PortfolioResponseBody},
         },
+        utils::crypto,
     },
     async_trait::async_trait,
     axum::body::Bytes,
@@ -27,7 +28,7 @@ use {
     hyper::Client,
     hyper_tls::HttpsConnector,
     serde::{Deserialize, Serialize},
-    tracing::log::error,
+    tracing::log::{error, info},
     url::Url,
 };
 
@@ -271,7 +272,19 @@ impl HistoryProvider for ZerionProvider {
                     chain: if f.relationships.chain.data.r#type != "chains" {
                         None
                     } else {
-                        Some(f.relationships.chain.data.id)
+                        // Try to convert chain id from human readable to caip2 format
+                        match crypto::string_chain_id_to_caip2_format(
+                            &f.relationships.chain.data.id,
+                        ) {
+                            Ok(chain) => Some(chain),
+                            Err(_) => {
+                                info!(
+                                    "Error on parsing chain id to caip2 format: {:?}",
+                                    f.relationships.chain.data.id
+                                );
+                                Some(f.relationships.chain.data.id)
+                            }
+                        }
                     },
                 },
                 transfers: f
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -7,8 +7,16 @@ use {
                 HistoryQueryParams,
                 HistoryResponseBody,
                 HistoryTransaction,
+                HistoryTransactionFungibleInfo,
                 HistoryTransactionMetadata,
+                HistoryTransactionMetadataApplication,
+                HistoryTransactionNFTContent,
+                HistoryTransactionNFTInfo,
+                HistoryTransactionNFTInfoFlags,
                 HistoryTransactionTransfer,
+                HistoryTransactionTransferQuantity,
+                HistoryTransactionURLItem,
+                HistoryTransactionURLandContentTypeItem,
             },
             portfolio::{PortfolioPosition, PortfolioQueryParams, PortfolioResponseBody},
         },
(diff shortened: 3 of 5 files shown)


src/providers/zerion.rs
```diff
@@ -93,7 +123,71 @@ pub struct ZerionTransactionAttributes {
     pub sent_to: String,
     pub status: String,
     pub nonce: usize,
-    pub transfers: Vec<HistoryTransactionTransfer>,
+    pub transfers: Vec<ZerionTransactionTransfer>,
```
geekbrother, line 126: Removing dependency from the `HistoryTransactionTransfer` structure here, because our response differs from the Zerion transaction response in this part since now.

src/providers/coinbase.rs
```diff
@@ -127,6 +127,8 @@ impl HistoryProvider for CoinbaseProvider {
                     sent_from: "Coinbase".to_string(),
                     sent_to: address.clone(),
                     status: f.status,
```
alanaladd21, line 129: remove dependencies

geekbrother: What exactly dependencies should be removed @alanaladd21 ?

glitch-txs (approved):

arein (changes requested):
can you make the chain id `caip25`?

glitch-txs:
let's expose it as it is now imo, then it would be great to abstract this into CAIP chains format

chris13524:
Am also in-favor of using CAIP-25 here. How will frontend consume this? I assume they'd have to do the mapping on their end anyway

glitch-txs:
> How will frontend consume this? 
on the wagmi side, Viem has a chain registry with all the chain data, so we'd map against that but we'd need to check if the naming matches which is not ideal.
This is for the Activity tab mainly

cc @tomiir

tomiir:
Honestly not sure we'd need to map anything given the chainId and the icon are present in the response and with that we could filter txs by chain and render appropiate icons/images. 

Am I missing some use-case @glitch-txs ?

glitch-txs:
Yep, filtering I meant

chris13524:
> Honestly not sure we'd need to map anything given the chainId and the icon are present in the response and with that we could filter txs by chain and render appropiate icons/images.

I presume filtering requires obtaining the chain name. What registry will you use to lookup the chain name and icon?

geekbrother:
> > Honestly not sure we'd need to map anything given the chainId and the icon are present in the response and with that we could filter txs by chain and render appropiate icons/images.
> 
> I presume filtering requires obtaining the chain name. What registry will you use to lookup the chain name and icon?

Looks like [Viem doesn't have an icon](https://github.com/wevm/viem/blob/main/src/chains/definitions/arbitrum.ts) cc @glitch-txs

chris13524:
I think @arein and I are actually referring to [CAIP-2](https://chainagnostic.org/CAIPs/caip-2) not CAIP-25.

> on the wagmi side, Viem has a [chain registry with all the chain data](https://github.com/wevm/viem/tree/main/src/chains/definitions), so we'd map against that but we'd need to check if the naming matches which is not ideal.

@glitch-txs Viem has a [method](https://rc.viem.sh/docs/utilities/extractChain.html) to find a chain by it's CAIP-2 chain ID. So blockchain API should return a CAIP-2 chain ID e.g. `eip155:1` and then the client would ensure that it has the `eip155` namespace and then pass the `reference` as the `id` parameter to `extractChain()`

glitch-txs:
returning the chain ID would be ideal as we make sure there are no conflicts with naming which sometimes can be ambiguos, e.g. in ethers devs config the chains themselves so the could you any name but it's better to rely on the chain ID which we know it's going to be more accurate.
Images are taken from the cloud with specifics ID that are keyed with chain IDs.

So:

- Ideally the best would be to get the chain ID of each transaction
- If we receive names we can use Viem registry to look for the matching name and return the ID from it, but this wouldn't work in ethers.

I really don't understand why this API returns a name in a chain_id flag xD

geekbrother:
> I really don't understand why this API returns a name in a chain_id flag xD should we report this as well? 🤔

I've asked the Zerion team in a Discord, what was the reason for this )

geekbrother:
Updates:
* Now we are trying to convert the chain id to the CAIP-2 format by the [function that uses ethers::types::Chain](https://github.com/WalletConnect/blockchain-api/pull/494/files#diff-43dc5270da6ae7bd3fbe8ee20c85643dcc3d3fb0feba36ad76570b327ef51667R38) and return the converted CAIP-2. Not all chains are there but most used are in.
* If it fails to find the chain from the human-readable list we will return the chain id as is and log an `info!` (I'll monitor these logs to find what aliases are probably missing to add it).
* Tested on a few addresses with a different history.

cc @chris13524, @glitch-txs, @tomiir, @arein

glitch-txs (approved):

src/providers/zerion.rs
```diff
@@ -271,7 +272,19 @@ impl HistoryProvider for ZerionProvider {
...
+                        match crypto::string_chain_id_to_caip2_format(
+                            &f.relationships.chain.data.id,
+                        ) {
+                            Ok(chain) => Some(chain),
+                            Err(_) => {
+                                info!(
+                                    "Error on parsing chain id to caip2 format: {:?}",
+                                    f.relationships.chain.data.id
+                                );
+                                Some(f.relationships.chain.data.id)
```
chris13524, line 285: I don't think we should fallback to an invalid value. Instead, the entry should be excluded from the results list.

chris13524 (changes requested):

src/providers/zerion.rs
geekbrother: I don't mind changing it to excluded in this case, but passing it through, we could catch the unsupported chain id at both the backend and frontend side, assuming that on the frontend side, there is a check for the CAIP-2 format, and if it's not, just behave like it's excluded.

chris13524: Why error in the client at all? We don't support the chain. We don't need to cause more errors

geekbrother: I'm not speaking about the error on the frontend side:

> assuming that on the frontend side, there is a check for the CAIP-2 format, and if it's not, just behave like it's excluded.

If there is a strong opinion not passing it through, I'll just exclude it in this case to unblock this feature asap.

geekbrother: [Just changed it](https://github.com/WalletConnect/blockchain-api/pull/494/commits/155db8ba40d19fa219a171512c29153f5847b8e7) to not pass through in case if it's not caip-2.

geekbrother:
> can you make the chain id `caip25`?

Now it's CAIP-2 compatible. cc @arein.

src/providers/zerion.rs
```diff
@@ -282,7 +282,7 @@ impl HistoryProvider for ZerionProvider {
                                     "Error on parsing chain id to caip2 format: {:?}",
                                     f.relationships.chain.data.id
                                 );
-                                Some(f.relationships.chain.data.id)
+                                None
```
chris13524, line 285: My thinking was actually that if the chain wasn't supported, then the transaction wouldn't appear in the list, not that the chain would be `null`. `null` would presumably still cause an error whereas skipping over transactions on an unsupported chain would be different. However, it's possible clients could recover by displaying a fallback image if that's all they need to do.

chris13524 (approved):

src/providers/zerion.rs
geekbrother: > However, it's possible clients could recover by displaying a fallback image if that's all they need to do.

I think that's the plan, right? cc @tomiir

integration/integration.test.ts
```diff
@@ -88,6 +88,13 @@ describe('blockchain api', () => {
       expect(resp.data.data).toHaveLength(50)
       expect(typeof resp.data.next).toBe('string')
       expect(resp.data.next).toHaveLength(80)
+      
+      const first = resp.data.data[0]
+      expect(first.id).toBeDefined()
+      expect(typeof first.metadata).toBe('object')
+      expect(first.metadata.chain).toBeDefined()
```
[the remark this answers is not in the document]
arein: Plus one this leaves a lot of room for regressions

geekbrother:
Just updated the integration test to check the `chain` is either CAIP-2 format or null.

PR #497, opened 2024-02-02
feat(ens): check for supported attributes and value format

geekbrother:
# Description

This PR adds a check for the supported name attributes to prevent registration with the attributes that are not supported or in the wrong format. The following changes are made:
* [check_attributes](https://github.com/WalletConnect/blockchain-api/commit/ab4f9b9ac9b5b579dbf3ed578c24a324fecd83d4#diff-f8a4ed25bf7d7d3b6998382ac049bed43c3275bfb07ac13bc8f4c47be087fa76R44) function is added to check keys in the attributes map to contain [only supported attributes and check the attribute value format by regexp](https://github.com/WalletConnect/blockchain-api/commit/ab4f9b9ac9b5b579dbf3ed578c24a324fecd83d4#diff-e382530ad7aeb0bbd6f4f0ac9bbddf4cc3944e24c387114ecc2dd640cf9e071fR19) for the certain attribute type.
* minor refactoring: [moved all utilities functions to the utils.rs](https://github.com/WalletConnect/blockchain-api/commit/ab4f9b9ac9b5b579dbf3ed578c24a324fecd83d4#diff-e382530ad7aeb0bbd6f4f0ac9bbddf4cc3944e24c387114ecc2dd640cf9e071fL48-L147)

## How Has This Been Tested?

* [Unit test](https://github.com/WalletConnect/blockchain-api/commit/ab4f9b9ac9b5b579dbf3ed578c24a324fecd83d4#diff-f8a4ed25bf7d7d3b6998382ac049bed43c3275bfb07ac13bc8f4c47be087fa76R134) for the new function,
* [Integration test](https://github.com/WalletConnect/blockchain-api/commit/ab4f9b9ac9b5b579dbf3ed578c24a324fecd83d4#diff-29ee1e76a6651d33c3e1a5117e948675ccda1e60165c885696a9accdc1bf0bb9R197).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/profile/mod.rs
+++ b/src/handlers/profile/mod.rs
@@ -1,19 +1,29 @@
 use {
-    ethers::types::H160,
+    once_cell::sync::Lazy,
+    regex::Regex,
     serde::{Deserialize, Serialize},
-    std::{
-        collections::HashMap,
-        str::FromStr,
-        time::{SystemTime, UNIX_EPOCH},
-    },
-    tap::TapFallible,
+    std::collections::HashMap,
 };
 
 pub mod lookup;
 pub mod register;
 pub mod reverse;
+pub mod utils;
 
 pub const UNIXTIMESTAMP_SYNC_THRESHOLD: u64 = 10;
+
+/// Attributes value max length
+const ATTRIBUTES_VALUE_MAX_LENGTH: usize = 255;
+
+/// List of supported attributes with the regex check pattern
+static SUPPORTED_ATTRIBUTES: Lazy<HashMap<String, Regex>> = Lazy::new(|| {
+    let mut map: HashMap<String, Regex> = HashMap::new();
+    map.insert(
+        "bio".into(),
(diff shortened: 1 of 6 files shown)


src/handlers/profile/mod.rs
```diff
@@ -45,103 +55,3 @@ pub struct RegisterRequest {
     /// Address
     pub address: String,
 }
-
```
geekbrother, line 48: Minor refactoring: moving to the `utils.rs` since the `mod.rs` grew significantly.

geekbrother:
> LGTM, but since the allowed list of attributes is finite I wonder if it would be better to use a `struct` for attributes (instead of a map) and then use the `validator` crate.

Yeah, probably this is a cleaner way than this one. In the case of using a struct, we should serialize it to iterate through keys as we are storing attributes in the `hstore` Postgres field and this is a map. 
Thanks, should be a followup.

PR #498, opened 2024-02-02
feat(analytics): onramp transactions history analytics

geekbrother:
# Description

This PR extracts the onramp transactions history analytics.
The following fields are included in the [onramp transaction analytics object](https://github.com/WalletConnect/blockchain-api/pull/498/files#diff-11beafba26762442363d5775947e0edd9c71eb2d24696e09eacf659725a3f07eR8-R24):

Common:

* `timestamp`
* `latency_secs`
* `lookup_address`
* `project_id`

Geo data:

* `origin`
* `region`
* `country`
* `continent`

Transaction:

* `transaction_status`: Status of the transaction:
  * `ONRAMP_TRANSACTION_STATUS_CREATED`
  * `ONRAMP_TRANSACTION_STATUS_IN_PROGRESS`
  * `ONRAMP_TRANSACTION_STATUS_SUCCESS`
  * `ONRAMP_TRANSACTION_STATUS_FAILED`
* `purchase_currency`: Crypto currency being purchased. Example: `USDC` 
* `purchase_network`: Network used to deliver crypto to the user’s wallet in CAIP-2 format. Example: `eip155:1`
* `purchase_amount`: Amount of crypto currency being purchased.

The new analytics S3 export prefix is `blockchain-api/onramp-history-lookups` with the name of `onramp-history_lookups`.

## How Has This Been Tested?

* Analytics is not tested.
* Correctness of the history endpoint response is tested by the integration test.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -1,7 +1,7 @@
 use {
     super::HANDLER_TASK_METRICS,
     crate::{
-        analytics::HistoryLookupInfo,
+        analytics::{HistoryLookupInfo, OnrampHistoryLookupInfo},
         error::RpcError,
         providers::ProviderKind,
         state::AppState,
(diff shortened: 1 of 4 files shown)


src/handlers/history.rs
```diff
@@ -183,64 +183,100 @@ async fn handler_internal(
...
+                    })
+                    .sum(),
+                origin,
+                region,
+                country,
+                continent,
+            ));
+        }
+        ProviderKind::Coinbase => {
+            for transaction in response.clone().data {
```
chris13524, line 246: Do we really want multiple analytics entries to be produced for a single request, rather than aggregating the data?

chris13524 (approved):
Don't see any problems with this, just wondering if the format is right

src/handlers/history.rs
geekbrother: We discussed this with @dnul and we want the data for each transaction without aggregating it here.

chris13524: Ok; there will be lots of duplicates though.

PR #505, opened 2024-02-06
fix(ens): name registration endpoint cleanup

geekbrother:
# Description

This PR cleans up the name registration endpoint implementation according to the [SPEC](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-server-api#register-the-name).

* Removing `coin_type`, `chain_id `, `address` from the signing message payload to remove duplicating it. 
   We have this data in the payload and the address is recovered from the signature, so we don't need to duplicate it.
* Storing the chain ID in the database in ENSIP-11 format without additional conversion from SLIP-44 to ENSIP-11 and vice versa. We are using the Ethereum namespace by default and agreed on the ENSIP-11 output format so we don't need to convert it to support future blockchains.
* Changing the name registration endpoint from `POST /v1/profile/account/:name` to `POST /v1/profile/account`. It doesn't make sense to duplicate the `name` in the query string and in the signed message payload (which is preferred).

## How Has This Been Tested?

* [Integration tests for name registration](https://github.com/WalletConnect/blockchain-api/pull/505/files#diff-28b661942cfc4edae43683709afac7c88cdd3d5bdcc47e602ee9ea4c876dbaf1)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/database/helpers.rs
+++ b/src/database/helpers.rs
@@ -1,8 +1,5 @@
 use {
-    crate::{
-        database::{error::DatabaseError, types, utils},
-        utils::crypto::convert_evm_chain_id_to_coin_type,
-    },
+    crate::database::{error::DatabaseError, types, utils},
     chrono::{DateTime, Utc},
     sqlx::{PgPool, Postgres},
     std::collections::HashMap,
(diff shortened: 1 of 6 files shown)


src/database/helpers.rs
```diff
@@ -149,17 +146,13 @@ pub async fn get_addresses_by_name(
...
-        } else {
-            convert_evm_chain_id_to_coin_type(row.chain_id.parse::<u32>().unwrap_or_default())
-        };
-
-        result_map.insert(ensip11_chain_id, types::Address {
-            address: row.address,
-            created_at: Some(row.created_at),
-        });
+        result_map.insert(
+            row.chain_id.parse::<u32>().unwrap_or_default(),
```
chris13524, line 150: Why are we parsing chain ID here? Shouldn't this already be an integer value?

`or_default()` seems like defensive instead of offensive programming.

geekbrother: We want to support other than Ethereum chains in the future and some of them use alphanumeric chain IDs. We can use the only integer as a struct field for now and add this cast later anyway.

PR #506, opened 2024-02-07
feat(providers): adding Base chain support for Pokt and Publicnode providers

geekbrother:
# Description

This PR adds Base (eip155:8453) chain support in Pokt/Grove and Publicnode providers. At the moment we are using only one provider for the Base, but we can spread the load across a few providers that support the Base chain.

* Base support for the Pokt/Grove provider,
* Base support for the Publicnode provider,
* Updating integration tests for the Pokt provider,
* Adding missed integration tests for the Publicnode provider.

## How Has This Been Tested?

* Updated integration tests in this PR.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/http/publicnode.rs
+++ b/tests/functional/http/publicnode.rs
@@ -1,0 +1,92 @@
+use {
+    super::check_if_rpc_is_responding_correctly_for_supported_chain,
+    crate::context::ServerContext,
+    rpc_proxy::providers::ProviderKind,
+    test_context::test_context,
+};
+
+#[test_context(ServerContext)]
+#[tokio::test]
+#[ignore]
+async fn publicnode_provider(ctx: &mut ServerContext) {
+    // Ethereum mainnet
+    check_if_rpc_is_responding_correctly_for_supported_chain(
+        ctx,
+        &ProviderKind::Publicnode,
+        "eip155:1",
+        "0x1",
+    )
+    .await;
+
+    // Ethereum goerli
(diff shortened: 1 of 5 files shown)


tests/functional/http/publicnode.rs
```diff
@@ -0,0 +1,92 @@
+use {
+    super::check_if_rpc_is_responding_correctly_for_supported_chain,
+    crate::context::ServerContext,
+    rpc_proxy::providers::ProviderKind,
+    test_context::test_context,
+};
+
+#[test_context(ServerContext)]
+#[tokio::test]
+#[ignore]
```
nopestack, line 10: minor: why was this ignored?

geekbrother: We are running provider tests, not all the time, but only when the provider's code is changed, and then we run only certain provider tests in our CI. [It's covered here](https://github.com/WalletConnect/blockchain-api/blob/a46cdfa208d157e773d883cb654bbc2d170c93cd/.github/workflows/sub-providers.yml#L86).

We had issues that were rate-limited by some providers and when running all tests all the time in this case it's blocking the CI/CD deployment (because the test is failing). 
So now, we are checking which provider's code was changed and running tests only for that provider. That's why it's ignored by default.

nopestack: thanks for the explanation 👍

nopestack (approved):
LGTM. Other than the question I asked it all seems pretty straightforward.

PR #507, opened 2024-02-07
fix(rpc): respond with `temporarily unavailable` when no working providers found

geekbrother:
# Description

At the moment when all provider weights for the chain are 0 because of rate limiting we are responding with the `Unsupported chain` to users which is confusing. 

This PR makes changes to respond with the `HTTP 503 Temporarily Unavailable` code and `Requested {chain_id} chain provider is temporarily unavailable` error message.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -142,13 +142,16 @@ impl ProviderRepository {
     }
 
     #[tracing::instrument(skip(self), level = "debug")]
-    pub fn get_provider_for_chain_id(&self, chain_id: &str) -> Option<Arc<dyn RpcProvider>> {
+    pub fn get_provider_for_chain_id(
+        &self,
+        chain_id: &str,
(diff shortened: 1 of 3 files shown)


arein (approved):

src/providers/mod.rs
```diff
@@ -158,11 +161,13 @@ impl ProviderRepository {
                 let random = dist.sample(&mut OsRng);
                 let provider = keys.get(random).unwrap();
 
-                self.providers.get(provider).cloned()
+                Ok(self.providers.get(provider).cloned().unwrap())
```
chris13524, line 164: How do we know `unwrap()` won't panic here? Can we refactor? I prefer to use `.expect()` to explain why the panic won't happen

geekbrother: Makes sense. I'll make this change now.

PR #509, opened 2024-02-07
feat(ens): updating name attributes handler

geekbrother:
# Description

This PR adds an endpoint handler to update name attributes according to the [SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/203).

* `POST /v1/profile/account/:name/attributes` endpoint is added to handle the name attributes updating,
  * Updated attributes map returns as a result. 
* [Integration test](https://github.com/WalletConnect/blockchain-api/pull/509/files#diff-28b661942cfc4edae43683709afac7c88cdd3d5bdcc47e602ee9ea4c876dbaf1R129-R157) for the new endpoint to check attributes after updating.

## How Has This Been Tested?

The integration test was added to this PR.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/names.test.ts
+++ b/integration/names.test.ts
@@ -18,20 +18,19 @@ describe('Account profile names', () => {
   const name = `${randomString}.connect.test`;
 
   // Create a message to sign
-  const messageObject = {
+  const registerMessageObject = {
       name,
-      coin_type,
       attributes,
       timestamp: Math.round(Date.now() / 1000)
   };
-  const message = JSON.stringify(messageObject);
+  const registerMessage = JSON.stringify(registerMessageObject);
 
   it('register with wrong signature', async () => {
     // Sign the message
     const signature = await wallet.signMessage('some other message');
 
     const payload = {
-      message,
+      message: registerMessage,
       signature,
       coin_type,
       address,
@@ -68,10 +67,10 @@ describe('Account profile names', () => {
 
   it('register new name', async () => {
     // Sign the message
-    const signature = await wallet.signMessage(message);
+    const signature = await wallet.signMessage(registerMessage);
 
     const payload = {
-      message,
+      message: registerMessage,
       signature,
       coin_type,
       address,
@@ -82,12 +81,13 @@ describe('Account profile names', () => {
     )
     expect(resp.status).toBe(200)
   })
+
   it('try register already registered name', async () => {
     // Sign the message
-    const signature = await wallet.signMessage(message);
+    const signature = await wallet.signMessage(registerMessage);
 
     const payload = {
-      message,
+      message: registerMessage,
       signature,
       coin_type,
       address,
@@ -98,17 +98,20 @@ describe('Account profile names', () => {
     )
     expect(resp.status).toBe(400)
   })
+
   it('name forward lookup', async () => {
     let resp: any = await httpClient.get(
       `${baseUrl}/v1/profile/account/${name}`
     )
     expect(resp.status).toBe(200)
     expect(resp.data.name).toBe(name)
+    expect(resp.data.attributes['bio']).toBe(attributes['bio'])
     expect(typeof resp.data.addresses).toBe('object')
     // ENSIP-11 using the 60 for the Ethereum mainnet
     const first = resp.data.addresses["60"]
     expect(first.address).toBe(address)
   })
+
   it('name reverse lookup', async () => {
     let resp: any = await httpClient.get(
       `${baseUrl}/v1/profile/reverse/${address}`
@@ -121,5 +124,35 @@ describe('Account profile names', () => {
     // ENSIP-11 using the 60 for the Ethereum mainnet
     const first_address = first_name.addresses["60"]
     expect(first_address.address).toBe(address)
+  })
+
+  it('update name attributes', async () => {
+    // Prepare updated attributes payload
+    const updatedAttributes = {
+      bio: 'integration test domain updated attribute',
+    };
+    const updateAttributesMessageObject = {
+      attributes: updatedAttributes,
+      timestamp: Math.round(Date.now() / 1000)
+    };
+    const updateMessage = JSON.stringify(updateAttributesMessageObject);
+
+    // Sign the message
+    const signature = await wallet.signMessage(updateMessage);
+
+    const payload = {
+      message: updateMessage,
+      signature,
+      coin_type,
+      address,
+    };
+    let resp: any = await httpClient.post(
+      `${baseUrl}/v1/profile/account/${name}/attributes`,
+      payload
+    );
+    
+    expect(resp.status).toBe(200)
+    expect(resp.data.name).toBe(name)
+    expect(resp.data.attributes['bio']).toBe(updatedAttributes['bio'])
   })
 })
--- a/integration/names.test.ts
+++ b/integration/names.test.ts
@@ -128,8 +128,10 @@ describe('Account profile names', () => {
 
   it('update name attributes', async () => {
     // Prepare updated attributes payload
+    const randomBioString = Array.from({ length: 24 }, 
+      () => (Math.random().toString(36)[2] || '0')).join('')
     const updatedAttributes = {
-      bio: 'integration test domain updated attribute',
+      bio: randomBioString,
     };
     const updateAttributesMessageObject = {
       attributes: updatedAttributes,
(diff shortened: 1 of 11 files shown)


src/utils/crypto.rs
```diff
@@ -55,14 +55,18 @@ pub fn string_chain_id_to_caip2_format(chain_id: &str) -> Result<String, anyhow:
     ))
 }
 
-/// Compare two strings in constant time to prevent timing attacks
-pub fn constant_time_eq(a: &str, b: &str) -> bool {
-    if a.len() != b.len() {
+/// Compare two values (either H160 or &str) in constant time to prevent timing
+/// attacks
+pub fn constant_time_eq(a: impl AsRef<[u8]>, b: impl AsRef<[u8]>) -> bool {
```
geekbrother, line 60: Making this function usable for strings and H160.

src/database/helpers.rs
```diff
@@ -68,7 +68,7 @@ pub async fn delete_name(
 }
 
 #[instrument(skip(postgres))]
-pub async fn update_name(
+pub async fn update_name_attributes(
```
geekbrother, line 71: Refactor this function name to reflect that it updates only attributes.

src/handlers/profile/attributes.rs
```diff
@@ -0,0 +1,180 @@
...
+    let payload = match serde_json::from_str::<UpdateAttributesPayload>(raw_payload) {
+        Ok(payload) => payload,
+        Err(e) => {
+            info!("Failed to deserialize register payload: {}", e);
+            return Ok((StatusCode::BAD_REQUEST, "").into_response());
+        }
+    };
+
+    // Check for the supported ENSIP-11 coin type
+    if request_payload.coin_type != 60 {
```
chris13524, line 63: Can we make 60 a constant somewhere rather than a magic number everywhere?

src/handlers/profile/attributes.rs
```diff
@@ -0,0 +1,180 @@
...
+    handler_internal(state, name, body)
+        .with_metrics(HANDLER_TASK_METRICS.with_name("profile_attributes_update"))
+        .await
+}
+
+#[tracing::instrument(skip(state))]
+pub async fn handler_internal(
+    state: State<Arc<AppState>>,
+    Path(name): Path<String>,
+    body: Bytes,
```
chris13524, line 42: Would this be simpler?
```suggestion
    Json(request_payload): Json<RegisterRequest>
```

src/handlers/profile/attributes.rs
```diff
@@ -0,0 +1,180 @@
...
+        return Ok((StatusCode::INTERNAL_SERVER_ERROR, "").into_response());
+    }
+
+    // Return the registered name, addresses and attributes
+    match get_name_and_addresses_by_name(name, &state.postgres.clone()).await {
+        Ok(response) => Ok(Json(response).into_response()),
+        Err(e) => match e {
+            SqlxError::RowNotFound => {
+                error!("New registered name is not found in the database: {}", e);
+                Ok((StatusCode::INTERNAL_SERVER_ERROR, "Name is not registered").into_response())
```
chris13524, line 172: Shouldn't a better-written update query, or a transaction, avoid this error?

src/handlers/profile/attributes.rs
geekbrother: Makes sense. I changed to use `Eip155SupportedChains` enum instead in the [followup commit](https://github.com/WalletConnect/blockchain-api/pull/509/commits/78100e960ec94ad2411a0e0cbf02674d458499eb).

src/handlers/profile/attributes.rs
geekbrother: Didn't know that. Changed it here and in the registration handler. Thanks.

src/handlers/profile/attributes.rs
geekbrother: Agree, it's also makes sense to return the updated attributes map after updating it. 
Changed this in the [follow-up commit](https://github.com/WalletConnect/blockchain-api/pull/509/commits/78100e960ec94ad2411a0e0cbf02674d458499eb#diff-9429941531a8d22a37f7502e9d2626df6dd78e7f449916a8301f1b3994ca342fR150-R155).

integration/names.test.ts
```diff
@@ -128,8 +128,10 @@ describe('Account profile names', () => {
 
   it('update name attributes', async () => {
     // Prepare updated attributes payload
+    const randomBioString = Array.from({ length: 24 }, 
+      () => (Math.random().toString(36)[2] || '0')).join('')
     const updatedAttributes = {
-      bio: 'integration test domain updated attribute',
+      bio: randomBioString,
```
geekbrother, line 134: Changes to use the random value here.

src/handlers/profile/mod.rs
```diff
@@ -1,15 +1,24 @@
...
 pub mod lookup;
 pub mod register;
 pub mod reverse;
 pub mod utils;
 
+/// List of supported Ethereum chains in ENSIP-11 format
+#[repr(u32)]
+#[derive(Debug, Clone, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+enum Eip155SupportedChains {
+    Mainnet = 60,
```
chris13524, line 19:
```suggestion
    EthereumMainnet = 60,
```

src/database/helpers.rs
```diff
@@ -68,21 +68,31 @@ pub async fn delete_name(
...
-    let insert_name_query = "
+) -> Result<HashMap<String, String>, sqlx::error::Error> {
+    let update_attributes_query = "
       UPDATE names SET attributes = $2::hstore, updated_at = NOW()
-        WHERE name = $1
+        WHERE name = $1 
+        RETURNING attributes::json
     ";
-    sqlx::query::<Postgres>(insert_name_query)
+    let result: (serde_json::Value,) = sqlx::query_as(update_attributes_query)
```
chris13524, line 81: Shouldn't this work?

```suggestion
    let result: serde_json::Value = sqlx::query_as(update_attributes_query)
```

src/database/helpers.rs
```diff
@@ -68,21 +68,31 @@ pub async fn delete_name(
...
-        // Convert JSON to String for hstore update
         .bind(&utils::hashmap_to_hstore(&attributes))
-        .execute(postgres)
-        .await
+        .fetch_one(postgres)
+        .await?;
+
+    let updated_attributes_result: Result<HashMap<String, String>, sqlx::Error> =
+        serde_json::from_value(result.0.clone()).map_err(|e| {
+            sqlx::Error::Protocol(format!(
```
chris13524, line 89: Hmm this is a [protocol error](https://docs.rs/sqlx/latest/sqlx/enum.Error.html#variant.Protocol) in communicating with the database, not an application-level error to fail to parse a value? You should use a custom error type here.

src/handlers/profile/attributes.rs
```diff
@@ -0,0 +1,160 @@
...
+    state: State<Arc<AppState>>,
+    Path(name): Path<String>,
+    request_payload: RegisterRequest,
+) -> Result<Response, RpcError> {
+    let raw_payload = &request_payload.message;
+    let payload = match serde_json::from_str::<UpdateAttributesPayload>(raw_payload) {
+        Ok(payload) => payload,
+        Err(e) => {
+            info!("Failed to deserialize update attributes payload: {}", e);
+            return Ok((StatusCode::BAD_REQUEST, "").into_response());
```
chris13524, line 49: Should there be a message here?

In general I prefer the pattern of implementing `IntoResponse` on a custom error type, rather than returning status codes everywhere.

src/database/helpers.rs
geekbrother: No, complaining about the trait. This code smells to me as well, so I [rewrote it](https://github.com/WalletConnect/blockchain-api/pull/509/commits/34bf8bbcd37bdd5cf976cdf963ecf7094b60acae). Thanks!

src/database/helpers.rs
geekbrother: Makes sense. It returns a proper error now. Thanks!

src/handlers/profile/attributes.rs
geekbrother: I missed a message here. Fixed, thanks!
Custom types are cool when they are reusable, sometimes it doesn't makes sense to make it for a single occurrence error.

PR #510, opened 2024-02-08
feat(ens): updating name address handler

geekbrother:
# Description

This PR adds an endpoint handler to update the name address according to the [SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/204).

* `POST /v1/profile/account/:name/address` endpoint is added to handle the name address updating,
  * Updated name address map returns as a result.
* [Integration test](https://github.com/WalletConnect/blockchain-api/pull/510/files#diff-28b661942cfc4edae43683709afac7c88cdd3d5bdcc47e602ee9ea4c876dbaf1R160-R199) for the new endpoint to check the name addresses after updating.

## How Has This Been Tested?

The [integration test](https://github.com/WalletConnect/blockchain-api/pull/510/files#diff-28b661942cfc4edae43683709afac7c88cdd3d5bdcc47e602ee9ea4c876dbaf1R160-R199) was added to this PR.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/profile/address.rs
+++ b/src/handlers/profile/address.rs
@@ -1,0 +1,178 @@
+use {
+    super::{
+        super::HANDLER_TASK_METRICS,
+        utils::is_timestamp_within_interval,
+        Eip155SupportedChains,
+        RegisterRequest,
+        UpdateAddressPayload,
+        UNIXTIMESTAMP_SYNC_THRESHOLD,
+    },
+    crate::{
+        database::{
+            helpers::{get_name_and_addresses_by_name, insert_or_update_address},
+            types::SupportedNamespaces,
+        },
+        error::RpcError,
+        state::AppState,
+        utils::crypto::{constant_time_eq, verify_message_signature},
+    },
+    axum::{
+        extract::{Path, State},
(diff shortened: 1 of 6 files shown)


src/handlers/profile/address.rs
```diff
@@ -0,0 +1,178 @@
...
+            "Unsupported coin type for name address update",
+        )
+            .into_response());
+    }
+
+    // Check is name registered
+    let name_addresses =
+        match get_name_and_addresses_by_name(name.clone(), &state.postgres.clone()).await {
+            Ok(result) => result,
+            Err(_) => {
```
chris13524, line 71: This can fail for reasons other than the row not existing, e.g. database error

chris13524 (changes requested):

src/handlers/profile/address.rs
geekbrother: [Fixed](https://github.com/WalletConnect/blockchain-api/pull/510/commits/3387ca929fdca3f3b7fffcda86587fca86dd699e). Thanks for catching this!

chris13524 (approved):
LGTM

PR #511, opened 2024-02-10
feat(ens): checks for allowed zones and name format

geekbrother:
# Description

This PR adds hardening checks on name registration and name address lookups. 
The following changes were added:

* `is_name_in_allowed_zones` function was added to check is the name in the allowed zones list (e.g. `wc.ink`).
* `is_name_format_correct` function was added to check the name format.
* The above checks were added for the name registration.
* The above checks were added for the name lookup.

## How Has This Been Tested?

* New unit tests for new functions.
* New integration tests to check registration and lookups with wrong zones and name format.

## Due Diligence

* [ ] Breaking change


src/handlers/profile/utils.rs
```diff
@@ -57,6 +57,23 @@ pub fn check_attributes(
...
+    if name_parts.len() != 3 {
+        return false;
+    }
+    let tld = format!("{}.{}", name_parts[1], name_parts[2]);
+    allowed_zones.contains(&tld.as_str())
+}
+
+/// Check if the given name is in the correct format
+pub fn is_name_format_correct(name: &str) -> bool {
+    let domain_regex = Regex::new(r"^[a-zA-Z0-9.-]+$").unwrap();
```
chris13524, line 72: Should use a static Lazy here to avoid recompiling the regex everytime

src/handlers/profile/register.rs
```diff
@@ -52,6 +58,18 @@ pub async fn handler_internal(
         }
     };
 
+    // Check if the name is in the correct format
+    if !is_name_format_correct(&payload.name) {
```
chris13524, line 62: Are we checking min/max length somewhere yet?

src/handlers/profile/utils.rs
```diff
@@ -161,4 +178,33 @@ mod tests {
             ATTRIBUTES_VALUE_MAX_LENGTH,
         ));
     }
+
+    #[test]
+    fn test_is_name_in_allowed_zones() {
```
chris13524, line 183: Also add test case that sub.test.wc.ink is rejected

chris13524 (approved):

nopestack (approved):

src/handlers/profile/utils.rs
geekbrother: Good catch, I missed this one. Thx!

src/handlers/profile/utils.rs
geekbrother: Makes sense. It's added now.

src/handlers/profile/register.rs
geekbrother: Want to do this in a separate PR but updated this one. [It's added](https://github.com/WalletConnect/blockchain-api/pull/511/commits/4b52256130028aea69a744b8177ebd7cf332009f) along with the unit and integration test.
Thanks a lot 🚀

PR #512, opened 2024-02-12
feat(ens): unifying errors

geekbrother:
# Description

This PR unifies errors in profile names handlers to use the `RpcError`.
This will deduplicate errors and unify them to handle in the frontend.

## How Has This Been Tested?

* Current integration tests are passed.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/profile/address.rs
+++ b/src/handlers/profile/address.rs
@@ -49,20 +49,12 @@ pub async fn handler_internal(
     let raw_payload = &request_payload.message;
     let payload = match serde_json::from_str::<UpdateAddressPayload>(raw_payload) {
         Ok(payload) => payload,
-        Err(e) => {
-            info!("Failed to deserialize update address payload: {}", e);
-            return Ok((StatusCode::BAD_REQUEST, "").into_response());
-        }
+        Err(e) => return Err(RpcError::SerdeJson(e)),
     };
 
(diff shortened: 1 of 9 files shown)


src/handlers/profile/address.rs
```diff
@@ -49,34 +49,20 @@ pub async fn handler_internal(
     let raw_payload = &request_payload.message;
     let payload = match serde_json::from_str::<UpdateAddressPayload>(raw_payload) {
         Ok(payload) => payload,
-        Err(e) => {
-            info!("Failed to deserialize update address payload: {}", e);
-            return Ok((StatusCode::BAD_REQUEST, "").into_response());
-        }
+        Err(e) => return Err(RpcError::SerdeJson(e)),
```
chris13524, line 52: SerdeJson is not mapped to a bad request

chris13524 (changes requested):

src/handlers/profile/address.rs
geekbrother: Axum responds HTTP 422 when the schema differs from the request. 
Let's also stick to that behavior in our serde json deserialization error.
Pushing that changes as https://github.com/WalletConnect/blockchain-api/pull/512/commits/d8d9724697728c0d775e6f75ff32cc401f4162c5

It should not be a 500 error as now when the request is in a bad format.

PR #514, opened 2024-02-14
feat(metrics): exporting CPU and Memory usage metrics to Prometheus

geekbrother:
# Description

This PR introduces gathering and exporting of CPU and Memory usage metrics to the Prometheus.

System CPU(s) usage and Memory usage metrics are gathered by the app tokio task once per minute and record the data into the Prometheus exported metrics.

The reason for making this change is that we want to disable AWS CloudWatch ContainerInsights for the ECS.
The full context: https://walletconnect.slack.com/archives/C068RRL89HC/p1704549567804539

### Follow-ups after this PR is merged:

* Changing Grafana panels to use new metrics for CPU and Memory.
* Disabling `ContainerInsights` in the terraform ECS config.

## How Has This Been Tested?

Not fully tested, just run it locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,9 @@ bytes = "1.4.0"
 regex = "1.10"
 sha256 = "1.2.2"
 
+# System CPU and Memory metrics
+sysinfo = "0.30"
+
 [dev-dependencies]
 jsonrpc = "0.14.0"
 test-context = "0.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2402,7 +2402,7 @@ dependencies = [
  "iana-time-zone-haiku",
  "js-sys",
  "wasm-bindgen",
- "windows-core",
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -71,6 +71,9 @@ bytes = "1.4.0"
 regex = "1.10"
 sha256 = "1.2.2"
 
+# System CPU and Memory metrics
+sysinfo = "0.30"
```
nopestack, line 75: qq did you check: https://docs.rs/sysinfo/latest/sysinfo/#good-practice--performance-tips

geekbrother: Checked it, makes sense to initialize only needed sysinfo instead of all to save some ops. Made changes to optimize it.
Thanks!

chris13524 (approved):

PR #516, opened 2024-02-16
feat(history): using the shared `reqwest` http client

geekbrother:
# Description

This PR changes to use the shared `reqwest` client for the history endpoints instead of constructing the `hyper` client each time in the handler.

This change should provide the following benefits:

* (code) Making JSON requests using the `reqwest` is much easier than the hyper itself (it handles serialization and deserialization), the code is much clearer.
* (performance) When using the shared `reqwest` HTTP client we will use the internal connection pooling and reuse it ([suggested in the reqwest docs](https://docs.rs/reqwest/latest/reqwest/struct.Client.html#) to create it once and reuse it).

## How Has This Been Tested?

Tested locally:

1. Start the server locally with the `cargo run`.
2. Run the integration tests by `RPC_URL=http://localhost:3000  yarn integration`.
3. Tests successfully passed.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ tower-http = { version = "0.4.0", features = ["cors", "trace", "request-id", "ut
 jsonrpc = "0.14.0"
 async-tungstenite = { version = "0.20.0", features = ["tokio", "tokio-runtime", "tokio-native-tls"] }
 url = "2.5"
+reqwest = { version= "0.11", features = ["deflate", "brotli", "gzip"] }
 
 # Serialization
 rmp-serde = "1.1"
--- a/src/handlers/history.rs
+++ b/src/handlers/history.rs
@@ -8,7 +8,6 @@ use {
         utils::network,
     },
     axum::{
-        body::Bytes,
         extract::{ConnectInfo, MatchedPath, Path, Query, State},
         response::{IntoResponse, Response},
         Json,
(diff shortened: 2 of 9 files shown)


src/handlers/history.rs
```diff
@@ -125,9 +124,8 @@ pub async fn handler(
     path: MatchedPath,
     headers: HeaderMap,
     address: Path<String>,
-    body: Bytes,
 ) -> Result<Response, RpcError> {
-    handler_internal(state, connect_info, query, path, headers, address, body)
```
geekbrother, line 130: We don't need to pass through `body` so it should be removed.

src/providers/coinbase.rs
```diff
@@ -85,36 +83,23 @@ impl HistoryProvider for CoinbaseProvider {
...
             .header("CBPAY-APP-ID", self.app_id.clone())
             .header("CBPAY-API-KEY", self.api_key.clone())
-            .body(hyper::body::Body::from(body))?;
+            .send()
+            .await?;
 
-        let response = self.http_client.request(hyper_request).await?;
-
-        if !response.status().is_success() {
+        if response.status() != reqwest::StatusCode::OK {
```
geekbrother, line 94: `is_success()` considering all the range of 200-299 codes are ok, but we are ok only with the 200 from the peer.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ tower-http = { version = "0.4.0", features = ["cors", "trace", "request-id", "ut
 jsonrpc = "0.14.0"
 async-tungstenite = { version = "0.20.0", features = ["tokio", "tokio-runtime", "tokio-native-tls"] }
 url = "2.5"
+reqwest = { version= "0.11", features = ["deflate", "brotli", "gzip"] }
```
geekbrother, line 28: Enabling compression for the client and responses to save on bandwidth.

src/providers/coinbase.rs
nopestack: worth checking in with the BE team

nopestack (approved):

PR #525, opened 2024-02-19
feat(onramp): adding onramp buy quotes endpoint

geekbrother:
# Description

This PR adds the new endpoint for getting buy quotes for the onramp payments from the [CB payment options endpoint](https://docs.cloud.coinbase.com/pay-sdk/docs/rest-api-reference#buy-quote).

* `/v1/onramp/buy/quotes` GET parameters:
  * `projectId` - project ID
  * `country` - [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) two-digit country code string representing the purchasing user’s country of residence, e.g., US.
  * `subdivision` (optional) - [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) two-digit country subdivision code representing the purchasing user’s subdivision of residence within their country, e.g. NY. Required if the country=“US” because certain states (e.g., NY) have state specific asset restrictions.
  * `purchaseCurrency` - ID of the crypto asset the user wants to purchase. Retrieved from the options API.
  * `purchaseNetwork` (optional) - Name of the network that the purchase currency should be purchased on. Retrieved from the options API. If omitted, the default network for the crypto currency is used.
  * `paymentAmount ` - Fiat amount the user wants to spend to purchase the crypto currency, inclusive of fees with two decimals of precision, e.g., 100.00.
  * `paymentCurrency ` - Fiat currency of the payment amount, e.g., USD.
  * `paymentMethod ` - ID of payment method used to complete the purchase. Retrieved from the options API.

The new endpoint response object contains the following fields:

* `paymentTotal ` - Object with amount and currency of the total fiat payment required to complete the purchase, inclusive of any fees. The currency will match the payment_currency in the request if it is supported, otherwise it falls back to USD.
* `paymentSubtotal ` - Object with amount and currency of the fiat cost of the crypto asset to be purchased, exclusive of any fees. The currency will match the payment_total currency.
* `purchaseAmount ` - Object with amount and currency of the crypto that to be purchased. The currency will match the purchase_currency in the request. The number of decimals will be based on the crypto asset.
* `coinbaseFee ` - Object with amount and currency of the fee changed by the Coinbase exchange to complete the transaction. The currency will match the payment_total currency.
* `networkFee ` - Object with amount and currency of the network fee required to send the purchased crypto to the user’s wallet. The currency will match the payment_total currency.
* `quoteId ` - Reference to the quote that should be passed into the initialization parameters when launching the Coinbase Pay widget via the SDK or URL generator.

## How Has This Been Tested?

1. [New integration test](https://github.com/WalletConnect/blockchain-api/pull/525/commits/af3e230558f6739e5492ba17ed90f2a5cd50ae4a) included in this PR.

2. Manual. 

  * Start the server locally: `cargo run`.
  * Run the following curl request: `http://localhost:3000/v1/onramp/buy/quotes?purchaseCurrency=BCH&paymentAmount=100.00&paymentCurrency=USD&paymentMethod=CARD&country=US&projectId=XXX' | jq .`

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/coinbase.rs
+++ b/src/providers/coinbase.rs
@@ -12,7 +12,10 @@ use {
                 HistoryTransactionTransfer,
                 HistoryTransactionTransferQuantity,
             },
-            onramp::options::{OnRampBuyOptionsParams, OnRampBuyOptionsResponse},
+            onramp::{
+                options::{OnRampBuyOptionsParams, OnRampBuyOptionsResponse},
+                quotes::{OnRampBuyQuotesParams, OnRampBuyQuotesResponse},
+            },
         },
         utils::crypto::string_chain_id_to_caip2_format,
     },
@@ -177,5 +180,33 @@ impl OnRampProvider for CoinbaseProvider {
         }
 
         Ok(response.json::<OnRampBuyOptionsResponse>().await?)
+    }
+
+    async fn get_buy_quotes(
+        &self,
+        params: OnRampBuyQuotesParams,
+        http_client: reqwest::Client,
+    ) -> RpcResult<OnRampBuyQuotesResponse> {
+        let base = format!("{}/buy/quote", &self.base_api_url);
+        let url = Url::parse(&base).map_err(|_| RpcError::OnRampParseURLError)?;
+
+        let response = http_client
+            .post(url)
+            .json(&params)
+            .header("Content-Type", "application/json")
+            .header("CBPAY-APP-ID", self.app_id.clone())
+            .header("CBPAY-API-KEY", self.api_key.clone())
+            .send()
+            .await?;
+
+        if response.status() != reqwest::StatusCode::OK {
+            error!(
+                "Error on CoinBase buy quotes response. Status is not OK: {:?}",
+                response.status(),
+            );
+            return Err(RpcError::OnRampProviderError);
+        }
+
+        Ok(response.json::<OnRampBuyQuotesResponse>().await?)
     }
 }
--- a/integration/onramp.test.ts
+++ b/integration/onramp.test.ts
@@ -3,10 +3,12 @@ import { getTestSetup } from './init';
 describe('OnRamp', () => {
   const { baseUrl, projectId, httpClient } = getTestSetup();
   const onRampPath = `${baseUrl}/v1/onramp`;
+  const country = 'US';
+  const subdivision = 'NY';
 
   it('buy options', async () => {
     let resp: any = await httpClient.get(
-      `${onRampPath}/buy/options?projectId=${projectId}&country=US&subdivision=NY`,
+      `${onRampPath}/buy/options?projectId=${projectId}&country=${country}&subdivision=${subdivision}`,
     )
     expect(resp.status).toBe(200)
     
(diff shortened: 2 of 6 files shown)


src/providers/coinbase.rs
```diff
@@ -178,4 +181,32 @@ impl OnRampProvider for CoinbaseProvider {
...
+    ) -> RpcResult<OnRampBuyQuotesResponse> {
+        let base = format!("{}/buy/quote", &self.base_api_url);
+        let url = Url::parse(&base).map_err(|_| RpcError::OnRampParseURLError)?;
+
+        let response = http_client
+            .post(url)
+            .json(&params)
+            .header("Content-Type", "application/json")
+            .header("CBPAY-APP-ID", self.app_id.clone())
+            .header("CBPAY-API-KEY", self.api_key.clone())
```
chris13524, line 198: Have you considered re-using the code that generates these headers?

src/providers/coinbase.rs
```diff
@@ -178,4 +181,32 @@ impl OnRampProvider for CoinbaseProvider {
...
+        params: OnRampBuyQuotesParams,
+        http_client: reqwest::Client,
+    ) -> RpcResult<OnRampBuyQuotesResponse> {
+        let base = format!("{}/buy/quote", &self.base_api_url);
+        let url = Url::parse(&base).map_err(|_| RpcError::OnRampParseURLError)?;
+
+        let response = http_client
+            .post(url)
+            .json(&params)
+            .header("Content-Type", "application/json")
```
chris13524, line 196: I think this header is not needed because you are using `.json()`

chris13524 (commented):
Looks like we are returning Coinbase's response as-is and not modifying it to e.g. make more generic. Do we need all of these fields as-is?

geekbrother:
> Looks like we are returning Coinbase's response as-is and not modifying it to e.g. make more generic. Do we need all of these fields as-is?

We don't have any other onramp providers now and are using CB as a reference. 
I think @tomiir needs all of these.

tomiir:
@chris13524 @geekbrother it'd be great to have an onramp abstraction but there's no list yet on other providers. 
If possible I'd prefer to keep as is while we don't have more info 🙏 

However I think in the future we'd need to discuss:
- Transaction history and possibility of unifying all history: each provider will probably have their own txs to query and we won't be able to scale and support all of their interfaces
- Pay URL generation: possibly need to abstract parameters to send
- Options / Quotes: we don't even know if they are supported but same case.

chris13524 (approved):

src/providers/coinbase.rs
geekbrother: Makes sense, I'll move it to the http client creation in a follow up. Thanks!

src/providers/coinbase.rs
geekbrother: Yep, [you are right it's added when using the json](https://github.com/seanmonstar/reqwest/blob/e639bdc1264c266a84dae3bae3193e07cae6861b/src/async_impl/request.rs#L445). I'll remove this in a chore followup to clean up everywhere.

PR #527, opened 2024-02-20
fix(server): increasing keep-alive timeouts

geekbrother:
# Description

We have a rare 502 errors coming from the AWS Load Balancer. 
There are no 5xx server errors or any crashes on the app side. Investigation of the LB access logs and traffic dump showed that the 502 error was caused due to the TCP issue. Further investigation shows that the issue can be in the very low keep-alive timeout in our axum server. 
There are a few [#1](https://www.tessian.com/blog/how-to-fix-http-502-errors/), [#2](https://adamcrowder.net/posts/node-express-api-and-aws-alb-502/) same issues found and possible solution is to increase the keep-alive to the AWS ALB default value of 60 seconds.

This PR increases keep-alive values to 65 seconds to fix the 502 Load Balancer errors.

Context: https://walletconnect.slack.com/archives/C03SCF66K2T/p1707949790838309

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,8 +71,8 @@ use {
 };
 
 const SERVICE_TASK_TIMEOUT: Duration = Duration::from_secs(15);
-const KEEPALIVE_IDLE_DURATION: Duration = Duration::from_secs(5);
-const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(5);
+const KEEPALIVE_IDLE_DURATION: Duration = Duration::from_secs(65);
+const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(65);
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
Some context of why these keepalives were set to 5s, there is a lot going on here: https://walletconnect.slack.com/archives/C03SCF66K2T/p1689430536471329

It's possible the source of the 502s is actually the service task timeout which does _not_ gracefully close the connections. There's an open issue for this: #286 

If you merge this keep an eye on the memory consumption. We don't want clients to be connected for too long and have keepalives or else they will use memory. But I think the service task timeout is actually what closes the connections to keep the memory down.

As the service task timeout is set to 15s, the keepalive will actually not be 65s

PR #528, opened 2024-02-21
feat(monitoring): changing to use Prometheus data for the CPU usage

geekbrother:
# Description

This PR is changing to use Prometheus data for the CPU usage in Grafana instead of the AWS CloudWatch Container Insights data which should be disabled soon.

* Data source and query are changed
* Alert for the CPU usage greater than `70` is changed.

This is a followup for #514 

## How Has This Been Tested?

It is tested manually by putting the query into the Grafana UI.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #530, opened 2024-02-21
feat(monitoring): disabling container insights

geekbrother:
# Description

This PR disables AWS CloudWatch Container Insights for the ECS after the CPU #528  and Memory #529 metrics are migrated to the Prometheus and we don't need the Container Insights anymore.

* ~~Bump `ci_workflows` tag to `0.2.6`,~~
* Changing ECS `containerInsights` to `disable`.

## How Has This Been Tested?

Not tested.

## Merging dependency policy 🚧

* [x] https://github.com/WalletConnect/ci_workflows/pull/5
* [x] #528
* [x] #529
* [x] Check is it hurt the autoscaling (by checking it on staging).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/ecs/cluster.tf
+++ b/terraform/ecs/cluster.tf
@@ -43,9 +43,10 @@ resource "aws_ecs_cluster" "app_cluster" {
   }
 
   # Exposes metrics such as the number of running tasks in CloudWatch
+  # Should be disabled because we use Prometheus for CPU and Memory monitoring
   setting {
     name  = "containerInsights"
-    value = "enabled"
+    value = "disabled"
   }
 }
 


terraform/ecs/cluster.tf
```diff
@@ -37,10 +37,11 @@ resource "aws_ecs_cluster" "app_cluster" {
     }
   }
 
-  # Exposes metrics such as the number of running tasks in CloudWatch
+  # Exposes metrics such as Network rx/tx, CPU and Memory usage in CloudWatch
+  # Should be disabled because we use Prometheus for CPU and Memory monitoring
   setting {
     name  = "containerInsights"
-    value = "enabled"
+    value = "disabled"
```
chris13524, line 44: What does autoscaling trigger on then? https://github.com/WalletConnect/blockchain-api/blob/cd66e9cebf9a1f9f91b009e7690a0217c46355a7/terraform/ecs/cluster_autoscaling.tf#L18

nopestack (approved):

terraform/ecs/cluster.tf
geekbrother: Disabling `ContainerInsights` should not hurt the Fargate autoscaling, because for the autoscaling the `AWS Predefined Scaling Metric` is used which is different from the `ContainerInsights`.
We are using [target_tracking_scaling_policy_configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appautoscaling_policy#target_tracking_scaling_policy_configuration-predefined_metric_specification) with the [ECSServiceAverageCPUUtilization | ECSServiceAverageMemoryUtilization](https://docs.aws.amazon.com/autoscaling/plans/APIReference/API_PredefinedScalingMetricSpecification.html).

PR #531, opened 2024-02-21
fix(errors): unifying provider reachability errors to the 503 HTTP Error

geekbrother:
# Description

This PR unifies provider reachability errors to the 503 `Service Temporary Unavailable` instead of 502 `Bad Gateway` which is the Load Balancer error. 
This change should distinguish app third-party providers errors from AWS ALB errors, that return 502 status.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -170,7 +170,7 @@ impl IntoResponse for RpcError {
             )
                 .into_response(),
             Self::ProviderError => (
-                StatusCode::BAD_GATEWAY,
+                StatusCode::SERVICE_UNAVAILABLE,
                 Json(new_error_response(
                     "unreachable".to_string(),
                     "We failed to reach the provider for your request".to_string(),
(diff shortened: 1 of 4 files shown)


tests/functional/http/mod.rs
```diff
@@ -38,7 +38,6 @@ async fn check_if_rpc_is_responding_correctly_for_supported_chain(
     let (status, rpc_response) = send_jsonrpc_request(client, addr, chaind_id, request).await;
 
     match status {
-        StatusCode::BAD_GATEWAY => {}
```
geekbrother, line 41: We shouldn't act any status codes other than `OK` as a pass to the provider's test. I think it was added sometimes to pass the CI test in case the provider is rate-limited. 
At the moment we have per-provider tests.

src/error.rs
```diff
@@ -194,7 +194,7 @@ impl IntoResponse for RpcError {
             )
                 .into_response(),
             Self::Throttled => (
```
chris13524, line 196: This error variant is never used, perhaps we should remove?

chris13524 (approved):

src/error.rs
geekbrother: Removed. Thanks!

PR #536, opened 2024-02-26
fix(monitoring): using 3 minutes CPU average

geekbrother:
# Description

This PR changes the CPU utilization chart and alerts to use 3-minute average instead of 1 minute. 
In 1-minute intervals, there is no average data because we are getting the CPU utilization once per minute.

## How Has This Been Tested?

Put the new configuration manually to the Grafana.

## Due Diligence

* [ ] Breaking change


arein (changes requested):
It appears we should use `max` and an interval that allows for the autoscaling to kick in. See for example what we use in the relay

arein:
> This change is for the chart and alerts only.

That's what I mean. Because we have autoscaling I don't think we need an alert on CPU that promply. But we can iterate here.

arein (approved):

geekbrother:
> That's what I mean. Because we have autoscaling I don't think we need an alert on CPU that promply. But we can iterate here.

Makes sense! We should allow auto-scaling provisioning without noisy alerting (while it's in progress). 
Thanks for the clarification!

geekbrother:
> > That's what I mean. Because we have autoscaling I don't think we need an alert on CPU that promply. But we can iterate here.
> 
> Makes sense! We should allow auto-scaling provisioning without noisy alerting (while it's in progress). Thanks for the clarification!

[Changes were added](https://github.com/WalletConnect/blockchain-api/pull/536/commits/826537526a54bc84852f82d11d13348d81ab7a61). 
PR description is updated.

PR #539, opened 2024-02-27
fix(monitoring): increasing CPU alarm threshold

geekbrother:
# Description

This PR increases the CPU utilization alarm threshold to 25 minutes same as we have in the Relay to allow autoscaling provisioning and alarm if it's not scaled along with the high CPU usage.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/monitoring/panels/ecs/cpu.libsonnet
+++ b/terraform/monitoring/panels/ecs/cpu.libsonnet
@@ -18,7 +18,7 @@ local overrides       = defaults.overrides;
       namespace     = 'RPC Proxy',
       name          = "RPC %s - High CPU usage" % vars.environment,
       message       = "RPC %s - High CPU usage" % vars.environment,
-      period        = '15m',
+      period        = '25m',
       frequency     = '3m',
       noDataState   = 'alerting',
       notifications = vars.notifications,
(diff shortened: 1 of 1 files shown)


nopestack (approved):

terraform/monitoring/panels/ecs/cpu.libsonnet
```diff
@@ -31,7 +31,7 @@ local overrides       = defaults.overrides;
           evaluatorType   = 'gt',
           operatorType    = 'or',
           queryRefId      = 'CPU_Avg',
-          queryTimeStart  = '15m',
+          queryTimeStart  = '25m',
```
chris13524, line 34: Changing the query start take makes it still a no-op

chris13524 (changes requested):

terraform/monitoring/panels/ecs/cpu.libsonnet
geekbrother: [We have the same parameters in Relay](https://github.com/WalletConnect/rs-relay/blob/67be2244b8950117eafe429467ef6df2b980b009/terraform/monitoring/panels/app/app_cpu_memory.libsonnet#L20-L33). I'm assuming that the query time start and period should be the same to trigger on the same timeframe.

chris13524 (approved):

PR #547, opened 2024-03-01
fix(providers): `Near` chain ID to be CAIP-2 compatible

geekbrother:
# Description

This PR changes the chain Id for the `Near` to be CAIP-2 compatible and becomes `near:mainnet` instead of `near`.
According to the Grafana chain Ids requests, at the moment we don't have any Near traffic yet (we've added it recently), so we can switch it without migration.

## How Has This Been Tested?

Integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/SUPPORTED_CHAINS.md
+++ b/SUPPORTED_CHAINS.md
@@ -24,7 +24,7 @@ Chain name with associated `chainId` query param to use.
 - Avalanche Fuji Testnet (`eip155:43113`)
 - zkSync Era (`eip155:324`)
 - zkSync Era Testnet (`eip155:280`)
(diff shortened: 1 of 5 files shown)


chris13524 (approved):
Technically there should be a CAIP-2 entry in [ChainAgnostic/namespaces](https://github.com/ChainAgnostic/namespaces) but I see an [abandoned PR](https://github.com/ChainAgnostic/CAIPs/pull/43) which matches what you've done here so approving 👍 

Opening this issue for standardizing this: https://github.com/ChainAgnostic/namespaces/issues/106

PR #563, opened 2024-03-04
feat: account balance endpoint

geekbrother:
# Description

This PR adds the new account balance endpoint `/v1/account/{address}/balance` according to the [specification](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-server-api#balance).

* A new `balance` provider trait was added,
* Zerion was added as a first account balance provider,
* Handler for the endpoint was implemented https://github.com/WalletConnect/blockchain-api/pull/563/commits/1dcd06ef90d7667da0d94687c57e717d177f3d1e,
* Fixing CAIP-2 to human readable chain IDs converter by using own enum list instead of Ethers-rs to be compatible with the Zerions chain-ids and Coinbase provider formats that is used for the filtering https://github.com/WalletConnect/blockchain-api/pull/563/commits/750c95f140dd78a1e85235277f65fb6c63cbf6d2,
* Integration tests for the new endpoint were added. https://github.com/WalletConnect/blockchain-api/pull/563/commits/d4d6b6b3e6119de00e3d9af0e8f500b508acbd60.

## How Has This Been Tested?

[New integration tests](https://github.com/WalletConnect/blockchain-api/pull/563/commits/d4d6b6b3e6119de00e3d9af0e8f500b508acbd60) and [updated unit tests](https://github.com/WalletConnect/blockchain-api/pull/563/files#diff-43dc5270da6ae7bd3fbe8ee20c85643dcc3d3fb0feba36ad76570b327ef51667R187-R220) are included in this PR.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/balance.test.ts
+++ b/integration/balance.test.ts
@@ -1,6 +1,6 @@
 import { getTestSetup } from './init';
 
-describe('Address balance', () => {
+describe('Account balance', () => {
   const { baseUrl, projectId, httpClient } = getTestSetup();
 
   const fulfilled_address = '0xf3ea39310011333095CFCcCc7c4Ad74034CABA63'
--- a/integration/balance.test.ts
+++ b/integration/balance.test.ts
@@ -1,0 +1,36 @@
+import { getTestSetup } from './init';
+
+describe('Address balance', () => {
+  const { baseUrl, projectId, httpClient } = getTestSetup();
+
+  const fulfilled_address = '0xf3ea39310011333095CFCcCc7c4Ad74034CABA63'
+  const empty_address = '0x5b6262592954B925B510651462b63ddEbcc22eaD'
+  const currency = 'usd'
+
+  it('fulfilled balance', async () => {
+    let resp: any = await httpClient.get(
+      `${baseUrl}/v1/account/${fulfilled_address}/balance?projectId=${projectId}&currency=${currency}`
+    )
+    expect(resp.status).toBe(200)
+    expect(typeof resp.data.balances).toBe('object')
+    expect(resp.data.balances.length).toBeGreaterThan(1)
+
+    for (const item of resp.data.balances) {
+      expect(typeof item.name).toBe('string')
+      expect(typeof item.symbol).toBe('string')
+      expect(item.chainId).toEqual(expect.stringMatching(/^(eip155:)?\d+$/))
+      expect(typeof item.price).toBe('number')
+      expect(typeof item.quantity).toBe('object')
+      expect(typeof item.iconUrl).toBe('string')
+    }
+  })
+
+  it('empty balance', async () => {
+    let resp: any = await httpClient.get(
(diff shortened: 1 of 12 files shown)


integration/history.test.ts
```diff
@@ -4,7 +4,7 @@ describe('Transactions history', () => {
   const { baseUrl, projectId, httpClient } = getTestSetup();
 
   const fulfilled_address = '0x63755B7B300228254FB7d16321eCD3B87f98ca2a'
-  const empty_history_address = '0x739ff389c8eBd9339E69611d46Eec6212179BB67'
+  const empty_history_address = '0x5b6262592954B925B510651462b63ddEbcc22eaD'
```
geekbrother, line 7: The old "empty" address for tests became non-empty (it has a transactions history now and balance), so I've updated it to the new one that's empty on history and has an empty balance.

integration/balance.test.ts
```diff
@@ -0,0 +1,36 @@
+import { getTestSetup } from './init';
+
+describe('Account balance', () => {
+  const { baseUrl, projectId, httpClient } = getTestSetup();
+
+  const fulfilled_address = '0xf3ea39310011333095CFCcCc7c4Ad74034CABA63'
+  const empty_address = '0x5b6262592954B925B510651462b63ddEbcc22eaD'
+  const currency = 'usd'
+
+  it('fulfilled balance', async () => {
```
nopestack, line 10: minor issue: consider adding a more descriptive title to the test such as "should retrieve an account's balance with funds"

integration/balance.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      expect(typeof item.name).toBe('string')
+      expect(typeof item.symbol).toBe('string')
+      expect(item.chainId).toEqual(expect.stringMatching(/^(eip155:)?\d+$/))
+      expect(typeof item.price).toBe('number')
+      expect(typeof item.quantity).toBe('object')
+      expect(typeof item.iconUrl).toBe('string')
+    }
+  })
+
+  it('empty balance', async () => {
```
nopestack, line 28: minor issue: consider adding a more descriptive title to the test such as "should retrieve an account's balance with no funds"

integration/balance.test.ts
geekbrother: It makes sense, but this is a test title.
When we are running the test the result looks like this:
```
Account balance
    ✓ fulfilled balance (1124 ms)
    ✓ empty balance (221 ms)
```
In case of using the long description it will look longer then actually needed without adding a context here.
Maybe it's better to add the test description as a comment instead of making the title too long.

nopestack (approved):

integration/balance.test.ts
nopestack: if this format is the case for other tests as well then yeah it makes sense to just optionally add said context as a comment though the code also is fairly legible so not urgent.

chris13524 (approved):
LGTM

PR #565, opened 2024-03-05
fix: correct Solana chain ID

chris13524:
# Description

Updates the Solana chain ID to be correct. [Slack conversation](https://walletconnect.slack.com/archives/C04SYT4LU2W/p1709656775702379?thread_ts=1709319338.504139&cid=C04SYT4LU2W)

Also removes `solana-mainnet` alias as there is no indication of traffic using this chain ID for last 30 days.

Removes `to_lowercase()` on chain ID as [chain IDs are case-sensitive](https://chainagnostic.org/CAIPs/caip-2#syntax).

## How Has This Been Tested?

New test

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/http/pokt.rs
+++ b/tests/functional/http/pokt.rs
@@ -188,7 +188,7 @@ async fn pokt_provider_solana_mainnet(ctx: &mut ServerContext) {
 #[test_context(ServerContext)]
 #[tokio::test]
 #[ignore]
-async fn pokt_provider_solana_mainnet_incorrect(ctx: &mut ServerContext) {
+async fn pokt_provider_solana_mainnet_non_standard(ctx: &mut ServerContext) {
     let addr = format!(
         "{}/v1?projectId={}&chainId=",
         ctx.server.public_addr, ctx.server.project_id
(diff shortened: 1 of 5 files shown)


tests/functional/http/pokt.rs
```diff
@@ -167,6 +167,41 @@ async fn pokt_provider_solana_mainnet(ctx: &mut ServerContext) {
...
+    assert!(rpc_response.error.is_none());
+
+    // Check chainId
+    assert_eq!(rpc_response.result::<String>().unwrap(), "ok")
+}
+
+#[test_context(ServerContext)]
+#[tokio::test]
+#[ignore]
+async fn pokt_provider_solana_mainnet_incorrect(ctx: &mut ServerContext) {
```
geekbrother, line 191: My first assumption for this test name was that we are testing incorrect request/response for the Solana in Pokt. Let's change this name to the `pokt_provider_solana_mainnet_legacy` or `pokt_provider_solana_mainnet_deprecated`.

geekbrother (approved):
LGTM. No traffic at all for `solana-mainnet` so we are safe to not migrate it. Thanks for this PR.

tests/functional/http/pokt.rs
chris13524, line 191:
```suggestion
async fn pokt_provider_solana_mainnet_non_standard(ctx: &mut ServerContext) {
```

chris13524: Changed to `non_standard` since no indication people will stop using it.

PR #570, opened 2024-03-07
feat(conversion): approve transaction endpoint implementation

geekbrother:
# Description

This PR implements `/v1/convert/build-approve` endpoint to build the approval transaction for eip2612 incompatible tokens swap (single chain) according to the [API SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/207/files#diff-f64bf5c2b17c9c6d32ae9d85d9a1ce5a4aa07cb680c6f4085de94d3c4915b9d9R334-R364).

* `1inch` provider implementation updates for the approval transaction agnostic trait,
* New endpoint handler implementation.

*This is an Alpha release implementation to support only single-chain swaps*

## How Has This Been Tested?

* [New integration test](https://github.com/WalletConnect/blockchain-api/pull/570/files#diff-d2a40c4ae4e4965cd2ea19db3965e98a04e70177954587a2a3edecfcf1263ce0) for the endpoint.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/convert/approve.rs
+++ b/src/handlers/convert/approve.rs
@@ -47,7 +47,7 @@ pub async fn handler(
     query: Query<ConvertApproveQueryParams>,
 ) -> Result<Response, RpcError> {
     handler_internal(state, query)
-        .with_metrics(HANDLER_TASK_METRICS.with_name("tokens_list"))
+        .with_metrics(HANDLER_TASK_METRICS.with_name("convert_approve_tx"))
         .await
 }
 
--- a/src/handlers/convert/approve.rs
+++ b/src/handlers/convert/approve.rs
@@ -1,0 +1,73 @@
+use {
+    super::super::HANDLER_TASK_METRICS,
+    crate::{error::RpcError, state::AppState},
+    axum::{
+        extract::{Query, State},
+        response::{IntoResponse, Response},
+        Json,
+    },
+    serde::{Deserialize, Serialize},
+    std::sync::Arc,
+    tap::TapFallible,
+    tracing::log::error,
+    wc::future::FutureExt,
+};
+
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct ConvertApproveQueryParams {
+    pub project_id: String,
+    pub amount: usize,
+    pub from: String,
+    pub to: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct ConvertApproveResponseBody {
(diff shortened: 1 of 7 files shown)


src/providers/one_inch.rs
```diff
@@ -165,4 +180,58 @@ impl ConversionProvider for OneInchProvider {
...
+        params: ConvertApproveQueryParams,
+    ) -> RpcResult<ConvertApproveResponseBody> {
+        let chain_id = crypto::disassemble_caip10(&params.from)?.1;
+        let (_, dst_chain_id, dst_address) = crypto::disassemble_caip10(&params.to)?;
+
+        // Check if from and to chain ids are different
+        // 1inch provider does not support cross-chain swaps
+        if dst_chain_id != chain_id {
+            return Err(RpcError::InvalidParameter(
+                "from and to chain ids are different in a single chain swap".into(),
```
chris13524, line 195: What makes it a single chain swap?

```suggestion
                "Swapping between chains is currently not supported".into(),
```

src/providers/one_inch.rs
```diff
@@ -165,4 +180,58 @@ impl ConversionProvider for OneInchProvider {
...
+            ));
+        }
+
+        let base = format!(
+            "{}/{}/approve/transaction",
+            &self.base_api_url,
+            chain_id.clone()
```
chris13524, line 202: You shouldn't have to clone this

```suggestion
            chain_id
```

src/handlers/convert/approve.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct ConvertApproveTxEip155 {
+    pub gas_price: String,
+}
+
+pub async fn handler(
+    state: State<Arc<AppState>>,
+    query: Query<ConvertApproveQueryParams>,
+) -> Result<Response, RpcError> {
+    handler_internal(state, query)
+        .with_metrics(HANDLER_TASK_METRICS.with_name("tokens_list"))
```
chris13524, line 50: Should `tokens_list` be something different?

chris13524 (approved):

src/providers/one_inch.rs
geekbrother: We will use one inch for only single swap transactions, so this should be better filtered in the handler function first and this case should not occur at the provider itself. I'll make a follow up to move this check at the provider choice at the handler. Thanks for raising this!

src/handlers/convert/approve.rs
geekbrother: Thanks for catching it. This is a nit. Fixed.

PR #571, opened 2024-03-08
fix: supported chains endpoint

### LINKED ISSUE #566: API to pull list of supported chains
[Slack conversation](https://walletconnect.slack.com/archives/C04SYT4LU2W/p1709658823506369?thread_ts=1709319338.504139&cid=C04SYT4LU2W)

chris13524:
# Description

Adds a basic endpoint to return a list of supported CAIP-2 chain IDs.

Resolves #566 

## How Has This Been Tested?

New integration test

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -20,6 +20,7 @@ use {
     axum_tungstenite::WebSocketUpgrade,
     hyper::http::HeaderValue,
     rand::{distributions::WeightedIndex, prelude::Distribution, rngs::OsRng},
+    serde::{Deserialize, Serialize},
     std::{
         collections::{HashMap, HashSet},
         fmt::{Debug, Display},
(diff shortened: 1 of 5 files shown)


src/providers/mod.rs
```diff
@@ -159,6 +161,7 @@ impl ProviderRepository {
         ));
 
         Self {
+            supported_chains: HashSet::new(),
```
nopestack, line 164: Consider simply calling `..Default::default` to avoid setting all of these empty HashSets/Maps

nopestack (approved):
Lgtm

geekbrother (changes requested):
Requesting changes, because I see an issue here:
* Why only ws supported chains were added? In this case, the return should reflect this somehow, e.g. move this list into the `ws` entity of the response.

The test:
`curl -v 'http://localhost:3000/v1/supported-chains?projectId=xxx' | jq .`, the response:
```
[
  "eip155:420",
  "eip155:1",
  "eip155:5",
  "eip155:42161",
  "eip155:10",
  "eip155:11155111",
  "eip155:421613",
  "eip155:7777777"
]
```
There are only ws supported chains.

I think this endpoint should reflect the `SUPPORTED_CHAINS.md`.

chris13524:
@geekbrother oops! Yeah meant to do HTTP RPC not WS supported chains. Good catch.

I suppose I'll put the list of HTTP RPC supported chains in an `http` entity of the response to allow room to support `ws` in the future? WS is deprecated and will probably be removed. Maybe I'll just put both `http` and `ws` in there, might as well and easy enough and models current structure of everything.

src/providers/mod.rs
chris13524: `Self` doesn't implement `Default`

geekbrother (approved):
Nice!

PR #572, opened 2024-03-08
feat(conversion): convert transaction builder endpoint implementation

geekbrother:
# Description

This PR implements `/v1/convert/build-transaction` endpoint to build the conversion transaction for tokens swap (single chain) according to the [API SPEC](https://github.com/WalletConnect/walletconnect-specs/pull/207/files#diff-f64bf5c2b17c9c6d32ae9d85d9a1ce5a4aa07cb680c6f4085de94d3c4915b9d9R366-R399).

* `1inch` provider implementation updates for the transaction builder,
* New endpoint handler implementation,
* Integration test was implemented for the new endpoint.

*This is an Alpha release implementation to support only single-chain swaps*

## How Has This Been Tested?

* [New integration test](https://github.com/WalletConnect/blockchain-api/pull/572/files#diff-d2a40c4ae4e4965cd2ea19db3965e98a04e70177954587a2a3edecfcf1263ce0R67-R96) for the endpoint.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/convert.test.ts
+++ b/integration/convert.test.ts
@@ -9,7 +9,8 @@ describe('Token conversion (single chain)', () => {
 
   const srcAsset = `${namespace}:${chainId}:0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee`;
   const destAsset = `${namespace}:${chainId}:0x111111111117dc0aa78b770fa6a738034120c302`;
-  const amount = 10000000000000000;
+  const userAddress = `${namespace}:${chainId}:0xf3ea39310011333095cfcccc7c4ad74034caba63`;
+  const amount = 100000;
 
   it('available tokens list', async () => {
     let resp: any = await httpClient.get(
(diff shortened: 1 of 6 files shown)


integration/convert.test.ts
```diff
@@ -9,7 +9,8 @@ describe('Token conversion (single chain)', () => {
 
   const srcAsset = `${namespace}:${chainId}:0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee`;
   const destAsset = `${namespace}:${chainId}:0x111111111117dc0aa78b770fa6a738034120c302`;
-  const amount = 10000000000000000;
+  const userAddress = `${namespace}:${chainId}:0xf3ea39310011333095cfcccc7c4ad74034caba63`;
+  const amount = 100000;
```
geekbrother, line 13: Lowering the amount to swap for tests.

src/providers/one_inch.rs
```diff
@@ -230,4 +253,77 @@ impl ConversionProvider for OneInchProvider {
...
+            return Err(RpcError::InvalidParameter(
+                "slippage parameter is necessary for this type of conversion".into(),
+            ));
+        }
+
+        let response = self.send_request(url, &self.http_client.clone()).await?;
+
+        if !response.status().is_success() {
+            error!(
+                "Error on building convert tx from 1inch provider. Status is not OK: {:?}",
```
nopestack, line 299:
```suggestion
                "Error on building convert tx from 1inch provider: {:?}",
```

nopestack: Consider that "Status is not OK" may be redundant if the log starts with "Error"

src/providers/one_inch.rs
```diff
@@ -230,4 +253,77 @@ impl ConversionProvider for OneInchProvider {
...
+    ) -> RpcResult<ConvertTransactionResponseBody> {
+        let (_, chain_id, src_address) = crypto::disassemble_caip10(&params.from)?;
+        let (_, dst_chain_id, dst_address) = crypto::disassemble_caip10(&params.to)?;
+        let (_, user_chain_id, user_address) = crypto::disassemble_caip10(&params.user_address)?;
+
+        // Check if from and to chain ids are different
+        // 1inch provider does not support cross-chain swaps
+        if (dst_chain_id != chain_id) || (user_chain_id != chain_id) {
+            return Err(RpcError::InvalidParameter(
+                "`from`, `to` and `userAddress` chain ids are different in a single chain swap"
```
nopestack, line 269: Consier:
```suggestion
                "`from`, `to` and `userAddress` chain IDs must have the same value"
```
Or 
```suggestion
                "Cross-chain swaps are not supported by provider"
```

nopestack (approved):
LGTM, just a few minor suggestions.

src/providers/one_inch.rs
geekbrother: Thanks, this is much better! Fixed to the first one.

src/providers/one_inch.rs
geekbrother: I'm actually filtering errors by this `Status is not OK` 😅 This is a pattern in all providers code. It makes sense to change this message, I should do this in a follow-up. Thanks!

PR #577, opened 2024-03-12
feat(providers): adding GetBlock provider

geekbrother:
# Description

This PR adds [GetBlock](https://getblock.io) to the RPC providers list. 

* GetBlock provider RPC proxy implementation,
  * Supported chains list: `eip155:1`, `eip155:56`, `eip155:137`, `eip155:324`, `eip155:17000`, `eip155:11155111`, `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp`.
* Integration tests implementation for the GetBlock provider and supported chains,
* Adding `RPC_PROXY_PROVIDER_GETBLOCK_ACCESS_TOKENS_JSON` environment variable and `getblock_api_tokens` terraform variable for passing API access tokens. Access tokens are passed by the environment variable in JSON format by using the key as a chain ID e.g. `{"eip155:1": "access_token"}`. This format was chosen because the provider is using unique API access tokens per each chain ID `https://go.getblock.io/ACCESS_TOKEN`.
* Grafana panels for the new provider were added.

## How Has This Been Tested?

* [New provider integration tests](https://github.com/WalletConnect/blockchain-api/pull/577/files#diff-e518eb42794c0f564302ee499cbbfb9c3bad12afeae78d5dbd104f8420229090).

## Merging policy 🚧

* [x] `getblock_access_tokens` TFC variable should be added with the API tokens.

## Due Diligence

* [ ] Breaking change


tests/functional/http/mod.rs
```diff
@@ -90,6 +91,37 @@ async fn check_if_rpc_is_responding_correctly_for_near_protocol(
     };
 }
 
+async fn check_if_rpc_is_responding_correctly_for_solana(
```
geekbrother, line 94: Extracting the logic to the function because we are now using a few providers for Solana.

src/providers/mod.rs
```diff
@@ -84,6 +86,7 @@ pub struct ProvidersConfig {
     pub coinbase_api_key: Option<String>,
     pub coinbase_app_id: Option<String>,
     pub one_inch_api_key: Option<String>,
+    pub getblock_access_tokens_json: Option<String>,
```
chris13524, line 89: IMO we don't need the `json` suffix here

```suggestion
    pub getblock_access_tokens: Option<String>,
```

terraform/ecs/cluster.tf
```diff
@@ -86,6 +86,8 @@ resource "aws_ecs_task_definition" "app_task" {
         { name = "RPC_PROXY_PROVIDER_COINBASE_API_KEY", value = var.coinbase_api_key },
         { name = "RPC_PROXY_PROVIDER_COINBASE_APP_ID", value = var.coinbase_app_id },
         { name = "RPC_PROXY_PROVIDER_ONE_INCH_API_KEY", value = var.one_inch_api_key },
+        { name = "RPC_PROXY_PROVIDER_GETBLOCK_ACCESS_TOKENS_JSON", value = var.getblock_api_tokens },
```
chris13524, line 89: Would be good to keep the names consistent

```suggestion
        { name = "RPC_PROXY_PROVIDER_GETBLOCK_ACCESS_TOKENS_JSON", value = var.getblock_access_tokens },
```

terraform/ecs/variables.tf
```diff
@@ -191,6 +191,12 @@ variable "one_inch_api_key" {
   sensitive   = true
 }
 
+variable "getblock_api_tokens" {
+  description = "API access tokens for GetBlock in JSON format"
```
chris13524, line 195:
```suggestion
  description = "Mapping of API access tokens for GetBlock in JSON format"
```

terraform/variables.tf
```diff
@@ -121,6 +121,12 @@ variable "one_inch_api_key" {
   sensitive   = true
 }
 
+variable "getblock_api_tokens" {
+  description = "API access tokens for GetBlock in JSON format"
```
chris13524, line 125:
```suggestion
  description = "Mapping of API access tokens for GetBlock in JSON format"
```

chris13524 (approved):
Some suggestions but looks great!

src/providers/mod.rs
geekbrother: Makes sense, Thanks. I added the description for this variable that it's a JSON.

src/providers/mod.rs
```diff
@@ -84,6 +86,8 @@ pub struct ProvidersConfig {
     pub coinbase_api_key: Option<String>,
     pub coinbase_app_id: Option<String>,
     pub one_inch_api_key: Option<String>,
+    /// GetBlock provider access tokens in JSON format
```
nopestack, line 89: is the value contained within the Option meant to be a JSON object or just a string of alphanumeric text?

nopestack (approved):

src/providers/mod.rs
geekbrother: We are passing the JSON string as an environment variable for config.

PR #579, opened 2024-03-13
fix: removing `unwrap` in updating weights

geekbrother:
# Description

This PR removed `unwrap()` in the weights update. This should properly handle the removal of the provider when the provider was removed, but there is still the data for it in Prometheus.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/weights.rs
+++ b/src/providers/weights.rs
@@ -12,7 +12,7 @@ use {
 #[derive(Debug, Copy, Clone)]
 pub struct Availability(u64, u64);
 
-pub type ParsedWeights = HashMap<String, (HashMap<ChainId, Availability>, Availability)>;
+pub type ParsedWeights = HashMap<ProviderKind, (HashMap<ChainId, Availability>, Availability)>;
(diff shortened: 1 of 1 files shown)


src/providers/weights.rs
```diff
@@ -116,6 +116,16 @@ fn calculate_chain_weight(
 #[tracing::instrument(skip_all)]
 pub fn update_values(weight_resolver: &WeightResolver, parsed_weights: ParsedWeights) {
     for (provider, (chain_availabilities, provider_availability)) in parsed_weights {
+        let provider_kind = match ProviderKind::from_str(&provider.clone()) {
```
chris13524, line 119: Wouldn't it be better to move this parse into `parsed_weights()` so that you move the error to be as early as possible?

geekbrother: Makes sense, thanks! Updated.

chris13524 (approved):

PR #583, opened 2024-03-14
fix: removing unwraps for proper handling and error context

geekbrother:
# Description

This PR adds minor changes for the proper `unwraps` handling where appropriate for adding additional error context and removing the possible app `panic`.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/project/storage/mod.rs
+++ b/src/project/storage/mod.rs
@@ -10,7 +10,7 @@ use {
         time::{Duration, Instant},
     },
     tap::TapFallible,
-    tracing::warn,
+    tracing::{error, warn},
 };
 
 mod config;
@@ -65,7 +65,7 @@ impl ProjectStorage {
         let serialized = match crate::storage::serialize(&data) {
             Ok(serialized) => serialized,
             Err(err) => {
-                warn!(?err, "failed to serialize cached project data");
+                error!(?err, "failed to serialize cached project data");
                 return;
             }
         };
--- a/src/project/storage/mod.rs
+++ b/src/project/storage/mod.rs
@@ -62,7 +62,13 @@ impl ProjectStorage {
     pub async fn set(&self, id: &str, data: &ProjectDataResult) {
         let cache_key = build_cache_key(id);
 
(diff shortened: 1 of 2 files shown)


src/project/storage/mod.rs
```diff
@@ -62,7 +62,13 @@ impl ProjectStorage {
     pub async fn set(&self, id: &str, data: &ProjectDataResult) {
         let cache_key = build_cache_key(id);
 
-        let serialized = crate::storage::serialize(&data).unwrap(); //?;
+        let serialized = match crate::storage::serialize(&data) {
+            Ok(serialized) => serialized,
+            Err(err) => {
+                warn!(?err, "failed to serialize cached project data");
```
chris13524, line 68: Being unable to serialize well-known data is very bad

```suggestion
                error!(?err, "failed to serialize cached project data");
```

chris13524 (approved):

src/providers/mod.rs
```diff
@@ -249,7 +249,9 @@ impl ProviderRepository {
         match WeightedIndex::new(weights) {
             Ok(dist) => {
                 let random = dist.sample(&mut OsRng);
-                let provider = keys.get(random).unwrap();
+                let provider = keys
+                    .get(random)
+                    .expect("Failed to get random provider: out of index");
```
nopestack, line 254: Is it ok to unwrap/expect here?

geekbrother: I think we are safe here because we are generating the random value from the known range where it shouldn't be out of range. If so, something very bad happened with our code, so panic should be expected.

nopestack: Can you add this explanation as a comment?

nopestack (approved):
lgtm but please address @chris13524 comments

src/providers/mod.rs
chris13524: I prefer putting explanation inside of the `expect()`

PR #589, opened 2024-03-21
fix(balance): using the correct implementation chain address instead of just first

geekbrother:
# Description

This PR fixes the bug where we are using the only first address token implementation instead of using the actual chain in the Zerion response.

## How Has This Been Tested?

* Manually tested by comparing the address with the actual token contract address.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -440,16 +440,20 @@ impl BalanceProvider for ZerionProvider {
                 symbol: f.attributes.fungible_info.symbol,
                 chain_id: crypto::ChainId::to_caip2(&f.relationships.chain.data.id),
                 address: {
-                    let formatted_address = f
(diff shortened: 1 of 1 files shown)


src/providers/zerion.rs
```diff
@@ -440,16 +440,20 @@ impl BalanceProvider for ZerionProvider {
...
                 address: {
-                    let formatted_address = f
+                    let chain_id_human = f.relationships.chain.data.id;
+                    let chain_address = f
                         .attributes
                         .fungible_info
                         .implementations
-                        .first()
+                        .iter()
+                        .find(|f| f.chain_id == chain_id_human)
```
geekbrother, line 449: Getting the address of the token for the implementation instead of just getting the first one.

nopestack (approved):

chris13524 (approved):

PR #591, opened 2024-03-22
fix: provider-returned non-200s should be 503s & fix autoscaling

chris13524:
# Description

If a provider returns a non-200 status code, we return this as-is instead of transforming into a 503 which we use to indicate a provider-side error. [Slack conversation](https://walletconnect.slack.com/archives/C03SCF66K2T/p1711134896606019?thread_ts=1711133669.730169&cid=C03SCF66K2T)

Also fixes that the default min/max autoscaling was set to 1 incorrectly. Also updated to match in Terraform.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -152,6 +152,7 @@ async fn handler_internal(
                 response.body()
             );
             state.metrics.add_failed_provider_call(provider.borrow());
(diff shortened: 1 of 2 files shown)


arein (approved):
LGTM

nopestack (approved):

geekbrother (approved):

PR #597, opened 2024-03-26
fix(zerion): handling properly `Polygon` native token address in balance

geekbrother:
# Description

According to our [API specification for balance](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-server-api#success-response-body-4) the token address should be null for the native tokens. Zerion endpoint returns `0x0000000000000000000000000000000000001010` for the Polygon native token that confusing our users. 
This PR adds the exception of handling the token address for the Polygon.

## How Has This Been Tested?

* Current integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -450,7 +450,15 @@ impl BalanceProvider for ZerionProvider {
                         .and_then(|f| f.address.clone());
                     let chain_id = crypto::ChainId::to_caip2(&chain_id_human);
                     if let Some(chain_address) = chain_address {
-                        chain_id.map(|chain_id| format!("{}:{}", &chain_id, chain_address))
+                        // For Polygon native token (Matic)
+                        // `0x0000000000000000000000000000000000001010`
+                        // address is returned, but address should be null
+                        // for native tokens
+                        if chain_address == "0x0000000000000000000000000000000000001010" {
+                            None
+                        } else {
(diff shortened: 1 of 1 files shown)


src/providers/zerion.rs
```diff
@@ -450,7 +450,15 @@ impl BalanceProvider for ZerionProvider {
                         .and_then(|f| f.address.clone());
                     let chain_id = crypto::ChainId::to_caip2(&chain_id_human);
                     if let Some(chain_address) = chain_address {
-                        chain_id.map(|chain_id| format!("{}:{}", &chain_id, chain_address))
+                        // For Polygon native token (Matic)
+                        // `0x0000000000000000000000000000000000001010`
+                        // address is returned, but address should be null
+                        // for native tokens
+                        if chain_address == "0x0000000000000000000000000000000000001010" {
```
nopestack, line 457: Can it be made a constant instead? And the comment could be an actual doc comment

nopestack (approved):
Lgtm though consider my comment

src/providers/zerion.rs
geekbrother: Thanks for a comment! Fixed in https://github.com/WalletConnect/blockchain-api/pull/597/commits/4915515c73540590b448fcb50da886f0f006c249

chris13524 (approved):

PR #600, opened 2024-03-29
feat: rate limiting

geekbrother:
# Description

This PR applies shared [rate limiting crate from utils-rs](https://github.com/WalletConnect/utils-rs/pull/14) to blockchain-api endpoints.
The combination of `IP` address and the endpoint is used as the identification key for the token bucket.

The following configuration environment variables were added:
* `RPC_PROXY_STORAGE_RATE_LIMITING_CACHE_REDIS_ADDR_READ` and `RPC_PROXY_STORAGE_RATE_LIMITING_CACHE_REDIS_ADDR_WRITE` are Redis endpoints for the L2 cache.
* `RPC_PROXY_RATE_LIMITING_MAX_TOKENS` token bucket size.
* `RPC_PROXY_RATE_LIMITING_REFILL_INTERVAL_SEC` token refill interval in seconds.
* `RPC_PROXY_RATE_LIMITING_REFILL_RATE` token refill rate per interval.

To enable rate-limiting all of the required variables must be set, otherwise, the rate-limiting will be disabled with the warning message at start.

Rate-limited counter metrics were added at e8a5cc486a69ae6ac11a70f25a0cd6c9ffc44ce9 to the Prometheus and Grafana panel to track how many rate-limited entries the L1 in-memory cache has so we can track how much users we are rate limiting at the moment and history.

Axum middleware was implemented and added to handlers in 3d0e989fe13090ab9d4864097ac62c72bdabd30e.

Default rate-limiting configuration in terraform variables in fd1e605cb18d079a9081d6a3f4284525e000a03f:
  * Maximum tokens in bucket: 100,
  * Refill rate: 2,
  * Refill interval: 1 second.

## How Has This Been Tested?

* New integration tests in 9cafe6c4df08bfd9d670e57fab714c27c089e1f2:
  * Flood more than max tokens and check is rate limited,
  * Flood less than max tokens to another endpoint and check is NOT rate-limited.

## Due Diligence

* [ ] Breaking change


src/lib.rs
```diff
@@ -79,6 +80,11 @@ const KEEPALIVE_IDLE_DURATION: Duration = Duration::from_secs(65);
 const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(65);
 const KEEPALIVE_RETRIES: u32 = 1;
 
+// Rate-limiting configuration
+const RATE_LIMIT_MAX_TOKENS: u32 = 300;
+const RATE_LIMIT_INTERVAL: Duration = Duration::from_secs(1);
+const RATE_LIMIT_REFILL_RATE: u32 = 100;
```
chris13524, line 86: Looks like this is 100 req/s per IP? Seems really high. What about 2 req/s with burst of 100?

Plus I think there should be a second rate limit for the project, e.g. 1k/s.

chris13524, line 86: Looks like this is 100 req/s per IP? Seems really high. What about 2 req/s with burst of 100?

Plus I think there should be a second rate limit for the project, e.g. 1k/s.

```suggestion
const RATE_LIMIT_MAX_TOKENS: u32 = 100;
const RATE_LIMIT_INTERVAL: Duration = Duration::from_secs(1);
const RATE_LIMIT_REFILL_RATE: u32 = 2;
```

geekbrother: Makes sense, probably it's too high.
I've moved the configuration to env variables in 09d67304f6a1544938fb4d1ad964d62a64e7b172 so we don't need to rebuild everything just to change it. 
Let's start with 100 burst and 2/sec and then I'll check the metric from e8a5cc486a69ae6ac11a70f25a0cd6c9ffc44ce9 how many are affected.

geekbrother: > Plus I think there should be a second rate limit for the project, e.g. 1k/s.

Speaking about project ID limiting I think this should be a follow-up PR. Let's ship this incrementally.

src/handlers/mod.rs
```diff
@@ -32,3 +36,25 @@ pub struct RpcQueryParams {
...
+    project_id: Option<&str>,
+) -> Result<(), RateLimitExceeded> {
+    state
+        .rate_limit
+        .as_ref()
+        .unwrap()
+        .is_rate_limited(
+            path.as_str(),
+            &network::get_forwarded_ip(headers.clone())
+                .unwrap_or_else(|| connect_info.0.ip())
```
chris13524, line 55: I think this defaulting to connecting IP is unsafe. If something breaks with the forwarding headers then we should _not_ be using the connecting IP as that's supposed to be the load balancer. We should at least be logging this as an error, not silently defaulting. And possibly we should not use a default at all and disable the rate limit or some other type of disable.

In Notify Server for the analytics if the forwarded header wasn't available I didn't lookup the connecting IP and recorded empty geo location values and emitted an `error!()`

src/handlers/history.rs
```diff
@@ -125,6 +125,14 @@ pub async fn handler(
     headers: HeaderMap,
     address: Path<String>,
 ) -> Result<Response, RpcError> {
+    handle_rate_limit(
+        state.clone(),
+        headers.clone(),
+        connect_info,
+        path.clone(),
```
chris13524, line 132: I think using the HTTP request path as the key is a bit naive and I would prefer an enum. For e.g. proxy there are a number of endpoints that resolve to the same handler. Not sure for history we might want 1 rate limit counter for the whole history API but not sure.

chris13524 (approved):

src/handlers/history.rs
geekbrother: Thanks for catching this!
The bucket key is a composite key for IP and matched path. The idea is that if there is a legitimate user he shouldn't exceed our max tokens anyway. So we can leave only IP-based rate-limiting btw. 
If there are malicious floods and we add an exact handler path instead of a matched path we can give an additional way to flood us. A matched path was added to the key composition for example for testing facilities, but not to give users making requests max tokens * on each path * path parameters.
So my point of view we should be ok with that and no legit users should be affected.

src/lib.rs
```diff
@@ -283,18 +320,28 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
...
+    };
+
     let app = app.with_state(state_arc.clone());
 
     info!("v{}", build_version);
-    info!("Running RPC Proxy on port {}", port);
+    info!("Running Blockchain-API server on port {}", port);
     let addr: SocketAddr = format!("{host}:{port}")
         .parse()
         .expect("Invalid socket address");
```
nopestack, line 347: can this be handled as an error instead of expecting?

src/utils/rate_limit.rs
```diff
@@ -0,0 +1,88 @@
...
+    interval: Duration,
+    refill_rate: u32,
+}
+
+impl RateLimit {
+    pub fn new(redis_addr: &str, max_tokens: u32, interval: Duration, refill_rate: u32) -> Self {
+        let redis_pool = Arc::new(
+            deadpool_redis::Config::from_url(redis_addr)
+                .create_pool(Some(deadpool_redis::Runtime::Tokio1))
+                .expect("Failed to create redis pool for rate limiting"),
```
nopestack, line 31: can this error be handled?

src/utils/rate_limit.rs
```diff
@@ -0,0 +1,88 @@
...
+        );
+        let mem_cache = Cache::builder()
+            .time_to_live(
+                interval
+                    .to_std()
+                    .expect("Failed to convert duration for rate limiting memory cache"),
```
nopestack, line 37: can this error be handled?

nopestack (approved):
Noticed some expects/unwrap at what appears to be boot time ops which is not egregious but I'd prefer them to be treated as errors if it makes sense. All ok apart from that

src/lib.rs
geekbrother: Thanks for the comment! This is an initial bootstrap and if the server can't bind to the listening address we must stop execution with the fatal error. So it's expected here.

src/utils/rate_limit.rs
geekbrother: Since it's a bootstrapping and shouldn't hurt it's better to skip the rate-limiting and start the server in case the Redis address is wrong in the configuration. Updated in 7ffe453593e647175ef73039fd32bffe319479bc thanks!

src/utils/rate_limit.rs
geekbrother: This is a conversion from std `Duration` and chrono's, because Moka uses a different `Duration`. If this happens something is really wrong with the app and probably we should expect here.

src/handlers/history.rs
chris13524: I think I assumed this was matched path (which is what it was). Still my suggestion was to not use the path as the rate limit and prefer something more explicit. E.g. there's both `/v1` and `/v1/` matched paths would be the proxy handler, and in the future we might want to group them. But no big deal right now, can always refactor later if needed.

But we should be aware that they can get 2x the rate limit for the proxy handler than we might expect them to be able to.

src/handlers/mod.rs
chris13524: What's the issue with local testing? Can't you set an `X-Forwarded` header?

PR #608, opened 2024-04-04
feat(providers): fungible price endpoint implementation

geekbrother:
# Description

This PR implements a fungible price endpoint according to the [SPEC proposal](https://walletconnect-specs-git-feat-tokenprices-walletconnect1.vercel.app/2.0/specs/servers/blockchain/blockchain-server-api#price) to get actual prices.
Since the API is compatible with different providers like Zerion and 1Inch, the Zerion provider adapter implementation was added.

## How Has This Been Tested?

* [New integration test](https://github.com/WalletConnect/blockchain-api/pull/608/files#diff-4493ff1a59d5bf63fc4e0b226e9908932f9c2ef71dafddf6ecf99f00c5739487).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/fungible_price.test.ts
+++ b/integration/fungible_price.test.ts
@@ -2,14 +2,21 @@ import { getTestSetup } from './init';
 
 describe('Fungible price', () => {
   const { baseUrl, projectId, httpClient } = getTestSetup();
+  const endpoint = `${baseUrl}/v1/fungible/price`;
 
   // RSR token address
   const implementation_address = 'eip155:1:0x320623b8e4ff03373931769a31fc52a4e78b5d70'
   const currency = 'usd'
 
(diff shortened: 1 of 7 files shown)


integration/fungible_price.test.ts
```diff
@@ -0,0 +1,25 @@
+import { getTestSetup } from './init';
+
+describe('Fungible price', () => {
```
nopestack, line 3: Wouldn't it make sense to also have tests for the error scenarios, e.g. a failure to parse a fungible token provider url?

geekbrother: Thanks for the comment. Agree! I think makes sense to add also check for malformed request arguments.

PR #610, opened 2024-04-06
fix: hardcode mainnet

chris13524:
# Description

Since the ethers library uses a hardcoded registry contract address, it doesn't support non-mainnet resolution anyway. So hardcoding mainnet. Additionally, changing the query params that this endpoint accepts to just be projectId since the other parameters were for the RPC endpoint.

[Slack conversation](https://walletconnect.slack.com/archives/C03RVH94K5K/p1712418300123819?thread_ts=1712373328.103609&cid=C03RVH94K5K)

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -144,7 +144,7 @@ async fn lookup_identity(
     address: H160,
     state: State<Arc<AppState>>,
     connect_info: ConnectInfo<SocketAddr>,
-    query: Query<RpcQueryParams>,
+    Query(query): Query<RpcQueryParams>,
     path: MatchedPath,
     headers: HeaderMap,
 ) -> Result<(IdentityLookupSource, IdentityResponse), RpcError> {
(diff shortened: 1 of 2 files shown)


src/handlers/identity.rs
```diff
@@ -188,14 +201,18 @@ async fn lookup_identity_rpc(
...
     path: MatchedPath,
     headers: HeaderMap,
 ) -> Result<IdentityResponse, RpcError> {
     let provider = Provider::new(SelfProvider {
         state: state.clone(),
         connect_info,
-        query,
+        query: Query(RpcQueryParams {
+            project_id,
+            chain_id: "1".to_string(),
```
nopestack, line 213: is it ok to have this hardcoded?

chris13524: Can you clarify your question? That's the whole point of this PR

geekbrother (approved):
LGTM!

src/handlers/identity.rs
nopestack: disregard the previous comment, I understood the context afterwards

PR #617, opened 2024-04-09
fix(server): enabling `tcp_nodelay` and `tcp_sleep_on_accept_errors`

geekbrother:
# Description

This PR making the following changes:
* Enables `tcp_nodelay` for a lower latency (this means data is sent immediately over the network without waiting for the buffer to fill or for an acknowledgment of previously sent data).
* Enables `tcp_sleep_on_accept_errors` to log errors if there are no unix sockets available and wait other than just silently shutdowns. So we can catch the exact error if this happens.
* Increasing retrying for keep-alive from 1 to 5.

## How Has This Been Tested?

* Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,7 +79,7 @@ use {
 const SERVICE_TASK_TIMEOUT: Duration = Duration::from_secs(15);
 const KEEPALIVE_IDLE_DURATION: Duration = Duration::from_secs(60);
 const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(10);
-const KEEPALIVE_RETRIES: u32 = 1;
+const KEEPALIVE_RETRIES: u32 = 5;
 
 mod analytics;
 pub mod database;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -79,7 +79,7 @@ use {
 const SERVICE_TASK_TIMEOUT: Duration = Duration::from_secs(15);
 const KEEPALIVE_IDLE_DURATION: Duration = Duration::from_secs(60);
 const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(10);
-const KEEPALIVE_RETRIES: u32 = 1;
+const KEEPALIVE_RETRIES: u32 = 5;
```
nopestack, line 82: What's the consequence of increasing these retries?

nopestack (approved):
Approving to unblock though please explain what's the expectation from merging this PR

geekbrother:
> Approving to unblock though please explain what's the expectation from merging this PR.

The part of our HTTP 502 errors related to the connection issues between App and LB. Enabling the `tcp_sleep_on_accept_errors` will log an error in case the issue is the leak of UNIX sockets/open files (now it just shuts down). `tcp_nodelay` is a common practice when using the service behind the LB and seems it's a common in case of using AWS ALB.

PR #620, opened 2024-04-10
fix(identity): handling properly errors in RPC call response

geekbrother:
# Description

This PR adds a proper handling of the following errors from the RPC call during the identity lookup:

* `Project Quota`: We have a lot of `HTTP 500` errors with the `NameLookup("RpcError: QuotaLimitReached")` message that should be returned as `HTTP 429`.
* `503 Service Unavailable`: When the RPC provider is unavailable.

All of the above errors are HTTP 500 now and should be distinguished.

## How Has This Been Tested?

* Current integration tests cover the good and bad names resolution. Quota limiting can't be tested as we have high quotas in prod.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -243,17 +243,27 @@ async fn lookup_name(
     address: Address,
 ) -> Result<Option<String>, RpcError> {
     provider.lookup_address(address).await.map_or_else(
-        |e| match e {
-            ProviderError::CustomError(e)
-                if &e == "SelfProviderError: RpcError: ProjectDataError(NotFound)" =>
-            {
-                Err(RpcError::ProjectDataError(ProjectDataError::NotFound))
+        |error| match error {
+            ProviderError::CustomError(e) if e.starts_with(SELF_PROVIDER_ERROR_PREFIX) => {
+                let error_detail = e.trim_start_matches(SELF_PROVIDER_ERROR_PREFIX);
+                // Exceptions for the detailed HTTP error return on RPC call
+                if error_detail.contains("ProjectDataError(NotFound)") {
+                    Err(RpcError::ProjectDataError(ProjectDataError::NotFound))
+                } else if error_detail.contains("QuotaLimitReached") {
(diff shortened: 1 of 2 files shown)


chris13524 (approved):
Good refactor!

geekbrother:
> Good refactor!

Thanks a lot! I've added other checks, so re-requesting a review.

src/handlers/identity.rs
```diff
@@ -243,17 +243,35 @@ async fn lookup_name(
...
+                if error_detail.contains("ProjectDataError(NotFound)") {
+                    Err(RpcError::ProjectDataError(ProjectDataError::NotFound))
+                } else if error_detail.contains("QuotaLimitReached") {
+                    Err(RpcError::QuotaLimitReached)
+                } else if error_detail.contains("503 Service Unavailable") {
+                    Err(RpcError::ProviderError)
+                } else if error_detail.contains("405 Method Not Allowed JSON-RPC result is 0x") {
+                    Err(RpcError::MethodNotAllowed(
+                        "JSON-RPC result is 0x".to_string(),
+                    ))
```
chris13524, line 259: I don't think it's appropriate to forward a 405 error. The client didn't make a mistake, this is an internal error. The upstream shouldn't even be returning a 405 to begin with.

src/handlers/identity.rs
```diff
@@ -243,17 +243,35 @@ async fn lookup_name(
...
+                } else if error_detail.contains("execution reverted") {
+                    Err(RpcError::ContractExecutionReverted)
```
chris13524, line 261: Not a client error

chris13524 (changes requested):

src/handlers/identity.rs
geekbrother: Thanks for the comment!
This and above 405 errors came from non-mainnet lookups which must be fixed by hardcoding the only mainnet lookup. Agree, that it doesn't make sense to handle them here anymore. Removed.

chris13524 (approved):

Issue #264, opened 2023-07-13
chore: refactor identity API to use internal RPC function


chris13524:
Right now we call the RPC handler and depend on the RPC handler for project ID authentication. But this is not a great design and also has the side-effect of providing unauthenticated access to cached lookup results.

Refactor this to call a shared internal function, and duplicate the authentication logic in both handlers.

PR #623, opened 2024-04-12
fix(identity): adding project ID validation first in the handler

geekbrother:
# Description

This PR adds project quota and validation as a first step to the identity handler. In this case, we are returning the quota limiting or wrong project ID early and we don't need to handle these errors from RPC later. Also, this will make availability metrics fair by excluding quota-limited requests from failed.

Resolves: #264 

## How Has This Been Tested?

* Current integration tests are passed.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -4,7 +4,6 @@ use {
         analytics::IdentityLookupInfo,
         error::RpcError,
         json_rpc::{JsonRpcError, JsonRpcResponse},
-        project::ProjectDataError,
         state::AppState,
         utils::{crypto, network},
     },
@@ -65,8 +64,11 @@ async fn handler_internal(
     headers: HeaderMap,
     Path(address): Path<String>,
 ) -> Result<Response, RpcError> {
-    let start = SystemTime::now();
+    state
+        .validate_project_access_and_quota(&query.project_id)
+        .await?;
 
+    let start = SystemTime::now();
     let address = address
         .parse::<Address>()
         .map_err(|_| RpcError::InvalidAddress)?;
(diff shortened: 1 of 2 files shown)


src/handlers/identity.rs
```diff
@@ -290,11 +292,7 @@ pub fn handle_rpc_error(error: ProviderError) -> Result<(), RpcError> {
         ProviderError::CustomError(e) if e.starts_with(SELF_PROVIDER_ERROR_PREFIX) => {
             let error_detail = e.trim_start_matches(SELF_PROVIDER_ERROR_PREFIX);
             // Exceptions for the detailed HTTP error return on RPC call
-            if error_detail.contains("ProjectDataError(NotFound)") {
```
geekbrother, line 293: Removing these two cases because now we are returning them early in the handler.

geekbrother, line 293: Removing these two cases because now we are returning them early in the handler.

chris13524: Isn't it possible for a race condition still and the quota goes to 0 after the new check above but before we call the proxy handler function?

chris13524: There is [an issue](https://github.com/WalletConnect/blockchain-api/issues/264) to refactor the proxy logic out of the proxy handler and into a shared function and then call that function here, rather than the HTTP handler. Maybe you've refactored closer to that now, not sure.

chris13524 (approved):

src/handlers/identity.rs
geekbrother: That's what I'm doing right now for the re-trying rpc call in case of the rate-limited provider 😅👍 Yes, makes sense to use it here instead of the handler. I'll update this PR on top of that with re-trying to omit the mentioned issue.

geekbrother: Updated with https://github.com/WalletConnect/blockchain-api/pull/623/commits to resolve #264

Issue #452, opened 2024-01-08
feat: fallback implementation for the providers


geekbrother:
Currently, we don't have a fallback in-case the provider fails or rate limits us. 
We either need:
* a fallback
* a [warning notice on our docs](https://walletconnect.slack.com/archives/C03US3FRU9L/p1701803594483259) that this provider may be unreliable

geekbrother:
Resolved by #624

PR #624, opened 2024-04-12
feat(rpc): adding rpc call retrying

### LINKED ISSUE #265: feat: retry failed RPC lookups 2 times
While we use weights to prioritize certain providers over others, we don't eliminate errors whenever we hit a low-weighted RPC provider.

We should retry the lookup to the weighting system 2 times to avoid getting errors as much as possible.

It should retry on a non-2xx status code. Possibly it could also detect that the body contains a JsonRpcResponse error or a non-JsonRpcResponse value.

We should also report metrics on this.

geekbrother:
# Description

This PR implements a retrying mechanism for RPC calls. This should be useful for the user experience when one of the providers returns a `rate-limited` error and we can retry up to max retries (currently is `3`) or fall with the appropriate error if all providers are rate-limited.

Metrics for attempts per chain ID are also added.

Resolve #265 and #452 

## How Has This Been Tested?

* Integration and provider tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -33,7 +33,7 @@ use {
         hash::Hash,
         sync::Arc,
     },
-    tracing::{error, info, log::warn},
+    tracing::{info, log::warn},
     wc::metrics::TaskMetrics,
 };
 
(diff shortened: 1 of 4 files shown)


chris13524:
Does this resolve https://github.com/WalletConnect/blockchain-api/issues/265?

src/providers/mod.rs
```diff
@@ -213,29 +214,43 @@ impl ProviderRepository {
...
         let weights: Vec<_> = providers.iter().map(|(_, weight)| weight.value()).collect();
         let keys = providers.keys().cloned().collect::<Vec<_>>();
+
+        let mut providers_result = vec![];
+
+        let providers_to_iterate = if max_providers > providers.len() {
+            providers.len()
+        } else {
+            max_providers
+        };
```
chris13524, line 224:
```suggestion
        let providers_to_iterate = std::cmp::min(max_providers, providers.len());
```

src/providers/mod.rs
```diff
@@ -213,29 +214,43 @@ impl ProviderRepository {
...
-
-                self.providers.get(provider).cloned().ok_or_else(|| {
-                    error!(
-                        "Provider not found during the weighted index check: {}",
-                        provider
-                    );
-                    RpcError::UnsupportedProvider(provider.to_string())
-                })
+                for _ in 0..providers_to_iterate {
+                    let provider = keys.get(dist.sample(&mut OsRng)).ok_or_else(|| {
```
chris13524, line 229: Noting that the same provider could be called multiple times. This risks a high priority provider going down and no alternative provider being attempted.

Here's a suggestion for an alternative that could be more robust:
1. Pick the first provider with our current algo (weighted random)
2. Excluding the first one, pick via weighted random. If only 1 provider available, pick the first one again
3. Pick the first provider again
4. Excluding the first and second one pick via weighted random. If only 2 are available, pick the second on again. If only 1 is available pick the first one again

src/providers/mod.rs
```diff
@@ -213,29 +214,43 @@ impl ProviderRepository {
...
-                })?;
-
-                self.providers.get(provider).cloned().ok_or_else(|| {
-                    error!(
-                        "Provider not found during the weighted index check: {}",
-                        provider
-                    );
-                    RpcError::UnsupportedProvider(provider.to_string())
-                })
+                for _ in 0..providers_to_iterate {
```
chris13524, line 228: Using `let providers_result = (0..providers_to_iterate).map(...)` here would be more rusty and save potential re-allocations of the Vec.

chris13524 (approved):
This should work. I left a suggestion on a more robust algorithm.

nopestack (approved):

src/providers/mod.rs
geekbrother: Totally! Thanks.

geekbrother:
> Does this resolve #265?

Looks like yes. Resolve #265 and #452

src/providers/mod.rs
geekbrother: Makes sense to use a map here. Thanks!

src/providers/mod.rs
geekbrother: Thanks a lot for the suggestion! 
I was also thinking about how to improve this case when the same provider was sampled twice. [I updated the PR](https://github.com/WalletConnect/blockchain-api/pull/624/commits/4b74afe5f7b23adf6bfc81cb50d8020303214338) with the removing from sampling provider that is already added to the list. So the fallback would be to a different available provider.
I'm thinking, is it makes sense to re-try the same provider (the case when there's only one provider for the chain). I think there should be some delay between re-trying the same provider.

src/providers/mod.rs
```diff
@@ -215,42 +215,48 @@ impl ProviderRepository {
...
+                        let provider = keys.get(dist_key).ok_or_else(|| {
+                            RpcError::WeightedProvidersIndex(format!(
+                                "Failed to get random provider for chain_id: {}",
+                                chain_id
+                            ))
+                        })?;
+
+                        // Update the weight of the provider to 0 to remove it from the next
+                        // sampling, as updating weights returns an error if
+                        // all weights are zero
```
chris13524, line 233: Is it possible all weights are already 0?

PR #630, opened 2024-04-15
feat: adding sdk type and version to CORS allowed headers

geekbrother:
# Description

This PR adds `x-sdk-type` and `x-sdk-version` headers to CORS allowed headers list.
This is useful for different behaviors based on the client SDK version. Currently, we need to make an exception for the certain SDK-version for the balances endpoint.

## How Has This Been Tested?

* Not tested.

## Due Diligence

* [ ] Breaking change


chris13524:
What kind of different behaviors are needed based on the SDK? Sounds to me like this would be better served with a "mode" flag

chris13524 (approved):

PR #631, opened 2024-04-15
fix(balance): respond with an empty balance for sdk version <= 4.1.8

geekbrother:
# Description

This PR adds skipping the balance Zerion call and responds with an empty array if the `x-sdk-version` was not passed in the headers. This should fix the issue of redundant calls in sdk versions <= 4.1.8 https://github.com/WalletConnect/web3modal/pull/2157. By checking the `x-sdk-version` presence we are distinguishing the SDK version requests <= version 4.1.8.

The context in the [Slack thread](https://walletconnect.slack.com/archives/C03SCF66K2T/p1713177390071089).

## How Has This Been Tested?

* Requesting the balance endpoint with the `x-sdk-version` header and without.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/balance.test.ts
+++ b/integration/balance.test.ts
@@ -6,10 +6,16 @@ describe('Account balance', () => {
   const fulfilled_address = '0xf3ea39310011333095CFCcCc7c4Ad74034CABA63'
   const empty_address = '0x5b6262592954B925B510651462b63ddEbcc22eaD'
   const currency = 'usd'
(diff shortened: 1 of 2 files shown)


chris13524 (commented):
Can we get a test case for this?

geekbrother:
> Can we get a test case for this?

Sure! Integration tests are updated.

PR #632, opened 2024-04-16
feat(analytics): adding analytics for balance lookups

geekbrother:
# Description

This PR adds analytics export for balance lookups with the following analytics item structure:

* timestamp,
* latency,
* symbol (token symbol),
* implementation chain ID
* quantity (token quantity according to decimals),
* value
* price
* currency
* address (lookup address)
* project_id,
* origin,
* region,
* country,
* continent.

Analytics export_prefix: `blockchain-api/balance-lookups`, export_name: `balance_lookups`.

## How Has This Been Tested?

* Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/analytics/balance_lookup_info.rs
+++ b/src/analytics/balance_lookup_info.rs
@@ -1,0 +1,51 @@
+use {
+    parquet_derive::ParquetRecordWriter,
+    serde::Serialize,
+    std::{sync::Arc, time::Duration},
+};
+
+#[derive(Debug, Clone, Serialize, ParquetRecordWriter)]
+#[serde(rename_all = "camelCase")]
+pub struct BalanceLookupInfo {
+    pub timestamp: chrono::NaiveDateTime,
+    pub latency_secs: f64,
+
+    pub symbol: String,
+    pub quantity: String,
+
+    pub address: String,
+    pub project_id: String,
+
+    pub origin: Option<String>,
+    pub region: Option<String>,
+    pub country: Option<Arc<str>>,
+    pub continent: Option<Arc<str>>,
+}
+
+impl BalanceLookupInfo {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        latency: Duration,
+        symbol: String,
+        quantity: String,
+        address: String,
+        project_id: String,
+        origin: Option<String>,
+        region: Option<Vec<String>>,
+        country: Option<Arc<str>>,
+        continent: Option<Arc<str>>,
+    ) -> Self {
+        Self {
+            timestamp: wc::analytics::time::now(),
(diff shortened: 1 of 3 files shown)


src/analytics/balance_lookup_info.rs
```diff
@@ -0,0 +1,51 @@
...
+            timestamp: wc::analytics::time::now(),
+            latency_secs: latency.as_secs_f64(),
+            symbol,
+            quantity,
+            address,
+            project_id,
+            origin,
+            region: region.map(|r| r.join(", ")),
+            country,
+            continent,
```
dnul, line 48: does chain apply to these lookups?  e.g. balance of usdc in polygon vs ethereum?

dnul (commented):
curious, do we get prices for this asset quotes? 

also see comments if chain applies in context

chris13524 (approved):

src/analytics/balance_lookup_info.rs
geekbrother: Yes, exactly! One token can be implemented in multiple chains like you mention about usdt or usdc. idk do we need to preserve the chain of implementation for the analytics, but I think it's worth having this.

geekbrother:
> curious, do we get prices for this asset quotes?

We have value and price for each balance. Do you want them to be exported as well @dnul?

src/analytics/balance_lookup_info.rs
dnul: can we add chain then? is it in context?

geekbrother: Sure. Added as `implementation_chain_id`.

geekbrother:
@dnul price and value depend on the currency of the request. 
I can add the currency to the analytic item as well or we can just add price and value only when the currency is `usd`.

dnul:
> @dnul price and value depend on the currency of the request. I can add the currency to the analytic item as well or we can just add price and value only when the currency is `usd`.

can I get an example? not sure I fully get this

            symbol,
            implementation_chain_id,
            quantity,
            value,
            price,

geekbrother:
> can I get an example?

@dnul Here is an example:
```
      "name": "Ethereum",
      "symbol": "ETH",
      "implementation_chain_id": "eip155:10", // Optimism
      "value": 0.4043452456561388, // USD value because the request is for USD
      "price": 3211.71, // Price of ETH
      "quantity": "0.000125897184258896" // quantity amount in token decimals.
```

dnul:
Thanks, much clearer now. let's add currency if possible

geekbrother:
> Thanks, much clearer now. let's add currency if possible

Done 👍

dnul (approved):
lgtm!

PR #636, opened 2024-04-17
feat(conversion): implementing gas price endpoint

geekbrother:
# Description

This PR implements a gas price endpoint according to [the SPEC](https://walletconnect-specs-git-feat-convertgasprice-walletconnect1.vercel.app/2.0/specs/servers/blockchain/blockchain-server-api#gas-price). The gas price required for the better UX for the conversion feature.

## How Has This Been Tested?

* A new integration test was added.

## Due Diligence

* [ ] Breaking change


integration/convert.test.ts
```diff
@@ -116,4 +116,25 @@ describe('Token conversion (single chain)', () => {
...
     expect(typeof resp.data.reasons[0].description).toBe('string')
   })
+
+  it('get gas price', async () => {
+    // Check for the Ethereum mainnet chain
+    let resp: any = await httpClient.get(
+      `${baseUrl}/v1/convert/gas-price?projectId=${projectId}&chainId=${caip2_chain_id}`
+    )
+    expect(resp.status).toBe(200)
+    expect(typeof resp.data.standard).toBe('string')
```
enesozturk, line 126: Shouldn't networks other than BSC return objects for instead of string? Guess they have the following structure:

```js
{
    "baseFee": "366686090537",
    "low": {
        "maxPriorityFeePerGas": "30000000000",
        "maxFeePerGas": "396686090537"
    },
    "medium": {
        "maxPriorityFeePerGas": "60000000000",
        "maxFeePerGas": "426686090537"
    },
    ..
}
```

geekbrother: Yes, it's covered by the [deserialization here](https://github.com/WalletConnect/blockchain-api/pull/636/files#diff-90f427e88833299b3715b6a62e733d00c4558c2c217036440db7f7bf721fd872R132). We have integration tests for Mainnet 1 and for BSC chain ID 56 to test two different cases. But our response is always the same as we agreed before.

enesozturk (approved):
Let's goo ⚡️ Left a comment for tests.

src/providers/one_inch.rs
```diff
@@ -373,4 +414,42 @@ impl ConversionProvider for OneInchProvider {
...
+        let base = format!(
+            "{}/gas-price/v1.5/{}",
+            &self.base_api_url,
+            evm_chain_id.clone()
+        );
+        let url = Url::parse(&base).map_err(|_| RpcError::ConversionParseURLError)?;
+
+        let response = self.send_request(url, &self.http_client.clone()).await?;
+
+        if !response.status().is_success() {
```
nopestack, line 433: why not use `response.status().is_client_error` here instead?

nopestack (approved):

src/providers/one_inch.rs
geekbrother: Thank you for the suggestion! In the case of using `is_client_error` we will catch only 4xx errors, but the provider can return 5xx which should be handled as well. What is missing here is the `bad request` handling. Added it. Thanks.

PR #637, opened 2024-04-18
feat(conversion): allowance endpoint

geekbrother:
# Description

This PR adds the swap allowance endpoint to check for the approved allowance before the swap according to the [API SPEC](https://walletconnect-specs-git-feat-allowanceconvert-walletconnect1.vercel.app/2.0/specs/servers/blockchain/blockchain-server-api#conversion-allowance).
The 1Inch provider is used as the source.

## How Has This Been Tested?

* New integration test.

## Due Diligence

* [ ] Breaking change


integration/convert.test.ts
```diff
@@ -137,4 +137,16 @@ describe('Token conversion (single chain)', () => {
     expect(typeof resp.data.instant).toBe('string')
 
   })
+
+  it('get allowance', async () => {
```
nopestack, line 141: consider porting to Rust instead

src/providers/one_inch.rs
```diff
@@ -452,4 +458,47 @@ impl ConversionProvider for OneInchProvider {
...
+        );
+        let mut url = Url::parse(&base).map_err(|_| RpcError::ConversionParseURLError)?;
+        url.query_pairs_mut()
+            .append_pair("tokenAddress", &token_address);
+        url.query_pairs_mut()
+            .append_pair("walletAddress", &wallet_address);
+
+        let response = self.send_request(url, &self.http_client.clone()).await?;
+
+        if !response.status().is_success() {
```
nopestack, line 482: is it not preferrable to use `response.status().is_client_error()`?

nopestack (approved):
few minor comments, all good though

src/providers/one_inch.rs
geekbrother: Thank you for the suggestion! It's the same as [here](https://github.com/WalletConnect/blockchain-api/pull/636#discussion_r1570421537):

> In the case of using is_client_error we will catch only 4xx errors, but the provider can return 5xx which should be handled as well...

integration/convert.test.ts
geekbrother: We have a [lot of tests](https://github.com/WalletConnect/blockchain-api/tree/master/integration) at the moment in Jest. 
This should probably take two days to move. I'll schedule this porting a bit later when at some free-time window, since it's a low prior now and last tech debt was spent for fixes and increasing service availability (that's a high prio). But yes, we should do this. 
The good thing in porting this to Rust is that we already have a schema to compare without rewriting this in JS.

PR #642, opened 2024-04-19
fix(Zerion): return only non-spam items

geekbrother:
# Description

This PR adds an additional filter to Zerion requests to filter out only non-spam items in transactions history and address balance since we don't need to return spam items. According to the Zerion API docs filter `filter[trash]` is applied.

## How Has This Been Tested?

* Current integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -223,6 +223,11 @@ pub struct ZerionMarketData {
     pub price: Option<f64>,
 }
 
+fn add_filter_non_trash_only(url: &mut Url) {
+    url.query_pairs_mut()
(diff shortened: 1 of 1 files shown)


nopestack (approved):

chris13524 (approved):
Maybe we could use a constant or a function or something to make a little more DRY? Lgtm tho. Thanks!!!

geekbrother:
> Maybe we could use a constant or a function or something to make a little more DRY? Lgtm tho. Thanks!!!

Makes sense, encapsulated in the function now. Thanks!!!

PR #643, opened 2024-04-22
fix(utils): properly handling of malformed CAIP-2 and 10

geekbrother:
# Description

This PR fixes a bug where malformed CAIP-2 and CAIP-10 addresses can cause panic in crypto utils, because of out-of-the-range vector access. Properly handling and additional errors were added.

## How Has This Been Tested?

* Updated unit tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -335,8 +335,8 @@ mod tests {
         assert_eq!(result.0, CaipNamespaces::Eip155);
         assert_eq!(result.1, "1".to_string());
 
-        let malformaed_caip2 = "eip1551";
-        let error_result = disassemble_caip2(malformaed_caip2);
+        let malformed_caip2 = "eip1551";
+        let error_result = disassemble_caip2(malformed_caip2);
         assert!(error_result.is_err());
     }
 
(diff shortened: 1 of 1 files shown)


src/utils/crypto.rs
```diff
@@ -166,12 +176,21 @@ pub fn disassemble_caip10(
...
+        Some(namespace) => match namespace.parse::<CaipNamespaces>() {
+            Ok(namespace) => namespace,
+            Err(_) => return Err(CryptoUitlsError::WrongNamespace(caip10.into())),
+        },
+        None => return Err(CryptoUitlsError::WrongNamespace(caip10.into())),
+    };
+    let chain_id = match parts.get(1) {
+        Some(chain_id) => chain_id.to_string(),
+        None => return Err(CryptoUitlsError::NoChainId(caip10.into())),
+    };
```
chris13524, line 189: `.unwrap_or()` would work well here

src/utils/crypto.rs
```diff
@@ -153,11 +157,17 @@ pub fn format_to_caip10(namespace: CaipNamespaces, chain_id: &str, address: &str
 /// Disassemble CAIP-2 to namespace and chainId
 pub fn disassemble_caip2(caip2: &str) -> Result<(CaipNamespaces, String), CryptoUitlsError> {
     let parts = caip2.split(':').collect::<Vec<&str>>();
```
chris13524, line 159: Perhaps a `parts.len()` check that it's always 2?

chris13524 (approved):

src/utils/crypto.rs
geekbrother: Makes sense. Looks like I overengineered here, it should be just a `len()` check 🤐. Thanks.

chris13524: No I mean you didn't necessarily over-engineer from my perspective. But I just wanted additional validation on the length that there wasn't additional `:` or something.

PR #644, opened 2024-04-23
feat(names): adding `eip1271` and `erc6492` support

geekbrother:
# Description

This PR adds EIP-1271 signatures support and ERC-6492 pre-deployed contracts support based on the shared `WalletConnectRust` implementation.
The local EOA verification was removed in favor of using unified ERC-6492 verification (making the `ecrecover` by calling the compiled contract bytecode). This should save additional RPC requests, assuming that most of the accounts will be smart accounts.
In the current flow we will support:
* ERC-6492 pre-deployed SA contracts signature check,
* EIP-1271 already deployed SA contracts signature check,
* EOA signature checks (no SA, no contracts, just an `ecrecover`).

## How Has This Been Tested?

* The shared [verify_eip6492 function has its own test](https://github.com/WalletConnect/WalletConnectRust/blob/2ce3a45f4d9ada7f05441a690a766ddf79f086f0/relay_rpc/src/auth/cacao/signature/eip1271/mod.rs#L109).
* The current EOA integration tests are passed.
* As a followup of this PR the integration test for SA should be added.

## Due Diligence

* [ ] Breaking change


src/handlers/profile/utils.rs
```diff
@@ -20,23 +18,6 @@ static DOMAIN_FORMAT_REGEX: Lazy<Regex> = Lazy::new(|| {
 const NAME_MIN_LENGTH: usize = 3;
 const NAME_MAX_LENGTH: usize = 64;
 
-#[tracing::instrument]
-pub fn verify_message_signature(
```
geekbrother, line 24: This is a duplicate of the test from utils/crypto. Removing it.

src/handlers/profile/utils.rs
```diff
@@ -111,43 +92,8 @@ mod tests {
     use {
         super::*,
         crate::handlers::profile::{ATTRIBUTES_VALUE_MAX_LENGTH, SUPPORTED_ATTRIBUTES},
-        ethers::types::H160,
-        std::str::FromStr,
     };
 
-    #[test]
-    fn test_verify_message_signature_valid() {
```
geekbrother, line 119: This is a duplicate of the test from utils/crypto. Removing it.

src/utils/crypto.rs
```diff
@@ -16,26 +22,122 @@ pub enum CryptoUitlsError {
...
 
+/// Veryfy message signature for eip1271 contract
+#[tracing::instrument]
+pub async fn is_address_has_code(
+    address: H160,
+    chain_id: &str,
+    rpc_project_id: &str,
+) -> Result<bool, CryptoUitlsError> {
+    let provider = Provider::<Http>::try_from(format!(
+        "https://rpc.walletconnect.com/v1?chainId={}&projectId={}",
```
geekbrother, line 94: This should be modified to use [rpc_call](https://github.com/WalletConnect/blockchain-api/blob/fec0ceed8d7924ea4993178adf223d1238fb6646/src/handlers/proxy.rs#L60) instead, but for alpha testing we can use the URL.

chris13524:
> To distinguish which verification flow (eip1271 or EOA) should be used the [is_address_has_code](https://github.com/WalletConnect/blockchain-api/pull/644/files#diff-43dc5270da6ae7bd3fbe8ee20c85643dcc3d3fb0feba36ad76570b327ef51667R88) utils function is used.

Do you want to just use `verify_6492()` instead and not do any of these checks?

geekbrother:
> > To distinguish which verification flow (eip1271 or EOA) should be used the [is_address_has_code](https://github.com/WalletConnect/blockchain-api/pull/644/files#diff-43dc5270da6ae7bd3fbe8ee20c85643dcc3d3fb0feba36ad76570b327ef51667R88) utils function is used.
> 
> Do you want to just use `verify_6492()` instead and not do any of these checks?

Does it support pure EOA (no SA deployed/pre-deployed)?

src/utils/crypto.rs
chris13524: I think you could implement a custom alloy transport for this. Also the WalletConnectRust API should be refactored to take an alloy `Provider` instead a `Url`. E.g. https://github.com/alloy-rs/alloy/blob/55a278c9db4e7e9b18b2bc2a290bd3de675d09f7/crates/transport-http/src/reqwest.rs

chris13524:
> I think we still need to check in case that's just an EOA address.

6492 verifies the EOA too, so you don't need to do any checks on your side to determine signature type. Just pass the signature and everything to `verify_6492()` and it supports all signature types.

src/utils/crypto.rs
```diff
@@ -211,45 +267,7 @@ pub fn constant_time_eq(a: impl AsRef<[u8]>, b: impl AsRef<[u8]>) -> bool {
...
 #[cfg(test)]
 mod tests {
-    use {
-        super::*,
-        ethers::types::H160,
-        std::{collections::HashMap, str::FromStr},
-    };
-
-    #[test]
-    fn test_verify_message_signature_valid() {
```
geekbrother, line 221: Removing these tests since we are not using the local `ecrecover` now and this is covered by [integration tests](https://github.com/WalletConnect/blockchain-api/blob/6ff30cb151180d3af3bed37e7e51bd6e08e89842/integration/names.test.ts#L186).

chris13524 (approved):

PR #648, opened 2024-05-06
feat(providers): using 1Inch instead of Zerion for fungible prices

geekbrother:
# Description

This PR changes the provider for fungibles prices from Zerion to 1Inch to support native token prices from 1Inch conversion parameters. 1Inch uses `0xeee` for the native tokens and Zerion uses just unique internal IDs. 
This change will fix errors while getting the conversion token prices. Before this change, we used the hardcoded list of the native tokens and Zerion ids.

## How Has This Been Tested?

Updated integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/one_inch.rs
+++ b/src/providers/one_inch.rs
@@ -1,25 +1,33 @@
 use {
     crate::{
         error::{RpcError, RpcResult},
-        handlers::convert::{
-            allowance::{AllowanceQueryParams, AllowanceResponseBody},
-            approve::{
-                ConvertApproveQueryParams,
-                ConvertApproveResponseBody,
-                ConvertApproveTx,
-                ConvertApproveTxEip155,
+        handlers::{
+            convert::{
+                allowance::{AllowanceQueryParams, AllowanceResponseBody},
+                approve::{
+                    ConvertApproveQueryParams,
+                    ConvertApproveResponseBody,
+                    ConvertApproveTx,
+                    ConvertApproveTxEip155,
+                },
+                gas_price::{GasPriceQueryParams, GasPriceQueryResponseBody},
+                quotes::{ConvertQuoteQueryParams, ConvertQuoteResponseBody, QuoteItem},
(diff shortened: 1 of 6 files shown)


src/providers/one_inch.rs
```diff
@@ -59,6 +67,79 @@ impl OneInchProvider {
...
     }
+
+    async fn get_token_price(
+        &self,
+        chain_id: &str,
+        address: &str,
+        currency: &PriceCurrencies,
+    ) -> Result<String, RpcError> {
+        let base = format!("{}/price/v1.1/{}/{}", &self.base_api_url, chain_id, address);
+        let mut url = Url::parse(&base).map_err(|_| RpcError::ConversionParseURLError)?;
```
chris13524, line 78: Using `join`:

```suggestion
        let mut url = self.base_api_url.join(format!("/price/v1.1/{}/{}", chain_id, address)).map_err(|_| RpcError::ConversionParseURLError)?;
```

Also are we properly validating/URL encoding `chain_id` and `address`? I'd default to URL encoding in-case something upstream forgets to validate.

chris13524 (approved):

nopestack (approved):

src/providers/one_inch.rs
geekbrother: Thanks for the good comments! Join is much nicer! 
In case of bad encoding (if someone will use other than utf8 in 2024 😅) there will be an error that the chain_id/address is in the wrong format with the bad characters in the error, so the error should be obvious if it happens.

chris13524: I meant if their chain ID or address has a slash in it and creates an unexpected URL

geekbrother: Yes, this makes sense. We need to check this earlier. Thanks for catching this.

geekbrother: I've added the proper format validation for the chain ID and address during the CAIP-2/10 disassembling in utils. This will cut such type of URL injection attack where CAIP2/10 disassembling is used.

PR #653, opened 2024-05-10
feat(convert): adding `referrer` to 1Inch provider

geekbrother:
# Description

This PR adds `referrer` to the 1Inch `quote` and `swap` calls to pass fees depositor address. This will add fees redirection. For security reason, the referrer is stored as a TFC variable.

## How Has This Been Tested?

Configuration variables test were updated.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -167,6 +167,9 @@ impl ProviderRepository {
             .clone()
             .unwrap_or("ONE_INCH_API_KEY".into());
         let one_inch_referrer = config.one_inch_referrer.clone();
+        if one_inch_referrer.is_none() {
+            warn!("ONE_INCH_REFERRER is not set");
+        }
 
         let zerion_provider = Arc::new(ZerionProvider::new(zerion_api_key));
         let one_inch_provider = Arc::new(OneInchProvider::new(one_inch_api_key, one_inch_referrer));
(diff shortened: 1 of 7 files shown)


src/providers/mod.rs
```diff
@@ -89,6 +89,7 @@ pub struct ProvidersConfig {
     pub coinbase_api_key: Option<String>,
     pub coinbase_app_id: Option<String>,
     pub one_inch_api_key: Option<String>,
+    pub one_inch_referrer: Option<String>,
```
chris13524, line 92: thought: Making this optional, while it enables easier local development, also creates a potential bug if TF Cloud is mis-configured.

Would be great to refactor in the future to support two different config entrypoints: local and deployed like [Notify Server has](https://github.com/WalletConnect/notify-server/tree/main/src/config) which would allow this optional in local deployment, but required in prod deployment.

chris13524 (approved):

src/providers/mod.rs
geekbrother: Makes sense. I will add a warning if it's not set for now for easy catching this in logs. At the moment if the TFC variable is not set, there should be a TF validation error as it's required by the TF configuration and expected to be set in TFC.

PR #655, opened 2024-05-14
feat(balance): adding forced balance update by the contract address

geekbrother:
# Description

This PR adds a forced balance update for a certain token by passing the ERC-20 token contract address. According to the SPEC the optional `forceUpdate` argument can be passed to the `/balance` endpoint with the list of token contract addresses to force the balance update by invoking the RPC call to the contract and overriding the balance response from Zerion for certain tokens contract addresses.
This improvement is useful to live update the balance at the UI when making token swaps or on-ramp payments by bypassing Zerion's cached balances.

## How Has This Been Tested?

* Manually tested by changing the wallet address balance and triggering the forced update.
* The integration test for the balance was updated.

## Due Diligence

* [ ] Breaking change


src/utils/crypto.rs
```diff
@@ -104,6 +108,39 @@ pub async fn verify_eip6492_message_signature(
...
+    rpc_project_id: &str,
+) -> Result<U256, CryptoUitlsError> {
+    abigen!(
+        ERC20Contract,
+        r#"[
+            function balanceOf(address account) external view returns (uint256)
+        ]"#,
+    );
+
+    let provider = Provider::<Http>::try_from(format!(
```
geekbrother, line 126: This will be refactored in a followup to use the SelfProvider and direct rpc calls, like in https://github.com/WalletConnect/blockchain-api/pull/649

nopestack (approved):

src/utils/crypto.rs
```diff
@@ -289,6 +326,23 @@ pub fn constant_time_eq(a: impl AsRef<[u8]>, b: impl AsRef<[u8]>) -> bool {
     result == 0
 }
 
+/// Format token amount to human readable format according to the token decimals
+pub fn format_token_amount(amount: U256, decimals: u32) -> String {
```
chris13524, line 330: Unit tests plz

src/utils/crypto.rs
```diff
@@ -289,6 +326,23 @@ pub fn constant_time_eq(a: impl AsRef<[u8]>, b: impl AsRef<[u8]>) -> bool {
...
+    // Handle cases where the total digits are less than or equal to the decimals
+    if amount_str.len() <= decimals_usize {
+        let required_zeros = decimals_usize - amount_str.len() + 1;
+        let zeros = "0".repeat(required_zeros);
+        return format!("0.{}{}", zeros, amount_str);
+    }
+
+    // Insert the decimal point at the correct position
+    let (integer_part, decimal_part) = amount_str.split_at(amount_str.len() - decimals_usize);
+    format!("{}.{}", integer_part, decimal_part)
```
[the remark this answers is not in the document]
chris13524: Did some research, not immediately obvious how to simplify this. Even in alloy not a clear solution. Probably fine

src/utils/crypto.rs
geekbrother: Updated [with tests](https://github.com/WalletConnect/blockchain-api/pull/655/files#diff-43dc5270da6ae7bd3fbe8ee20c85643dcc3d3fb0feba36ad76570b327ef51667R501-R526).

chris13524 (approved):

PR #661, opened 2024-05-24
fix(balance): injecting forced balance update to the zero balance token response

geekbrother:
# Description

This PR fixes the bug where a forced balance update is invoked, but no updated balance in response due to the absence of the token in the balance object due to zero balance. 
As a fix, we are injecting the token balance item if there are no token item found in the current Zerion response that overrides the recent balance update from JSON-RPC or contract response.

[Slack conversation.](https://walletconnect.slack.com/archives/C03RVH94K5K/p1716557794631039?thread_ts=1713176741.098649&cid=C03RVH94K5K)

## How Has This Been Tested?

* [A new integration test was implemented](https://github.com/WalletConnect/blockchain-api/pull/661/files#diff-13d7369ef8e242470eace4b6a7039baf1685c66f865763226588f2c6bc652dccR93).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/balance.rs
+++ b/src/handlers/balance.rs
@@ -1,5 +1,5 @@
 use {
-    super::HANDLER_TASK_METRICS,
+    super::{SupportedCurrencies, HANDLER_TASK_METRICS},
     crate::{
         analytics::BalanceLookupInfo,
         error::RpcError,
@@ -15,7 +15,6 @@ use {
     hyper::HeaderMap,
     serde::{Deserialize, Serialize},
     std::{
-        fmt::Display,
         net::SocketAddr,
         sync::Arc,
         time::{Duration, SystemTime},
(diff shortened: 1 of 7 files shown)


src/handlers/balance.rs
```diff
@@ -25,45 +24,11 @@ use {
     wc::future::FutureExt,
 };
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
```
geekbrother, line 28: Unifying to the `SupportedCurrencies`

src/handlers/fungible_price.rs
```diff
@@ -1,57 +1,23 @@
...
     },
     serde::{Deserialize, Serialize},
-    std::{fmt::Display, sync::Arc},
+    std::sync::Arc,
     tap::TapFallible,
     tracing::log::error,
     wc::future::FutureExt,
 };
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
```
geekbrother, line 16: Unifying to the `SupportedCurrencies`

src/handlers/balance.rs
```diff
@@ -240,8 +206,52 @@ async fn handler_internal(
...
+                    contract_address
+                );
+                RpcError::BalanceProviderError
+            })?;
+
+            response.balances.push(BalanceItem {
+                name: token_info.name.clone(),
+                symbol: token_info.symbol.clone(),
+                chain_id: Some(caip2_chain_id.clone()),
+                address: if contract_address == H160::repeat_byte(0xee) {
```
chris13524, line 238: `H160::repeat_byte(0xee)` should be a constant

src/handlers/balance.rs
```diff
@@ -223,6 +188,7 @@ async fn handler_internal(
                     balance.price,
                     balance.quantity.decimals.parse::<u32>().unwrap_or(0),
                 ));
+                continue;
```
chris13524, line 191: I would use `if/else if/else` instead of `continue`

chris13524 (approved):

src/handlers/balance.rs
geekbrother: Makes sense. Thanks!

src/handlers/balance.rs
```diff
@@ -240,8 +206,52 @@ async fn handler_internal(
...
                 }
             }
+            // Appending the token item to the response if it's not in
+            // the balance response due to the zero balance
+            let get_price_info = state
+                .providers
+                .fungible_price_provider
+                .get_price(
+                    &chain_id.clone(),
+                    format!("{:#x}", contract_address).as_str(),
```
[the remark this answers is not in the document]
geekbrother: > accept an Address rather than a string

Thanks! This is a good catch but the idea of using a string here is to support non-EVM addresses in the future since the `Address` is a H160.

PR #666, opened 2024-05-30
feat: rpc source analytics

chris13524:
# Description

Adds the source of the RPC call to the analytics export. [Slack conversation](https://walletconnect.slack.com/archives/C04HK47TNBB/p1717082696327969?thread_ts=1717064623.738729&cid=C04HK47TNBB)

Also fixes `devloop` so that it works w/o tflint and tfsec. Also runs checks under stable Rust as this is what CI uses.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/justfile
+++ b/justfile
@@ -159,7 +159,7 @@ cargo-test-default: _check-cmd-cargo
 # Run project tests with all features activated
 cargo-test-all: _check-cmd-cargo
   @printf '==> Testing project ({{ light-green }}all features{{ nocolor }})\n'
-  cargo +nightly test --all-features
+  cargo test --all-features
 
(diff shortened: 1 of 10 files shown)


dnul (approved):
lgtm, 

what would be the different `source`  values?

chris13524:
> what would be the different source values?

@dnul you can see them used in this PR. They are defined [here](https://github.com/WalletConnect/blockchain-api/pull/666/files#diff-3f3e894cb2c6e39811b88b04486c3ba0e7e468ec0d28a41450c93754d1ab89c7R62) but are in snake_cake

PR #672, opened 2024-06-14
feat(metrics): implementing account names count metrics watcher

geekbrother:
# Description

This PR implements an account names counter metrics watcher that periodically checks the names rows count in the `names` database table and stores the value to the `account_names_count` metrics gauge.

The new Grafana dashboard PR #674 

## How Has This Been Tested?

* New unit test for database helper function,

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/database/helpers.rs
+++ b/src/database/helpers.rs
@@ -284,7 +284,7 @@ pub async fn insert_or_update_address<'e>(
     Ok(result_map)
 }
 
-#[instrument(skip(postgres))]
+#[instrument(skip(postgres), level = "debug")]
 pub async fn get_account_names_stats(
     postgres: &PgPool,
 ) -> std::result::Result<AccountNamesStats, sqlx::error::Error> {
--- a/src/database/helpers.rs
+++ b/src/database/helpers.rs
@@ -288,7 +288,7 @@ pub async fn insert_or_update_address<'e>(
 pub async fn get_account_names_stats(
     postgres: &PgPool,
 ) -> std::result::Result<AccountNamesStats, sqlx::error::Error> {
-    let query = "SELECT COUNT(name) FROM names AS count";
+    let query = "SELECT COUNT(*) FROM names AS count";
     let stats = sqlx::query_as::<Postgres, AccountNamesStats>(query)
         .fetch_one(postgres)
         .await?;
--- a/src/database/helpers.rs
+++ b/src/database/helpers.rs
@@ -1,7 +1,7 @@
 use {
     crate::database::{error::DatabaseError, types, utils},
     chrono::{DateTime, Utc},
-    sqlx::{PgPool, Postgres, Row},
+    sqlx::{FromRow, PgPool, Postgres, Row},
     std::collections::HashMap,
     tracing::{error, instrument},
 };
@@ -12,6 +12,11 @@ struct RowAddress {
     chain_id: String,
     address: String,
     created_at: DateTime<Utc>,
+}
+
+#[derive(Debug, FromRow)]
+pub struct AccountNamesStats {
+    pub count: i64,
 }
 
 /// Initial name registration insert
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -395,6 +396,17 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
         tokio::spawn(weights_updater),
         tokio::spawn(system_metrics_updater),
         tokio::spawn(profiler),
+        // Spawning a new task to observe metrics from the database by interval polling
+        tokio::spawn({
+            let postgres = state_arc.postgres.clone();
+            async move {
+                let mut interval = tokio::time::interval(DB_STATS_POLLING_INTERVAL);
```
nopestack, line 403: Why poll every 60 seconds?

geekbrother: It looks like a reasonable interval for the Grafana chart since I don't see any reason to poll it more often. 
We can even increase it but the cost of running the SQL query like `COUNT(x)` on the indexed fields is very low since it's just an index size and no any table seeks is needed.

nopestack: Depending on how big the table is, the performance cost may not be as low as you think. Consider that the db needs to scan all rows irrespective of the index to count because the query has no where clause to reduce the scope.

geekbrother: Good catch! You are right. Interesting... Looks like since Postgre12 the database will try to utilize indexes if using `COUNT(x)` and not `COUNT(*)` and x is in the index, but that's not guaranteed. 
We can possibly use `reltuples` from `pg_class`, but in this case, we need to wait for autovacuum to get the exact row count.

What I'm thinking about:
Even if we have 1M name records the table size will be relatively small, even if the db decided to scan it. But in that case the polling period should be increased to a safer ~5 mins interval.

chris13524: Once an hour would be good I would think

src/database/helpers.rs
```diff
@@ -278,3 +283,14 @@ pub async fn insert_or_update_address<'e>(
 
     Ok(result_map)
 }
+
+#[instrument(skip(postgres))]
+pub async fn get_account_names_stats(
+    postgres: &PgPool,
+) -> std::result::Result<AccountNamesStats, sqlx::error::Error> {
+    let query = "SELECT COUNT(name) FROM names AS count";
```
chris13524, line 291: You don't need so specify the name?

```suggestion
    let query = "SELECT COUNT(*) FROM names AS count";
```

src/metrics.rs
```diff
@@ -537,4 +549,25 @@ impl Metrics {
         self.add_memory_total(system.total_memory() as f64);
         self.add_memory_used(system.used_memory() as f64);
     }
+
+    /// Update the account names count from database
+    #[instrument(skip_all)]
```
chris13524, line 554:
```suggestion
    #[instrument(skip_all, level = "debug")]
```

src/database/helpers.rs
```diff
@@ -278,3 +283,14 @@ pub async fn insert_or_update_address<'e>(
 
     Ok(result_map)
 }
+
+#[instrument(skip(postgres))]
```
chris13524, line 287:
```suggestion
#[instrument(skip(postgres), level = "debug")]
```

src/lib.rs
geekbrother: Personally, I would poll often than 1 hr, maybe 15 mins, but ok to go with an hour if you guys think it's ok.
The polling interval is 1 hour now.

src/database/helpers.rs
geekbrother: I believe by specifying (limiting) the field Postgres will try to use the index, since this column is in the index, while `*` will force to seek.

nopestack: @geekbrother you can verify if the query indeed benefits from the index by running the query planner like so:

```sql
EXPLAIN ANALYZE SELECT COUNT(name) FROM names;
```
and see if it gives you anything different from COUNT(*)

geekbrother: Interesting that despite of using filtering in `COUNT(x)` the `COUNT(*)` is even faster during tests.

> Leaving the choice up to the optimizer by using COUNT(*) lets it optimize based on conditions automatically.

Changing to `COUNT(*)`.

geekbrother: [PostgreSQL doc reference](https://wiki.postgresql.org/wiki/Slow_Counting) regarding the `COUNT` and this topic.

PR #676, opened 2024-06-17
feat(analytics): adding names registrations analytics

geekbrother:
# Description

This PR adds name registration analytics with the following data:
* Name,
* Owner's address,
* Registration chain ID (in case the user is not at the mainnet),
* Geo data: origin, region, country, continent.

The analytics data is stored in the `blockchain-api/name-requests/name_registrations` S3 bucket.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/analytics/mod.rs
+++ b/src/analytics/mod.rs
@@ -1,3 +1,9 @@
+pub use {
+    account_names_info::AccountNameRegistration, balance_lookup_info::BalanceLookupInfo,
+    config::Config, history_lookup_info::HistoryLookupInfo,
+    identity_lookup_info::IdentityLookupInfo, message_info::*,
+    onramp_history_lookup_info::OnrampHistoryLookupInfo,
+};
 use {
     aws_sdk_s3::Client as S3Client,
     std::{net::IpAddr, sync::Arc, time::Duration},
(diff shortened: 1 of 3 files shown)


src/analytics/mod.rs
```diff
@@ -287,6 +293,27 @@ impl RPCAnalytics {
...
+
+        let observer = Observer(DataKind::NameRegistrations);
+        let name_registrations = BatchCollector::new(
+            CollectorConfig {
+                data_queue_capacity: DATA_QUEUE_CAPACITY,
+                ..Default::default()
+            },
+            ParquetBatchFactory::new(Default::default()).with_observer(observer),
+            AwsExporter::new(AwsConfig {
+                export_prefix: "blockchain-api/name-requests".to_owned(),
```
chris13524, line 312:
```suggestion
                export_prefix: "blockchain-api/name-registrations".to_owned(),
```

chris13524 (approved):
Is registration all we want to track? What about name updates? Should this be modeled in a different way to support?

nopestack (approved):

geekbrother:
> Is registration all we want to track? What about name updates?

The registration is the only we need at the moment since we do not have UI for the name updates yet (( Just the registration and resolution.

>  Should this be modeled in a different way to support?

It should be a different struct for updates.

dnul (approved):
LGTM! sorry for the delay!

PR #678, opened 2024-06-18
fix(network): using the last forwarded IP from the list

geekbrother:
# Description

This PR changes the user IP to get the last IP from the list in `X-Forwarded-For` instead of the first since the AWS Load Balancer appends the IP to the end.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/network.rs
+++ b/src/utils/network.rs
@@ -73,3 +73,30 @@ pub fn get_forwarded_ip(headers: HeaderMap) -> Option<IpAddr> {
         .and_then(|header| header.split(',').last())
         .and_then(|client_ip| client_ip.trim().parse::<IpAddr>().ok())
 }
+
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
I think this works, but can we switch to this extension? https://github.com/WalletConnect/push-server/blob/8b7d285271d011061b6aa282eacecc8587b2529b/src/lib.rs#L202

geekbrother:
> I think this works, but can we switch to this library? https://github.com/WalletConnect/push-server/blob/8b7d285271d011061b6aa282eacecc8587b2529b/src/lib.rs#L202

Added the unit test to make sure it works as expected. Yes, this should be a follow-up after this quick fix.

PR #679, opened 2024-06-19
fix(tests): tuning rate-limiting integration test

geekbrother:
# Description

This PR tuning max token bucket size variables to `30` for the integration tests as a follow-up from #675. 
This should fix the rate-limiting [integration test error](https://github.com/WalletConnect/blockchain-api/actions/runs/9580015523/job/26420144164#step:5:49).

## How Has This Been Tested?

* Tested by running tests locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/ratelimiting.test.ts
+++ b/integration/ratelimiting.test.ts
@@ -4,8 +4,8 @@ describe('Rate limiting', () => {
   const { baseUrl, projectId, httpClient } = getTestSetup();
 
   it('Simulate flood and check is rate limited', async () => {
-    // Using default max tokens of 100
-    const max_tokens = 100;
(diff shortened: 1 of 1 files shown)


nopestack:
Why was the test broken exactly? saw it failing because it received 30 requests? did I read that correctly?

geekbrother:
> Why was the test broken exactly? 

We changed the token bucket size from 100 to 30 in #675, but didn't change the size for the integration test (it can't be run locally without modification because of using only X-forwarded IP).

> saw it failing because it received 30 requests? did I read that correctly?

Yes, that's correct!

chris13524 (approved):

PR #681, opened 2024-06-21
feat(sessions): irn client scaffolding

geekbrother:
# Description

This PR implements the IRN client abstraction into the blockchain-api. The following changes are made:

**CI/CD and build**
* `WalletConnect/irn-node` repository was added as a git submodule.
* `irn_api` and `network` packages were added as dependencies from the `irn-node`.
* `BSL-1.0` license was added to the allowed list.

**The following IRN client configuration parameters were added:**
 * `RPC_PROXY_IRN_NODE` environment variable and `irn_node` terraform variable for the IRN node address to connect to in IP:Socket format.
 * `RPC_PROXY_IRN_KEY` environment variable and `irn_key` terraform variable for the IRN client private key.
 * `RPC_PROXY_IRN_NAMESPACE` environment variable and `irn_namespace` terraform variable for the IRN storage namespace.
 * `RPC_PROXY_IRN_NAMESPACE_SECRET` environment variable and `irn_namespace_secret` terraform variable for the IRN storage namespace secret.

**The following IRN client storage commands were added:**
* Key-value storage `set`,`get`, `del`.
* Hashmap storage `hset`, `hget`, `hdel`, `hfields`, `hvalues`.

IRN client abstraction is bound to the `irn` global state to be accessible from handlers.

The record TTL is 30 days. (Which should be ok for sessions).

## How Has This Been Tested?

* [Unit test](https://github.com/WalletConnect/blockchain-api/pull/681/files#diff-90dd0ed862129223a6bf84ebb30db6483bbbde5628a5dda8139c636fca912d9bR194) for the helper function was implemented and run in CI.
* [Test to run `get`, `set`, `del`](https://github.com/WalletConnect/blockchain-api/pull/681/files#diff-90dd0ed862129223a6bf84ebb30db6483bbbde5628a5dda8139c636fca912d9bR216) were implemented. The test is ignored by default, because we don't want to spin up a cluster for each CI (and will be covered in the handler integration test). 
  * It can be run locally by starting the local cluster and run `cargo test -- storage::irn::tests::test_irn_client_set_get_del --exact --show-output --ignored`. The test were successfully passed in the local environment.
* [Test to run hashmap operations](https://github.com/WalletConnect/blockchain-api/pull/681/files#diff-90dd0ed862129223a6bf84ebb30db6483bbbde5628a5dda8139c636fca912d9bR248) were implemented. The test is ignored by default, because we don't want to spin up a cluster for each CI (and will be covered in the handler integration test). 
  * It can be run locally by starting the local cluster and run `cargo test -- storage::irn::tests::test_irn_client_hashmap --exact --show-output --ignored`. The test were successfully passed in the local environment.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,3 +31,9 @@ export RPC_PROXY_POSTGRES_URI="postgres://postgres@localhost/postgres"
 # export RPC_PROXY_RATE_LIMITING_MAX_TOKENS=100
 # export RPC_PROXY_RATE_LIMITING_REFILL_INTERVAL_SEC=1
 # export RPC_PROXY_RATE_LIMITING_REFILL_RATE=2
+
+# Uncomment for using the IRN client
+# export RPC_PROXY_IRN_NODE=127.0.0.1:3011
+# export RPC_PROXY_IRN_KEY=base64_key
+# export RPC_PROXY_IRN_NAMESPACE=namespace
+# export RPC_PROXY_IRN_NAMESPACE_SECRET=namespace_secret
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -6,6 +6,7 @@ use {
         profiler::ProfilerConfig,
         project::{storage::Config as StorageConfig, Config as RegistryConfig},
         providers::{ProviderKind, ProvidersConfig, Weight},
+        storage::irn::Config as IRNConfig,
         utils::rate_limit::RateLimitingConfig,
     },
     serde::de::DeserializeOwned,
@@ -48,6 +49,7 @@ pub struct Config {
     pub profiler: ProfilerConfig,
     pub providers: ProvidersConfig,
     pub rate_limiting: RateLimitingConfig,
+    pub irn: IRNConfig,
 }
 
 impl Config {
@@ -61,6 +63,7 @@ impl Config {
             profiler: from_env("RPC_PROXY_PROFILER_")?,
             providers: from_env("RPC_PROXY_PROVIDER_")?,
             rate_limiting: from_env("RPC_PROXY_RATE_LIMITING_")?,
+            irn: from_env("RPC_PROXY_IRN_")?,
         })
     }
 }
@@ -85,6 +88,7 @@ mod test {
             profiler::ProfilerConfig,
             project,
             providers::ProvidersConfig,
+            storage::irn::Config as IRNConfig,
             utils::rate_limit::RateLimitingConfig,
         },
         std::net::Ipv4Addr,
@@ -168,6 +172,11 @@ mod test {
             ("RPC_PROXY_RATE_LIMITING_MAX_TOKENS", "100"),
             ("RPC_PROXY_RATE_LIMITING_REFILL_INTERVAL_SEC", "1"),
             ("RPC_PROXY_RATE_LIMITING_REFILL_RATE", "10"),
+            // IRN config.
+            ("RPC_PROXY_IRN_NODE", "node"),
+            ("RPC_PROXY_IRN_KEY", "key"),
+            ("RPC_PROXY_IRN_NAMESPACE", "namespace"),
+            ("RPC_PROXY_IRN_NAMESPACE_SECRET", "namespace"),
         ];
 
         values.iter().for_each(set_env_var);
@@ -241,6 +250,12 @@ mod test {
                     max_tokens: Some(100),
                     refill_interval_sec: Some(1),
                     refill_rate: Some(10),
+                },
+                irn: IRNConfig {
+                    node: Some("node".to_owned()),
+                    key: Some("key".to_owned()),
+                    namespace: Some("namespace".to_owned()),
+                    namespace_secret: Some("namespace".to_owned()),
                 },
             }
         );
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -6,7 +6,7 @@ use {
         profiler::ProfilerConfig,
         project::{storage::Config as StorageConfig, Config as RegistryConfig},
         providers::{ProviderKind, ProvidersConfig, Weight},
-        storage::irn::Config as IRNConfig,
+        storage::irn::Config as IrnConfig,
         utils::rate_limit::RateLimitingConfig,
     },
     serde::de::DeserializeOwned,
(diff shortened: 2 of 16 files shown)


src/storage/irn/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        node_addr: String,
+        key_base64: String,
+        namespace: String,
+        namespace_secret: String,
+    ) -> Result<Self, StorageError> {
+        let key = irn_api::auth::client_key_from_bytes(
+            key_base64.as_bytes(),
+            irn_api::auth::Encoding::Base64,
+        )
+        .map_err(|_| StorageError::WrongCredentialsFormat("wrong key is provided".to_string()))?;
```
chris13524, line 39: suggestion: Separate the `WrongCredentialsFormat` into 2 errors instead of passing a string to it

src/storage/irn/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        let peer_id = irn_api::auth::peer_id(&key.verifying_key());
+        let node_addr = node_addr.parse::<SocketAddr>().map_err(|_| {
+            StorageError::WrongArgument("node_addr is not in socket address format".to_string())
```
chris13524, line 42: Similar here: just make a variant for this specific error and pass the original error

src/storage/irn/mod.rs
geekbrother: Makes sense, thanks! Added in https://github.com/WalletConnect/blockchain-api/pull/681/commits/f44679c7a277dba56adebf108575a8a26abc2aa2

src/storage/irn/mod.rs
geekbrother: Makes sense, thanks! Added in https://github.com/WalletConnect/blockchain-api/pull/681/commits/f44679c7a277dba56adebf108575a8a26abc2aa2

src/storage/irn/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        key_base64: String,
+        namespace: String,
+        namespace_secret: String,
+    ) -> Result<Self, StorageError> {
+        let key = irn_api::auth::client_key_from_bytes(
+            key_base64.as_bytes(),
+            irn_api::auth::Encoding::Base64,
+        )
+        .map_err(|_| StorageError::WrongKey(key_base64))?;
+        // IRN connection error Network(ConnectionHandler(NoAvailablePeers)) when using
```
geekbrother, line 44: @heilhead I have `NoAvailablePeers` when generating `peer_id` from the existing key. It works when generating a new one. Any idea where I am wrong? 🤔

heilhead: As discussed on slack, you're trying to use the client peer ID for server address, and this isn't supported by our network machinery. The proper server peer IDs should be parsed as part of the node address, see the following [relay code](https://github.com/WalletConnect/rs-relay/blob/c788b52d2a05f2314894e5904f96f6737b9de7b8/src/config/mod.rs#L152-L164) on how to parse them from a `{PEER_ID}-{MULTIADDR}` string.

.env.example
```diff
@@ -31,3 +31,9 @@ export RPC_PROXY_POSTGRES_URI="postgres://postgres@localhost/postgres"
 # export RPC_PROXY_RATE_LIMITING_MAX_TOKENS=100
 # export RPC_PROXY_RATE_LIMITING_REFILL_INTERVAL_SEC=1
 # export RPC_PROXY_RATE_LIMITING_REFILL_RATE=2
+
+# Uncomment for using the IRN client
+# export RPC_PROXY_IRN_NODE=127.0.0.1:3011
```
xDarksome, line 36: I suggest using multiple IRN nodes, as a fall-back mechanism. One node can have a downtime even under normal conditions

geekbrother: Makes sense. Good follow-up. Created it as https://github.com/WalletConnect/blockchain-api/issues/685

xDarksome:
> The record TTL is 90 days. (Which should be ok for sessions).
does it need to be 90 days, our current longest living data is 30 day

I'm talking from the perspective of data migrations, if our highest value is going to be 90 days, then next time we decide to do full migration we would need to shadow the data for 90 days to be sure

And it probably will happen in the next 3 month, as we will be transitioning to the permissioned network

geekbrother:
> > The record TTL is 90 days. (Which should be ok for sessions).
> 
> does it need to be 90 days? our current longest living data is 30 days
> 
> I'm talking from the perspective of data migrations, if our highest value is going to be 90 days, then next time we decide to do full migration we would need to shadow the data for 90 days to be sure
> 
> And it probably will happen in the next 3 month, as we will be transitioning to the permissioned network

It decreased to 30 days. The context: [slack](https://walletconnect.slack.com/archives/C0760BM9AHH/p1719322586648969).

src/env/mod.rs
```diff
@@ -48,6 +49,7 @@ pub struct Config {
     pub profiler: ProfilerConfig,
     pub providers: ProvidersConfig,
     pub rate_limiting: RateLimitingConfig,
+    pub irn: IRNConfig,
```
xDarksome, line 52: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

> In CamelCase, acronyms count as one word: use Uuid rather than UUID. In snake_case, acronyms are lower-cased: is_xid_start.

geekbrother: Hmm, Clippy missed this. Don't fully rely on clippy 😅 Fixed in https://github.com/WalletConnect/blockchain-api/pull/681/commits/87fb41f150c2c15ba8bd806c7f5d3e2fdd847b5c. Thanks!

src/storage/irn/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+            .await
+            .map_err(StorageError::IrnClientError)
+    }
+
+    /// Get a value from the storage
+    pub async fn get(&self, key: String) -> Result<Option<String>, StorageError> {
+        let result = self.client.get(self.key(key.as_bytes().into())).await;
+
+        match result {
+            Ok(Some(data)) => match String::from_utf8(data) {
```
xDarksome, line 107: Why are you returning a `String` instead of a `Vec<u8>`? I'm going to guess that it's going to be deserialized later anyway

xDarksome (approved):

PR #682, opened 2024-06-21
feat(terraform): IRN VPC Peering

xDarksome:
# Description

Adds VPC peering to the IRN VPC

## How Has This Been Tested?

Manually

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/res_network.tf
+++ b/terraform/res_network.tf
@@ -38,6 +38,22 @@ module "vpc" {
   vpc_flow_log_tags         = module.this.tags
 }
 
+resource "aws_vpc_peering_connection" "irn" {
+  count = var.irn_vpc_id != null ? 1 : 0
+
+  vpc_id      = module.vpc.vpc_id
+  peer_vpc_id = var.irn_vpc_id
+  auto_accept = true
+}
+
+resource "aws_route" "irn" {
+  count = var.irn_vpc_id != null ? length(module.vpc.private_route_table_ids) : 0
+
+  route_table_id            = module.vpc.private_route_table_ids[count.index]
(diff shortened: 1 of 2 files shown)


terraform/res_network.tf
```diff
@@ -38,6 +38,22 @@ module "vpc" {
   vpc_flow_log_tags         = module.this.tags
 }
 
+resource "aws_vpc_peering_connection" "irn" {
+  count = var.irn_vpc_id != null ? 1 : 0
```
chris13524, line 42: I think this should be required not optional?

chris13524 (commented):
What are the CIDR ranges of the Blockchain API and IRN VPCs? Just curious how this fits together

terraform/res_network.tf
xDarksome: Don't know, I made it optional for relay, because it may not be needed for all envs. 
@geekbrother do we need it always here?

xDarksome:
> What are the CIDR ranges of the Blockchain API and IRN VPCs? Just curious how this fits together
Fortunately, fits fine. `10.0.0.0` here, `10.2.0.0` in relay

terraform/res_network.tf
geekbrother: > Don't know, I made it optional for relay, because it may not be needed for all envs. @geekbrother do we need it always here?

Since sessions will be a part of the blockchain-api and no plans to run it in a cloud without it, I believe we can make this required.

geekbrother (approved):
Thanks! Looks good to me. Can't test this now, but adding the test to connect to the IRN to #681 so we will be sure that works.

PR #688, opened 2024-06-25
fix(Dockerfile): moving copy before the `chef cook`

geekbrother:
# Description

This PR moves files copying *before* the `chef cook` to copy the `irn` submodule files and fix [the build error](https://github.com/WalletConnect/blockchain-api/actions/runs/9669791859/job/26677120309#step:8:372).

## How Has This Been Tested?

Tested by running `docker build --no-cache -t test_app .` locally and successfully built the image with the change from this PR.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,9 +48,10 @@ ARG                 BUILD_PROFILE="--profile ${PROFILE}"
 WORKDIR             /app
 # Cache dependancies
 COPY --from=plan    /app/recipe.json recipe.json
-COPY                . .
+COPY                irn ./irn
 RUN                 cargo chef cook ${BUILD_PROFILE} --recipe-path recipe.json 
 # Build the local binary
+COPY                . .
 RUN                 cargo build --bin rpc-proxy ${RELEASE}
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


chris13524 (changes requested):
This breaks the purpose of using `cargo chef` i.e. caching dependency builds.

Instead, add a new `COPY /irn /irn` and defer the copying of the source code until _after_ the `cargo chef cook` command.

geekbrother:
> This breaks the purpose of using `cargo chef` i.e. caching dependency builds.
> 
> Instead, add a new `COPY /irn /irn` and defer the copying of the source code until _after_ the `cargo chef cook` command.

Thanks for pointing out this!
Weird, but adding the copying of the `irn` along with the `src` [at the first stage](https://github.com/WalletConnect/blockchain-api/blob/3c0302cf1695ff87250b4a14e0eefc25221468bd/Dockerfile#L34) didn't work. I've added just copying of the `irn` instead of all in https://github.com/WalletConnect/blockchain-api/pull/688/commits/97aba24cdd8a811b3a10545a181b7fe511d4862c.

chris13524 (approved):

PR #690, opened 2024-06-26
feat(network): using different CIDR for the prod and staging environments

geekbrother:
# Description

This PR adds using of different CIDR networks for the production and staging environments.
Depending on the stage:
* `10.0.0.0/16` for production (same as at the moment),
* `13.0.0.0/16` for staging.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/res_network.tf
+++ b/terraform/res_network.tf
@@ -1,5 +1,5 @@
 locals {
-  vpc_cidr                = "10.0.0.0/16"
+  vpc_cidr                = module.this.stage == "prod" ? "10.0.0.0/16" : "13.0.0.0/16"
   vpc_azs                 = slice(data.aws_availability_zones.available.names, 0, 3)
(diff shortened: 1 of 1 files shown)


terraform/res_network.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  vpc_cidr                = "10.0.0.0/16"
+  vpc_cidr                = module.this.stage == "prod" ? "10.0.0.0/16" : "13.0.0.0/16"
```
chris13524, line 2: Can you perhaps use 13 and 14 instead? `10.0.0.0/16` while doesn't' currently conflict, is quite close to `10.1.0.0/16` that the relay uses: https://www.notion.so/walletconnect/Private-IP-range-allocation-5fba8350d0a9453ca589dfa73affd508?pvs=4

chris13524 (approved):

geekbrother:
Closing this since the IRN now uses public IPs.

PR #699, opened 2024-07-08
feat(sessions): implementing permission revoking

geekbrother:
# Description

This PR implements the permission context revoking endpoint `/v1/sessions/{address}/revoke` according to the [API SPEC draft](https://github.com/WalletConnect/walletconnect-specs/pull/242).

For the request authentication, the signature (signed message) is used. 
As a signing message, the permission controller unique identifier is used (PCI) and signed by the signing key (created during the session creation request). 
Then the signature is verified at the server by the verification key which is stored during the [session creation](https://github.com/WalletConnect/walletconnect-specs/pull/242/files#diff-585df27a5f0c0b1216c7296d8bbe3c30877bc469bcbdad4c905c6803c1e441dcR53) in the permission session object.

## How Has This Been Tested?

* [A new integration test was implemented](https://github.com/WalletConnect/blockchain-api/pull/699/files#diff-d2b6bfa16ff410867b43a05eebec7cf74f434e26320f85fdf4681537ebb32372).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/sessions/context.rs
+++ b/src/handlers/sessions/context.rs
@@ -3,6 +3,7 @@ use {
     crate::{
         error::RpcError,
         state::AppState,
+        storage::irn::OperationType,
         utils::crypto::{disassemble_caip10, json_canonicalize, verify_ecdsa_signature},
     },
     axum::{
@@ -41,7 +42,9 @@ async fn handler_internal(
         .hget(address.clone(), request_payload.pci.clone())
         .await?
         .ok_or_else(|| RpcError::PermissionNotFound(request_payload.pci.clone()))?;
-    state.metrics.add_irn_latency(irn_call_start, "hget".into());
+    state
+        .metrics
+        .add_irn_latency(irn_call_start, OperationType::Hget.into());
     let mut storage_permissions_item =
         serde_json::from_str::<StoragePermissionsItem>(&storage_permissions_item)?;
 
@@ -66,7 +69,9 @@ async fn handler_internal(
             serde_json::to_string(&storage_permissions_item)?.into(),
         )
         .await?;
-    state.metrics.add_irn_latency(irn_call_start, "hset".into());
+    state
+        .metrics
+        .add_irn_latency(irn_call_start, OperationType::Hset.into());
 
     Ok(Json(storage_permissions_item).into_response())
 }
--- a/src/handlers/sessions/context.rs
+++ b/src/handlers/sessions/context.rs
@@ -44,7 +44,7 @@ async fn handler_internal(
         .ok_or_else(|| RpcError::PermissionNotFound(request_payload.pci.clone()))?;
     state
         .metrics
-        .add_irn_latency(irn_call_start, OperationType::Hget.into());
+        .add_irn_latency(irn_call_start, OperationType::Hget);
     let mut storage_permissions_item =
         serde_json::from_str::<StoragePermissionsItem>(&storage_permissions_item)?;
 
@@ -71,7 +71,7 @@ async fn handler_internal(
         .await?;
     state
         .metrics
-        .add_irn_latency(irn_call_start, OperationType::Hset.into());
+        .add_irn_latency(irn_call_start, OperationType::Hset);
 
     Ok(Json(storage_permissions_item).into_response())
 }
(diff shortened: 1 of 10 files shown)


src/handlers/sessions/revoke.rs
```diff
@@ -0,0 +1,61 @@
...
+    // Checking the CAIP-10 address format
+    disassemble_caip10(&address)?;
+
+    // Get the PCI object from the IRN
+    let irn_call_start = SystemTime::now();
+    let storage_permissions_item = irn_client
+        .hget(address.clone(), request_payload.pci.clone())
+        .await?
+        .ok_or_else(|| RpcError::PermissionNotFound(request_payload.pci.clone()))?;
+    state.metrics.add_irn_latency(irn_call_start, "hget".into());
```
chris13524, line 44: Would be good to make this `"hget"` thing an enum

src/handlers/sessions/revoke.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    // Checking the CAIP-10 address format
+    disassemble_caip10(&address)?;
+
+    // Get the PCI object from the IRN
+    let irn_call_start = SystemTime::now();
+    let storage_permissions_item = irn_client
+        .hget(address.clone(), request_payload.pci.clone())
+        .await?
+        .ok_or_else(|| RpcError::PermissionNotFound(request_payload.pci.clone()))?;
```
chris13524, line 43: Shouldn't this result in a 2xx status code rather than a 4xx one? Idempotency?

chris13524 (changes requested):
I think this needs to be idempotent

src/handlers/sessions/revoke.rs
geekbrother: The user trying to revoke permission which does not exist, so the passed PCI is not correct, that's why we are returning `Bad Request` here, pointing out that the passed parameter is wrong.

src/handlers/sessions/revoke.rs
geekbrother: Good catch, we should get rid of magic string. Fixed in 4cf2436ddb2982e35acc239be410694daace0ea6 by implementing `OperationType` enum.

src/handlers/sessions/revoke.rs
chris13524: What if it was already revoked but the client didn't get back the response (e.g. error in network), then it would try to revoke a second time and then get an error it doesn't expect.

Why not just return OK if it's already revoked?

src/handlers/sessions/context.rs
```diff
@@ -41,7 +42,9 @@ async fn handler_internal(
         .hget(address.clone(), request_payload.pci.clone())
         .await?
         .ok_or_else(|| RpcError::PermissionNotFound(request_payload.pci.clone()))?;
-    state.metrics.add_irn_latency(irn_call_start, "hget".into());
+    state
+        .metrics
+        .add_irn_latency(irn_call_start, OperationType::Hget.into());
```
chris13524, line 47: Good to change to enum but I would go a step further and convert the enum into a string _inside_ `add_irn_latency()`

chris13524 (approved):
I think the revoke endpoint should be idempotent

src/handlers/sessions/revoke.rs
geekbrother: Let's return OK, but shoot a `warning!` in this case to track possible issues with that.

geekbrother:
> I think the revoke endpoint should be idempotent

I missed that we need to check the request signature later, we need to get the verifying key for that. If the PCI not exists we can't verify/authorize the request.

chris13524:
> I missed that we need to check the request signature later, we need to get the verifying key for that. If the PCI not exists we can't verify/authorize the request.

You only need to verify it if you are going to perform an action. If the action is already performed then you are good.

geekbrother:
> > I missed that we need to check the request signature later, we need to get the verifying key for that. If the PCI not exists we can't verify/authorize the request.
> 
> You only need to verify it if you are going to perform an action. If the action is already performed then you are good.

Changed it. We are returning [OK and warn on it](https://github.com/WalletConnect/blockchain-api/pull/699/commits/8fd31b1f6d383ce14128e44d79b4c7f1812cd200) now.

src/handlers/sessions/context.rs
geekbrother: Done https://github.com/WalletConnect/blockchain-api/pull/699/commits/3ddc187089112ae5f55339a813a3de4bbe815ff4. Thanks!

PR #701, opened 2024-07-09
feat(sessions): storing a signing key during permission creation

geekbrother:
# Description

This PR adds a signing key to the storage along with the verifying key during the permission creation request. The generated signing key will be used for making the signature during the co-signing request.

## How Has This Been Tested?

* Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/sessions/mod.rs
+++ b/src/handlers/sessions/mod.rs
@@ -75,6 +75,7 @@ pub struct StoragePermissionsItem {
     permissions: PermissionItem,
     context: Option<PermissionContextItem>,
     verification_key: String,
+    signing_key: String,
 }
 
 /// Permission revoke request schema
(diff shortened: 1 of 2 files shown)


src/handlers/sessions/mod.rs
```diff
@@ -69,6 +69,7 @@ pub struct StoragePermissionsItem {
     permissions: PermissionItem,
     context: Option<PermissionContextItem>,
     verification_key: String,
+    signing_key: String,
```
chris13524, line 72: Won't this result in deserialization errors for existing entries?

geekbrother: We are not using existing items yet, so it's safe to change.

chris13524 (approved):

PR #703, opened 2024-07-11
feat(providers): adding Solana devnet and testnet to supported chains

geekbrother:
# Description

This PR adds Solana devnet `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1` and testnet `4uhcVJyU9pJkvQyS88uRDiswHXSCkY3z` to the supported chains list. 
Quicknode is used as a provider for these two new supported chains.

The following changes for the Quicknode provider implementation were made:
* `RPC_PROXY_QUICKNODE_API_TOKEN` environment variable changed to `RPC_PROXY_QUICKNODE_API_TOKENS`
* The `RPC_PROXY_QUICKNODE_API_TOKENS` now represents the JSON array with the `{"chainId":"token"}` format, since Quicknode generates the unique token for each RPC endpoint/chainId.

[Solana namespace CAIP spec](https://namespaces.chainagnostic.org/solana/caip10).

## How Has This Been Tested?

* The [test for the Quicknode provider](https://github.com/WalletConnect/blockchain-api/pull/703/files#diff-1a868ca223307cd8271be3bc8e873d830e54f7e2a0e08030f46edd66ca02b935) was updated and successfully passed.

## Due Diligence

* [ ] Remove the old `QUICKNODE_API_TOKEN` terraform environment variable after deploy.
* [ ] Breaking change


src/env/quicknode.rs
```diff
@@ -33,17 +34,74 @@ impl ProviderConfig for QuicknodeConfig {
...
-            (
-                "snowy-chaotic-hill.zksync-mainnet".into(),
-                Weight::new(Priority::High).unwrap(),
-            ),
+            "snowy-chaotic-hill.zksync-mainnet".into(),
+        ),
+        // Solana Mainnet
+        (
+            "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp".into(),
+            "indulgent-thrumming-bush.solana-mainnet".into(),
```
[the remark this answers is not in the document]
geekbrother: It's uniquely generated for us 😅

src/env/quicknode.rs
```diff
@@ -33,17 +34,74 @@ impl ProviderConfig for QuicknodeConfig {
...
-fn default_supported_chains() -> HashMap<String, (String, Weight)> {
+fn extract_supported_chains(access_tokens_json: String) -> HashMap<String, (String, Weight)> {
+    let access_tokens: HashMap<String, String> = match serde_json::from_str(&access_tokens_json) {
+        Ok(tokens) => tokens,
+        Err(_) => {
+            error!(
+                "Failed to parse JSON with API access tokens for QuickNode provider. Using empty \
+                 tokens."
+            );
+            return HashMap::new();
```
chris13524, line 45: I feel like this should prevent startup of the service, that way we don't continue with broken config. The ECS rollout won't shift traffic to the new version until it is healthy.

geekbrother: Good point to make this consistent across all providers. 
Some providers' tokens are expected and break the startup and some are optional like here, GetBlock, etc. We should make this consistent as a follow-up.

src/providers/quicknode.rs
```diff
@@ -50,12 +50,20 @@ impl RateLimited for QuicknodeProvider {
...
 
-        let uri = format!("https://{}.quiknode.pro/{}", chain, self.api_token);
+        // Get the chain subdomain
+        let chain_subdomain =
+            self.chain_subdomains
+                .get(chain_id)
+                .ok_or(RpcError::InvalidConfiguration(format!(
+                    "Quicknode subdomain not found for chainId: {}",
+                    chain_id
+                )))?;
```
chris13524, line 65: It would seem to me this error shouldn't happen, because `proxy()` should only be called with chains that are supported?

It would be nice to remove this, I suppose that could be done with an enum per-provider and then checking supported chains would happen by converting the string to the enum.

chris13524 (approved):

src/providers/quicknode.rs
geekbrother: This can happen if we miss to fulfill the subdomain for the supported chain somehow, so we will know what's happened. I've made changes in https://github.com/WalletConnect/blockchain-api/pull/703/commits/76b6c3cfb54aa27497ba9b14f16a2b47fd9a7d39 to use a single HashMap for supported chains and subdomains, instead of two. This will eliminate possible mistakes when adding/removing chains and this error.

PR #704, opened 2024-07-11
feat(identity): enable browser caching

### LINKED ISSUE #257: feat(identity): browser caching
Consider setting a `Cache-Control` header so that identity responses can be cached locally for a time.

What's a good value for this? 1 minute? 1 hour?

This is different from the Redis cache and cache invalidation from blockchain events which is tracked here: https://github.com/WalletConnect/rpc-relay/issues/6

1 hour may be too long, especially if used for transactions. 1 minute may be sufficient as an easy win. The identity API is pretty fast already so this is not important.

chris13524:
# Description

Adds `Cache-Control` header with a value that matches the current TTL of the Redis cache.

Resolves #257 

## How Has This Been Tested?

New unit tests

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -136,7 +136,7 @@ async fn handler_internal(
         // Only happens during initial rollout when `resolved_at` is None, so we don't need to go overboard on the cache
         .unwrap_or(TimeDelta::hours(1))
         .num_seconds();
-    let cache_control = format!("max-age={ttl_secs} s-maxage={ttl_secs}");
+    let cache_control = format!("public, max-age={ttl_secs}, s-maxage={ttl_secs}");
 
     Ok(([(CACHE_CONTROL, cache_control)], Json(res)).into_response())
 }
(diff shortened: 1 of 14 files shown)


arein (approved):
LGTM - consider also covering in integration tests

src/handlers/identity.rs
```diff
@@ -239,9 +258,9 @@ async fn lookup_identity(
             debug!("Saving to cache");
             let cache = cache.clone();
             let res = res.clone();
-            let cache_ttl = Duration::from_secs(60 * 60 * 24);
             // Do not block on cache write.
             tokio::spawn(async move {
+                let cache_ttl = CACHE_TTL.to_std().expect("valid duration");
```
geekbrother, line 263: Probably `invalid duration` not valid?

geekbrother (approved):
lgtm, thanks!

src/handlers/identity.rs
chris13524: Thanks, fixed in 33c224e67baf209a785815417d499038806b57c4

PR #705, opened 2024-07-12
fix(identity): adding version to the cache key format

geekbrother:
# Description

This PR updates the Redis identity resolver results cache key format into `${key}-v1` to add the keys version and update the current storage schema (the current records can't be migrated due to the format of the messagePack).

## How Has This Been Tested?

* Tested locally by passing the integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -184,6 +184,7 @@ async fn lookup_identity(
     headers: HeaderMap,
 ) -> Result<(IdentityLookupSource, IdentityResponse), RpcError> {
     let address_with_checksum = to_checksum(&address, None);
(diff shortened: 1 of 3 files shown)


src/handlers/identity.rs
```diff
@@ -184,6 +184,7 @@ async fn lookup_identity(
     headers: HeaderMap,
 ) -> Result<(IdentityLookupSource, IdentityResponse), RpcError> {
     let address_with_checksum = to_checksum(&address, None);
+    let cache_record_key = format!("{}-v1", address_with_checksum.clone());
```
chris13524, line 187: You shouldn't need `clone()` here

PR #706, opened 2024-07-12
feat(readme): adding manual integration tests run description

geekbrother:
# Description

This PR adds a description to the `README.md` file on how to run manual integration tests.

## How Has This Been Tested?

* Command in the description was tested locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ The following environment variables are expected to be passed:
 If you want to run certain test from the suite you can pass `-t` with the name of the
 test or test suite to run.
 
+For passing `Exact provider request` and disabled cache tests you need to provide a valid `RPC_PROXY_TESTING_PROJECT_ID`.
+
 ### Docker
 
 ```console
--- a/README.md
+++ b/README.md
@@ -50,6 +50,17 @@ curl -X POST "http://localhost:3000/v1?chainId=eip155:1&projectId=someid" --data
 just devloop
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +50,17 @@ curl -X POST "http://localhost:3000/v1?chainId=eip155:1&projectId=someid" --data
 just devloop
 ```
 
+### Manual integrations test run
+
+You can run manual integration tests by invoking the `yarn integration` command. 
+The following environment variables are expected to be passed:
+
+* `RPC_URL` - URL of the server to test. Use `http://localhost:3000` for the local testing.
+* `PROJECT_ID` - Unique project identifier.
```
chris13524, line 59: Does this have to match `RPC_PROXY_TESTING_PROJECT_ID`?

chris13524 (approved):

README.md
geekbrother: Not realy, because `RPC_PROXY_TESTING_PROJECT_ID` is just for internal use only, for the exact provider call, which is not permitted for all Project IDs other than that. Makes sense to mention this in this doc also.

chris13524: It's also used for bypass cache, which is used in the test cases. So if these don't match not all of the test cases will pass?

geekbrother: Correct! I'll update the doc with this also 👍

PR #707, opened 2024-07-13
feat(sessions): co-signer endpoint implementation

geekbrother:
# Description

This PR implements mvp/alpha version of the Co-Signer session permission endpoint according to the [draft API spec](https://github.com/WalletConnect/walletconnect-specs/pull/244/files).

The simplified (mvp/testing) version of Co-Signer performs the following operations:
* Verifying the user signature in the UserOperation according to the ERC-4337.
* Pull the signing key for the PCI and make a signature of the UserOperation.
* Concat the PCI signature and user signature then update the signature in the UserOperation.
* Send the UserOperation to the bundler and return the operation receipt.

## How Has This Been Tested?

* Manually tested, integration test TBD as a follow-up since the initiator should sponsor the transaction for the bundler and I'm still figuring out how to do this best.

## Due Diligence

* [ ] Breaking change


src/handlers/sessions/cosign.rs
```diff
@@ -0,0 +1,161 @@
...
+        ))
+    })?;
+    let signature: DerSignature = signer.sign(user_op_hash.as_bytes());
+    let signature_hex = hex::encode(signature.as_bytes());
+
+    // Concat permission and user signature, update the userOperation signature
+    let concatenated_signature = format!("{}{}", signature_hex, request_payload.user_op.signature);
+    user_op.signature = concatenated_signature;
+
+    // Using the Biconomy bundler to send the userOperation
```
geekbrother, line 133: This part should be extracted to `providers` to support multiple bundlers in the future and more clean code.

src/utils/crypto.rs
```diff
@@ -450,6 +468,57 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+    bundler_url: &str,
+    entry_point: &str,
+    simulation_type: &str,
+) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
+    let bundler_client = Client::new();
+
+    // Send the UserOperation to the bundler
+    let response: serde_json::Value = bundler_client
+        .post(bundler_url)
+        .json(&json!({
```
chris13524, line 483: A proper `struct` here would be nice vs the `json!` macro

src/utils/crypto.rs
```diff
@@ -450,6 +468,57 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+    // Check if there was an error in the response
+    if response.get("error").is_some() {
+        return Err(Box::new(std::io::Error::new(
+            std::io::ErrorKind::Other,
+            response["error"].to_string(),
+        )));
+    }
+
+    // Get the transaction hash from the response
+    let tx_hash = response["result"].as_str().ok_or("No result in response")?;
```
chris13524, line 503: Same here, `struct`

src/handlers/sessions/cosign.rs
```diff
@@ -0,0 +1,161 @@
...
+        serde_json::from_str::<StoragePermissionsItem>(&storage_permissions_item)?;
+
+    // Sign the user operation with the permission signing key
+    let signing_key = BASE64_STANDARD
+        .decode(storage_permissions_item.signing_key)
+        .map_err(|e| RpcError::WrongBase64Format(e.to_string()))?;
+    let signer = SigningKey::from_pkcs8_der(&signing_key).map_err(|e| {
+        RpcError::EcdsaError(format!(
+            "Error during conversion signing key to pkcs8 DER format: {:?}",
+            e
```
chris13524, line 123: I would prefer if we passed the error directly, rather than formatting it as a string

src/handlers/sessions/cosign.rs
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    // Send the userOperation to the bundler and get the receipt
+    let receipt =
+        send_user_operation_to_bundler(&user_op, &bundler_url, entry_point, simulation_type)
+            .await
+            .map_err(|e| {
+                RpcError::BundlerError(format!(
+                    "Error during sending the user operation to the bundler: {:?}",
+                    e
```
chris13524, line 156: Same here, would prefer if we avoided converting error to string and passed it all the way up the stack

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+    simulation_type: &str,
+) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
+    let bundler_client = Client::new();
+
+    // Send the UserOperation to the bundler
+    let jsonrpc_send_userop_request = JsonRpcRequest {
+        id: "1".to_string(),
+        jsonrpc: "2.0".to_string(),
+        method: "eth_sendUserOperation".to_string(),
+        params: json!([user_op, entry_point, {"simulation_type": simulation_type}]),
```
chris13524, line 494: Can we get a `struct` for this instead of using `json!`

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
     balance_f64 * price
 }
 
+/// Function to send UserOperation to the bundler and return receipt
+pub async fn send_user_operation_to_bundler(
+    user_op: &UserOperation,
+    bundler_url: &str,
+    entry_point: &str,
+    simulation_type: &str,
+) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
```
chris13524, line 486: Would prefer if we had a `thiserror` enum instead of using `Box<Error>`

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+    let response: serde_json::Value = bundler_client
+        .post(bundler_url)
+        .json(&jsonrpc_send_userop_request)
+        .send()
+        .await?
+        .json()
+        .await?;
+
+    // Check if there was an error in the response
+    if response.get("error").is_some() {
```
chris13524, line 505: Same here, would prefer a `struct` instead of dynamic stuff

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+        )));
+    }
+
+    // Get the transaction hash from the response
+    let tx_hash = response["result"].as_str().ok_or("No result in response")?;
+
+    // Get the transaction receipt
+    let jsonrpc_get_receipt_request = JsonRpcRequest {
+        id: "1".to_string(),
+        jsonrpc: "2.0".to_string(),
```
chris13524, line 518: Using a `once_cell::sync::Lazy<Arc<String>>` would be a nice optimization here to avoid extra allocations. E.g. https://github.com/WalletConnect/WalletConnectRust/blob/edd6b6bcaa360e0887b4930ba78d56c958700612/relay_rpc/src/rpc.rs#L21

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+        method: "eth_getTransactionReceipt".to_string(),
```
chris13524, line 519: Same here, `once_cell::sync::Lazy<Arc<String>>`

src/handlers/sessions/cosign.rs
nopestack: if using anyhow you can call `.context()` on the result and add the message there, this wraps the original error and adds the addional context you provided @geekbrother

src/utils/crypto.rs
```diff
@@ -450,6 +477,59 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
     balance_f64 * price
 }
 
+/// Function to send UserOperation to the bundler and return receipt
+pub async fn send_user_operation_to_bundler(
```
nopestack, line 481: please use anyhow if you don't care about type erasing the underlying errors from reqwest or other deps and wrap them using `.context()` otherwise please create proper error types using `thiserror`

src/handlers/sessions/cosign.rs
geekbrother: Makes sense. Thanks. Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6

src/handlers/sessions/cosign.rs
geekbrother: Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6. Thanks!

src/utils/crypto.rs
geekbrother: Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6 thanks!

src/utils/crypto.rs
geekbrother: Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6 Thanks!

src/utils/crypto.rs
geekbrother: [Fixed](https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6). Thanks!

src/utils/crypto.rs
geekbrother: [Fixed](https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6). Thanks!

src/utils/crypto.rs
geekbrother: Changed to `thiserror` in https://github.com/WalletConnect/blockchain-api/pull/707/commits/41b4afe27a19f6dd477ddd256e10507c24319eb6. Thanks!

src/utils/crypto.rs
geekbrother: Hmm, good point. We can probably try to deserialize the result or error structs and handle the structure we got.

src/utils/mod.rs
```diff
@@ -1,5 +1,22 @@
...
 pub mod network;
 pub mod rate_limit;
 pub mod suggestions;
+
+pub fn generate_random_string(len: usize) -> String {
+    let rng = rand::thread_rng();
+    rng.sample_iter(&Alphanumeric)
+        .filter_map(|b| {
+            let c = b as char;
+            if c.is_ascii_alphanumeric() && c.is_ascii_lowercase() || c.is_ascii_digit() {
```
chris13524, line 14: Why filter out uppercase character here? Seems unnecessary and results in non-deterministic ID generation speed

src/handlers/sessions/cosign.rs
```diff
@@ -0,0 +1,152 @@
...
+    // Checking the CAIP-10 address format
+    let (namespace, chain_id, address) = disassemble_caip10(&address)?;
+    if namespace != CaipNamespaces::Eip155 {
+        return Err(RpcError::UnsupportedNamespace(namespace));
+    }
+    let chain_id_caip2 = format!("{}:{}", namespace, chain_id);
+
+    // Verify the user signature
+    let mut user_op = request_payload.user_op.clone();
+    let user_op_hash = keccak256(
```
chris13524, line 62: I don't think has any benefit here but with alloy I think keccak256 accepts a bytes iterator which means you could chain the bytes of each of these strings rather than allocating the string.

src/handlers/sessions/cosign.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    // Concat permission and user signature, update the userOperation signature
+    let concatenated_signature = format!("{}{}", signature_hex, request_payload.user_op.signature);
+    user_op.signature = concatenated_signature;
+
+    // Using the Biconomy bundler to send the userOperation
+    let entry_point = "0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789";
+    let simulation_type = "validation_and_execution";
+    let bundler_url = format!(
+        "https://bundler.biconomy.io/api/v2/{}/{}",
```
chris13524, line 134: Might be good to make these constants or something with link to docs

src/utils/crypto.rs
```diff
@@ -442,6 +500,64 @@ pub fn convert_token_amount_to_value(balance: U256, price: f64, decimals: u32) -
...
+    user_op: &UserOperation,
+    bundler_url: &str,
+    entry_point: &str,
+    simulation_type: &str,
+) -> Result<serde_json::Value, CryptoUitlsError> {
+    let bundler_client = Client::new();
```
chris13524, line 510: Can we reuse this client to avoid establishing a new connection each time?

chris13524 (approved):
Looks good to me! Sorry I guess I forgot to approve this before after leaving all the nits 🤣

src/handlers/sessions/cosign.rs
geekbrother: Agree here. I'm planning to migrate to alloy soon anyway.

src/handlers/sessions/cosign.rs
geekbrother: It's makes sense for the prod or stable, but in MVP at the moment this is probably will change and we are still thinking about the bundler repository or list to use...

src/utils/mod.rs
geekbrother: Good catch! 
This function was moved from the test utils where we used it for the random names generation. Since we are using it not only there it makes sense to allow any case. Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/3fb9038740b4799e57d9eeccd07466aa9361e4cc
Thanks!

src/utils/crypto.rs
geekbrother: Good optimization catch. Fixed in https://github.com/WalletConnect/blockchain-api/pull/707/commits/3fb9038740b4799e57d9eeccd07466aa9361e4cc
Thanks!

PR #709, opened 2024-07-17
fix(sessions): changing to use secp256k1, removing signatures from context update and revoking

geekbrother:
# Description

This PR makes the following changes based on [this Slack thread](https://walletconnect.slack.com/archives/C0760BM9AHH/p1721110736693969):

* Removing the signature authorization from the context update and revoking endpoints.
* Using the secp256k1 instead of P256 (secp256r1) during the PCI creation and returns the public, not a private key.
* Using keccak256 for message hashing.
* Removing unused dependencies.

## How Has This Been Tested?

* Updated integration and unit tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -3,17 +3,15 @@ use {
     alloy_primitives::Address,
     base64::prelude::*,
     ethers::{
+        core::k256::ecdsa::{signature::Verifier, Signature, VerifyingKey},
         prelude::abigen,
         providers::{Http, Middleware, Provider},
         types::{H160, H256, U256},
+        utils::keccak256,
     },
-    olpc_cjson::CanonicalFormatter,
     once_cell::sync::Lazy,
-    p256::ecdsa::{signature::Verifier, DerSignature, VerifyingKey},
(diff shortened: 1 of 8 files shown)


nopestack (approved):

src/utils/crypto.rs
```diff
@@ -151,15 +142,16 @@ pub fn verify_ecdsa_signature(
...
 
-    let signature = DerSignature::from_bytes(
-        &BASE64_STANDARD
-            .decode(signature)
-            .map_err(|_| RpcError::WrongBase64Format(signature.to_string()))?,
-    )
-    .map_err(|e| RpcError::SignatureFormatError(e.to_string()))?;
+    let signature_bytes = &BASE64_STANDARD
+        .decode(signature)
+        .map_err(|_| RpcError::WrongBase64Format(signature.to_string()))?;
```
nopestack, line 147: Consider not ignoring the underlying errors though

geekbrother: Makes sense. Thanks!

PR #711, opened 2024-07-18
fix(grafana): fixing no data irn latency alert and registered names counter

geekbrother:
# Description

This PR consists the following fixes: 
* Fixes IRN latency alert by changing no alerting when there is no data instead of alert.
* Fixes Names counter query.

## How Has This Been Tested?

Tested by putting the query and config to the Grafana UI.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/monitoring/panels/names/registered.libsonnet
+++ b/terraform/monitoring/panels/names/registered.libsonnet
@@ -14,7 +14,7 @@ local targets   = grafana.targets;
 
     .addTarget(targets.prometheus(
       datasource  = ds.prometheus,
-      expr        = 'rate(account_names_count_total{}[$__rate_interval])',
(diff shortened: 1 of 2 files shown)


terraform/monitoring/panels/names/registered.libsonnet
```diff
@@ -14,7 +14,7 @@ local targets   = grafana.targets;
 
     .addTarget(targets.prometheus(
       datasource  = ds.prometheus,
-      expr        = 'rate(account_names_count_total{}[$__rate_interval])',
+      expr        = 'account_names_count',
```
geekbrother, line 17: Changing the query because `account_names_count` is an observable gauge not a histogram.

chris13524 (approved):
LGTM

PR #713, opened 2024-07-19
feat(sessions): implementing additional cosigner steps

geekbrother:
# Description

This PR modifying co-signing steps according to the recent changes ([Slack thread](https://walletconnect.slack.com/archives/C0760BM9AHH/p1721314692128899)).

The updated flow:
* Get the `userOperation` hash by calling `getUserOpHash(userOps)` on the v07 entrypoint.
* Sign the resulting hash value with the PCI private key.
* ABI-encode two signatures (co-signer and `userOperation` signature) into one.
* Set the above-encoded signature to the updated `userOperation` and call `getSignature` method on the userOpBuilder contract.
* Set the returned signature from userOpBuilder call to `userOperation`.
* Send the resulting `userOperation` to the bundler.

## How Has This Been Tested?

Tested manually step by step.
Contracts call unit tests that can be [run manually](https://github.com/WalletConnect/blockchain-api/pull/713#discussion_r1684516876) at the moment.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/sessions/cosign.rs
+++ b/src/handlers/sessions/cosign.rs
@@ -1,13 +1,12 @@
 use {
     super::{super::HANDLER_TASK_METRICS, CoSignRequest, StoragePermissionsItem},
     crate::{
-        analytics::MessageSource,
         error::RpcError,
         state::AppState,
         storage::irn::OperationType,
         utils::crypto::{
-            disassemble_caip10, send_user_operation_to_bundler, verify_message_signature,
-            CaipNamespaces,
+            abi_encode_two_bytes_arrays, call_get_signature, call_get_user_op_hash,
+            disassemble_caip10, send_user_operation_to_bundler, CaipNamespaces,
         },
     },
     axum::{
@@ -21,7 +20,7 @@ use {
             ecdsa::{signature::Signer, Signature, SigningKey},
             pkcs8::DecodePrivateKey,
         },
-        utils::keccak256,
+        types::{Bytes, H160},
     },
     serde::{Deserialize, Serialize},
     std::{sync::Arc, time::SystemTime},
@@ -55,28 +54,13 @@ async fn handler_internal(
     if namespace != CaipNamespaces::Eip155 {
         return Err(RpcError::UnsupportedNamespace(namespace));
     }
+    let h160_address = address
+        .parse::<H160>()
+        .map_err(|_| RpcError::InvalidAddress)?;
     let chain_id_caip2 = format!("{}:{}", namespace, chain_id);
-
-    // Verify the user signature
     let mut user_op = request_payload.user_op.clone();
-    let user_op_hash = keccak256(
-        format!(
-            "{}{}{}{}{}{}{}{}{}{}{}",
-            user_op.sender,
-            user_op.nonce,
-            user_op.init_code,
-            user_op.call_data,
-            user_op.call_data,
-            user_op.call_gas_limit,
-            user_op.verification_gas_limit,
-            user_op.pre_verification_gas,
-            user_op.max_fee_per_gas,
-            user_op.max_priority_fee_per_gas,
-            user_op.paymaster_and_data
-        )
-        .as_bytes(),
-    );
-    let user_op_hash = hex::encode(user_op_hash);
+
+    // Project ID for internal json-rpc calls
     let rpc_project_id = state
         .config
         .server
@@ -84,23 +68,23 @@ async fn handler_internal(
         .as_ref()
         .ok_or_else(|| {
             RpcError::InvalidConfiguration(
-                "Missing testing project id in the configuration for eip1271 lookups".to_string(),
+                "Missing testing project id in the configuration for the balance RPC lookups"
+                    .to_string(),
             )
         })?;
-    let sinature_check = verify_message_signature(
-        &user_op_hash,
-        &request_payload.user_op.signature.clone(),
-        &address,
-        &chain_id_caip2,
+
+    // Get the userOp hash
+    // Entrypoint v07 contract address
+    let contract_address = "0x0000000071727De22E5E9d8BAf0edAc6f37da032"
+        .parse::<H160>()
+        .map_err(|_| RpcError::InvalidAddress)?;
+    let user_op_hash = call_get_user_op_hash(
         rpc_project_id,
-        MessageSource::SessionCoSignSigValidate,
+        &chain_id_caip2,
+        contract_address,
+        user_op.clone(),
     )
     .await?;
-    if !sinature_check {
-        return Err(RpcError::SignatureValidationError(
-            "Signature verification error".into(),
-        ));
-    }
 
     // Get the PCI object from the IRN
     let irn_client = state.irn.as_ref().ok_or(RpcError::IrnNotConfigured)?;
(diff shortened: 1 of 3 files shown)


src/utils/crypto.rs
```diff
@@ -728,4 +892,101 @@ mod tests {
...
+
+        let mut context = Vec::new();
+        context.extend_from_slice(&validator_address);
+        context.extend_from_slice(&mode);
+        context.extend_from_slice(&signer_id);
+
+        Bytes::from(context)
+    }
+
+    #[ignore]
```
geekbrother, line 951: Ignored for now and can be run manually by filling the `Project ID`. We need to change the shared CI workflow to pass down the `Project ID` for tests. Will make this as a follow-up to automate the flow along with the new integration test.

src/utils/crypto.rs
```diff
@@ -728,4 +892,101 @@ mod tests {
         )
         .is_err());
     }
+
+    #[ignore]
```
geekbrother, line 896: Ignored for now and can be run manually by filling the `Project ID`. We need to change the shared CI workflow to pass down the `Project ID` for tests. Will make this as a follow-up to automate the flow along with the new integration test.

src/handlers/sessions/cosign.rs
```diff
@@ -55,52 +54,37 @@ async fn handler_internal(
     if namespace != CaipNamespaces::Eip155 {
         return Err(RpcError::UnsupportedNamespace(namespace));
     }
+    let h160_address = address
+        .parse::<H160>()
+        .map_err(|_| RpcError::InvalidAddress)?;
```
nopestack, line 59: consider embedding the original error in the RpcError

src/handlers/sessions/cosign.rs
```diff
@@ -115,18 +99,51 @@ async fn handler_internal(
...
     user_op.signature = concatenated_signature;
 
+    // Get the Signature
+    // UserOpBuilder contract address
+    let user_op_builder_contract_address = permission_context_item
+        .context
+        .signer_data
+        .user_op_builder
+        .parse::<H160>()
+        .map_err(|_| RpcError::InvalidAddress)?;
```
nopestack, line 133: same here

src/utils/crypto.rs
```diff
@@ -95,26 +96,84 @@ struct BundlerSimulationType {
...
-    pub call_data: String,
-    pub call_gas_limit: String,
-    pub verification_gas_limit: String,
-    pub pre_verification_gas: String,
-    pub max_fee_per_gas: String,
-    pub max_priority_fee_per_gas: String,
-    pub paymaster_and_data: String,
-    pub signature: String,
+    pub sender: EthersAddress,
+    pub nonce: U256,
```
nopestack, line 103: weren't these refs to U256 changed to a String before due to a limitation of JS?

src/handlers/sessions/cosign.rs
geekbrother: Thanks, I'm catched this as well. I'll make this as a separate PR since this `RpcError::InvalidAddress` error is used many times and this PR became huge and messy to review.

src/utils/crypto.rs
```diff
@@ -290,6 +349,111 @@ async fn get_balance(
...
+    rpc_project_id: &str,
+    chain_id: &str,
+    contract_address: H160,
+    smart_account_address: H160,
+    user_operation: UserOperation,
+    context: Bytes,
+) -> Result<Bytes, CryptoUitlsError> {
+    abigen!(
+        Safe7579UserOperationBuilder,
+        r#"[
```
nopestack, line 416: consider making this a constant

src/utils/crypto.rs
geekbrother: Correct! But we have missed that ethers U256 can decode it successfully from the string (when testing the frontend), so we can use it again.

src/utils/crypto.rs
```diff
@@ -728,4 +892,101 @@ mod tests {
         )
         .is_err());
     }
+
+    #[ignore]
+    #[tokio::test]
+    async fn test_call_get_user_op_hash() {
+        let rpc_project_id = ""; // Fill the project ID
```
nopestack, line 899: consider adding a `todo!()` so it panics when ignored and the project_id is not set so it's even more obvious

src/utils/crypto.rs
```diff
@@ -728,4 +892,101 @@ mod tests {
...
+        context.extend_from_slice(&mode);
+        context.extend_from_slice(&signer_id);
+
+        Bytes::from(context)
+    }
+
+    #[ignore]
+    #[tokio::test]
+    async fn test_call_get_signature() {
+        let rpc_project_id = ""; // Fill the project ID
```
nopestack, line 954: same here

src/utils/crypto.rs
```diff
@@ -290,6 +349,111 @@ async fn get_balance(
...
+    abigen!(
+        Safe7579UserOperationBuilder,
+        r#"[
+            struct v07UserOperation { address sender; uint256 nonce; bytes initCode; bytes callData; bytes32 accountGasLimits; uint256 preVerificationGas; bytes32 gasFees; bytes paymasterAndData; bytes signature}
+            function getSignature(address smartAccount, v07UserOperation calldata userOperation, bytes calldata context) public view returns (bytes)
+        ]"#,
+    );
+
+    let provider = Provider::<Http>::try_from(format!(
+        "https://rpc.walletconnect.com/v1?chainId={}&projectId={}",
```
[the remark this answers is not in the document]
geekbrother: Thanks! Ideally, this should be extracted into the utils function, because we are using the same a lot.

src/handlers/sessions/cosign.rs
geekbrother: Thanks! Same as here https://github.com/WalletConnect/blockchain-api/pull/713#discussion_r1684529546 Should be the followup, as a lot of places in the code should be changed.

src/handlers/sessions/cosign.rs
```diff
@@ -55,52 +54,37 @@ async fn handler_internal(
...
         })?;
-    let sinature_check = verify_message_signature(
-        &user_op_hash,
-        &request_payload.user_op.signature.clone(),
-        &address,
-        &chain_id_caip2,
+
+    // Get the userOp hash
+    // Entrypoint v07 contract address
+    let contract_address = "0x0000000071727De22E5E9d8BAf0edAc6f37da032"
```
nopestack, line 78: consider making this a constant

src/utils/crypto.rs
geekbrother: Fixed.

src/utils/crypto.rs
geekbrother: Added, thanks!

src/utils/crypto.rs
geekbrother: I think this should be probably extracted to the json and imported at the compile time.

nopestack (approved):

src/utils/crypto.rs
```diff
@@ -95,26 +96,84 @@ struct BundlerSimulationType {
...
+}
+
+fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
+    std::array::from_fn(|i| {
+        if let Some(i) = i.checked_sub(a.len()) {
+            b[i]
+        } else {
+            a[i]
+        }
+    })
```
chris13524, line 163: It took me a while to understand what this was doing; maybe something like this would be more readable:

```suggestion
    let mut result = [0; 32];
    result[..16].copy_from_slice(&a);
    result[16..].copy_from_slice(&b);
    result
```

PR #715, opened 2024-07-22
feat(sessions): updating the cosign signing and the bundler

geekbrother:
# Description

This PR is making the following changes to the Co-Signing endpoint:
* Changing the signature to comply with the Ethereum signing flow, using the `LocalWallet` and packing the signature.
* Changing the bundler function and variables to not be sticked to the Biconomy (`BICONOMY_BUNDLER_*` -> `BUNDLER_*`) and changing the default bundler to the Pimlico.
* Changing the request `userOperation` struct to be `v07` from `v06` to proper calculation of the packed `UserOperation`.
* Adding calculation of the additional fields (`paymaster`, `factory`) for the packed `UserOperation` structure builder to support paymasters.
* Multiple fixes and crypto utils functions refactors.

## How Has This Been Tested?

This flow was tested manually with the testing dapp cosigning flow.

## Due Diligence

* [ ] Breaking change


src/handlers/identity.rs
```diff
@@ -466,8 +466,7 @@ impl JsonRpcClient for SelfProvider {
         let id = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .expect("Time should't go backwards")
-            .as_millis()
-            .to_string();
+            .as_secs();
```
geekbrother, line 469: We have a shared JsonRPC request struct and the bundler complained that the `Id` should be the int.

src/utils/crypto.rs
```diff
@@ -163,11 +205,34 @@ fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
     })
 }
 
-pub fn add_eip191(message: &str) -> String {
-    format!("\x19Ethereum Signed Message:\n{}{}", message.len(), message)
+/// Convert message to EIP-191 compatible format
+pub fn to_eip191_message(message: &[u8]) -> Vec<u8> {
```
geekbrother, line 209: Changing this function to use bytes instead of Strings, to remove double conversions since we are using bytes for the keccak256 hashing.

src/utils/crypto.rs
```diff
@@ -163,11 +205,34 @@ fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
...
+    let prefix = format!("\x19Ethereum Signed Message:\n{}", message.len());
+    let mut eip191_message = Vec::with_capacity(prefix.len() + message.len());
+    eip191_message.extend_from_slice(prefix.as_bytes());
+    eip191_message.extend_from_slice(message);
+    eip191_message
+}
+
+/// Pack signature into a single byte array to Ethereum compatible format
+pub fn pack_signature(unpacked: &EthSignature) -> Bytes {
```
geekbrother, line 218: Packing of `s,r,v` of the ecdsa signature into the Ethereum format.

src/handlers/sessions/cosign.rs
```diff
@@ -142,31 +168,26 @@ async fn handler_internal(
     )
     .await?;
 
-    // Update the userOp with the signature
-    user_op.signature = get_signature_result;
+    // Todo: remove this debug line before production stage
+    info!("UserOpPacked final JSON: {:?}", serde_json::json!(user_op));
```
geekbrother, line 172: We need this userOp json to trace it in case of issues for the alpha version. Since there is no any traffic to this endpoint, there shouldn't be a problem with the Cloudwatch cost.

nopestack (approved):

src/utils/crypto.rs
```diff
@@ -163,11 +205,34 @@ fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
     })
 }
 
-pub fn add_eip191(message: &str) -> String {
-    format!("\x19Ethereum Signed Message:\n{}{}", message.len(), message)
+/// Convert message to EIP-191 compatible format
+pub fn to_eip191_message(message: &[u8]) -> Vec<u8> {
+    let prefix = format!("\x19Ethereum Signed Message:\n{}", message.len());
+    let mut eip191_message = Vec::with_capacity(prefix.len() + message.len());
```
chris13524, line 211: Nice using `with_capacity`

src/utils/crypto.rs
chris13524: Nice refactor

src/utils/crypto.rs
```diff
@@ -163,11 +205,34 @@ fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
...
+    // Extract r, s, and v from the signature
+    let r = unpacked.r;
+    let s = unpacked.s;
+    let v = if unpacked.v == 27 { 0x1b } else { 0x1c };
+    let mut r_bytes = [0u8; 32];
+    let mut s_bytes = [0u8; 32];
+    r.to_big_endian(&mut r_bytes);
+    s.to_big_endian(&mut s_bytes);
+    // Pack r, s, and v into a single byte array
+    let mut packed_signature = Vec::with_capacity(65);
```
chris13524, line 228:
```suggestion
    let mut packed_signature = Vec::with_capacity(r_bytes.len() + s_bytes.len() + 1);
```

chris13524 (approved):
LGTM

We'll need tests here soon!

geekbrother:
> We'll need tests here soon!

Totally! I'm testing by the manual run of the integration test now, since we need a gas to make the full flow. 
I'm thinking to reduce the test to just the validation/simulation step for the CI/CD integration tests.

src/handlers/sessions/cosign.rs
```diff
@@ -105,27 +116,42 @@ async fn handler_internal(
...
         .clone()
         .ok_or_else(|| RpcError::PermissionContextNotUpdated(request_payload.pci.clone()))?;
-    let permission_context = hex::decode(permission_context_item.context.permissions_context)
-        .map_err(|e| RpcError::WrongHexFormat(e.to_string()))?;
+    let permission_context = hex::decode(
+        permission_context_item
+            .context
+            .permissions_context
+            .clone()
+            .trim_start_matches("0x"),
```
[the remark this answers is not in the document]
geekbrother: The client sends this as a String representation of eth hex bytes "0x...", to pass this to the contract we need to convert this representation into `ethers::Bytes` as the contract expects the Bytes format. To do this we need to hex decode it and trim `0x` to decode. That's why we need to trim `0x` when converting it.

src/handlers/identity.rs
geekbrother: Makes sense. We are using this in many places so this should be a follow-up. https://github.com/WalletConnect/blockchain-api/issues/716 was created for this. Thanks for catching this!

PR #717, opened 2024-07-26
fix(sessions): removing getting of the receipt from the co-sign request

geekbrother:
# Description

This PR removes the getting receipt step from the co-signer request handler. Getting the receipt should be moved to a separate endpoint. 
The reason is the receipt can be obtained later than the request for user operation was made, so we need to poll for the receipt after the co-sign request from a dedicated endpoint for that operation.
The co-sign endpoint will now return the user operation tx hash instead of the receipt.

## How Has This Been Tested?

* Manually tested.

## Due Diligence

* [x] Requires a documentation API SPEC update
* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -41,7 +41,7 @@ pub const JSON_RPC_BUNDLER_METHOD_STR: &str = "eth_sendUserOperation";
 pub static JSON_RPC_BUNDLER_METHOD: once_cell::sync::Lazy<Arc<str>> =
     once_cell::sync::Lazy::new(|| Arc::from(JSON_RPC_BUNDLER_METHOD_STR));
 
-pub const JSON_RPC_GET_RECEIPT_METHOD_STR: &str = "eth_getTransactionReceipt";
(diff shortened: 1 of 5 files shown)


src/utils/crypto.rs
```diff
@@ -41,7 +41,7 @@ pub const JSON_RPC_BUNDLER_METHOD_STR: &str = "eth_sendUserOperation";
 pub static JSON_RPC_BUNDLER_METHOD: once_cell::sync::Lazy<Arc<str>> =
     once_cell::sync::Lazy::new(|| Arc::from(JSON_RPC_BUNDLER_METHOD_STR));
 
-pub const JSON_RPC_GET_RECEIPT_METHOD_STR: &str = "eth_getTransactionReceipt";
+pub const JSON_RPC_GET_RECEIPT_METHOD_STR: &str = "eth_getUserOperationReceipt";
```
geekbrother, line 44: We should use `eth_getUserOperationReceipt` for the bundler user op hash instead.

geekbrother:
2f1533589ecbb198ecc978323e27b262664cd85e changes are to make updated clippy happy and not related to this PR.

chris13524 (approved):
LGTM

nopestack (approved):

PR #720, opened 2024-07-31
fix(o11y): ENS Metric tracked without aggregate

arein:
The ENS metric shows one metric per dimension instead of an aggregate.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
Thanks for adding this!

PR #724, opened 2024-08-06
fix(identity): properly handling of the avatar NFT absence

geekbrother:
# Description

This PR adds properly handling exceptional cases where the ENS name avatar pointing to the ERC721 NFT which is absent or the contract address is wrong.
Currently, when ENS name have an `avatar record` that points to the wrong contract address we are returning `HTTP 500` error (because the contract call result is `0x`), which is wrong since the name was resolved and we should just return `avatar: null` in such cases instead of the server error.

## How Has This Been Tested?

* The following identity request triggers `HTTP 500`: `/v1/identity/0x08D3a14b98D1E15B197bb86545d119E1C769Df45`.
* This Pr changes fixing the response to became `{"name":"bubnov.eth", "avatar":null }`.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -348,12 +348,21 @@ pub fn handle_rpc_error(error: ProviderError) -> Result<(), RpcError> {
     match error {
         ProviderError::CustomError(e) if e.starts_with(SELF_PROVIDER_ERROR_PREFIX) => {
             let error_detail = e.trim_start_matches(SELF_PROVIDER_ERROR_PREFIX);
-            // Exceptions for the detailed HTTP error return on RPC call
+            // Exception for no available JSON-RPC providers
             if error_detail.contains("503 Service Unavailable") {
-                Err(RpcError::ProviderError)
-            } else {
-                Err(RpcError::IdentityLookup(error_detail.to_string()))
+                return Err(RpcError::ProviderError);
             }
+            // Proceed with Ok() if the error is related to the contract call error
+            // since there should be a wrong NFT avatar contract address.
+            if error_detail.contains("Contract call error") {
+                warn!(
+                    "Contract call error while looking up identity: {:?}",
+                    error_detail
+                );
+                return Ok(());
+            }
+
+            Err(RpcError::IdentityLookup(error_detail.to_string()))
         }
         ProviderError::CustomError(e) => {
             debug!("Custom error while looking up identity: {:?}", e);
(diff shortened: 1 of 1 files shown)


src/handlers/identity.rs
```diff
@@ -506,21 +518,22 @@ impl JsonRpcClient for SelfProvider {
...
+                // panic when looking for an avatar. This is a workaround for the ethers-rs
+                // when avatar pointing to the wrong ERC-721 contract address.
                 if r.result == EMPTY_RPC_RESPONSE {
-                    return Err(SelfProviderError::ProviderError {
-                        status: StatusCode::METHOD_NOT_ALLOWED,
-                        body: format!("JSON-RPC result is {}", EMPTY_RPC_RESPONSE),
-                    });
+                    return Err(SelfProviderError::ContractCallError(
+                        "Empty response from the contract call".into(),
+                    ));
```
chris13524, line 526: Why are you changing this?

src/handlers/identity.rs
```diff
@@ -506,21 +518,22 @@ impl JsonRpcClient for SelfProvider {
...
                 }
             }
         };
-        let result = serde_json::from_value(result).map_err(|_| {
-            SelfProviderError::GenericParameterError(
-                "Caller always provides generic parameter R=Bytes".into(),
-            )
+        let result = serde_json::from_value(result).map_err(|e| {
+            SelfProviderError::GenericParameterError(format!(
+                "Result should always provide generic parameter which is deserializable: {}",
```
chris13524, line 534: Why are you changing this value? It needs to be `Bytes` not any value

chris13524 (approved):

src/handlers/identity.rs
geekbrother: This change is to distinguish [provider status error](https://github.com/WalletConnect/blockchain-api/blob/d87fc1c5d0765ae1df4a0b001eb75d4184e9b3af/src/handlers/identity.rs#L492) and the contract call error (like in this case) and map this error later to return different result.

src/handlers/identity.rs
geekbrother: The JSON-RPC result which passing here to became a serde_json value [should be DeserializeOwned](https://github.com/WalletConnect/blockchain-api/blob/d87fc1c5d0765ae1df4a0b001eb75d4184e9b3af/src/handlers/identity.rs#L462) not just a `Bytes`. The error message was changed to reflect this.

chris13524: While the type constraint says it must be `DeserializeOwned` the actual value provided must be `Bytes` (which implements `DeserializeOwned`). If it isn't then this `from_value()` call will throw an error (I don't remember exactly what it is).

If it only needed to be `DeserializeOwned` then there wouldn't be an error in this spot and we wouldn't need this error message.

geekbrother: I've checked what the exact result representation is and it's Bytes representation. Changed the error message to the suggested. Thanks!

PR #725, opened 2024-08-08
fix(identity): properly handling of the JSON-RPC code `-32000`

geekbrother:
# Description

This PR adds proper handling of the JSON-RPC code `-32000` when trying to get the identity avatar.
Currently, if the avatar contains an NFT contract address and the execution was reverted there is a JSON-RPC error `-32000` that results in the `HTTP 500` error response. In such cases, we should return just an empty avatar instead of the server error.

## How Has This Been Tested?

* Identity request `/v1/identity/0x2531Dac8784CC16b0F1047C5FB97A4a629c44566` results in the `HTTP 500` error, because the name `shibun.eth` contains the `eip155:1/erc721:0x373075bab7d668ed2473d8233ebdebcf49eb758e/1` avatar contract record that returns execution reverted.
* Calling the same request in the current PR results to the `{"name":"shibun.eth","avatar":null,"resolvedAt":"2024-08-08T13:15:45.889236Z"}` response, which is the same as at the https://ens.app/shibun.eth

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/identity.rs
+++ b/src/handlers/identity.rs
@@ -355,7 +355,7 @@ pub fn handle_rpc_error(error: ProviderError) -> Result<(), RpcError> {
             // Proceed with Ok() if the error is related to the contract call error
             // since there should be a wrong NFT avatar contract address.
             if error_detail.contains("Contract call error") {
-                warn!(
+                debug!(
                     "Contract call error while looking up identity: {:?}",
                     error_detail
                 );
@@ -363,7 +363,7 @@ pub fn handle_rpc_error(error: ProviderError) -> Result<(), RpcError> {
             }
             // Proceed with Ok() if the error is related to the JSON-RPC -32000 error code.
             if error_detail.contains("code: -32000") {
-                warn!(
+                debug!(
                     "JsonRpcError code -32000 while looking up identity: {:?}",
                     error_detail
                 );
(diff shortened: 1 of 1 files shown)


src/handlers/identity.rs
```diff
@@ -361,6 +361,14 @@ pub fn handle_rpc_error(error: ProviderError) -> Result<(), RpcError> {
                 );
                 return Ok(());
             }
+            // Proceed with Ok() if the error is related to the JSON-RPC -32000 error code.
+            if error_detail.contains("code: -32000") {
+                warn!(
```
chris13524, line 366: I don't think we should warn here unless the issue is our fault

geekbrother: We are using the default `Error` log level at the production, so this will not flood our logs. 
But that's a good point for debugging, why the avatar is null despite having the `avatar` record. 
Otherwise, there are no tracing/debug errors in this case because we are returning `Ok`.

chris13524: How I think of the log levels:
- Fatal/error -> something is very wrong and UX is degraded, we should get alerts for this and look into them.
- Warn -> something is wrong we weren't expecting, but UX not degraded. We should get alerts (we don't currently, which is fine). We should look into these issues.
- Info/debug/trace -> various levels of detail on stuff going on, to help debug

Fatal/error/warn means we should look into what's happening and resolve them. Warn means it's not necessarily urgent since it doesn't affect UX.

Sounds like we've identified an error, and are blaming it on the user, so it's not our fault, so we don't need to log it as a warning.

chris13524 (approved):

src/handlers/identity.rs
geekbrother: It should be `debug` instead.

PR #732, opened 2024-08-15
fix: remove special TCP flags

chris13524:
# Description

Removes special TCP flags... these seem to be leading to TCP resets.

The ALB is responsible for closing client connections, not these flags, so we need to tune on the ALB instead. Looks like we customized these values in the load balancer to 60 and 60 already, but didn't put in Terraform.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
I'm not sure that this will fix 502s and connection resets, but fingers crossed 🤞 and let's check this out.

PR #738, opened 2024-08-20
fix: min 1 weight

### LINKED ISSUE #737: feat(providers): make a test call to provider with zero weight.
We should make a test call(s) to the provider which is failed before resuming the traffic. 
Currently, when the provider weight decreases to zero we are not forwarding any traffic and reset the weight after 3 hours and forwarding the traffic even if the provider is not yet up.

chris13524:
# Description

Sets the minimum weight to 1 when selecting a provider so that there's always a little bit of traffic to a provider.

Resolves #737 

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -226,7 +226,11 @@ impl ProviderRepository {
             return Err(RpcError::UnsupportedChain(chain_id.to_string()));
         }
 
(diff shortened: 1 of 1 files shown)


geekbrother (approved):
LGTM 👍

PR #739, opened 2024-08-20
feat(providers): implementing SolScan provider for the solana address balance

geekbrother:
# Description

This PR adds the SolScan as a Solana address balance provider. The provider uses the [v1 Account Tokens API](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/account-tokens) and `chainId` query parameter to determine which namespace should be used.

## How Has This Been Tested?

* New integration tests were added.

## Merging policy 🚧

* [x] The API token key must be updated to the production one.

## Due Diligence

* [ ] Breaking change
* [x] Requires a documentation update


src/providers/solscan.rs
```diff
@@ -0,0 +1,97 @@
...
+            .into_iter()
+            .map(|f| BalanceItem {
+                name: f
+                    .token_name
+                    .unwrap_or_else(|| f.token_symbol.clone().unwrap_or_default()),
+                symbol: f.token_symbol.unwrap_or_default(),
+                chain_id: None,
+                address: Some(f.token_address),
+                value: None,
+                price: 0.0,
```
geekbrother, line 82: We need to fetch current tokens pricing from the different endpoint. Doing this in the same request for each token would be excessive so there will be a follow-up PR where we will fetch the current tokens' pricing once per `N` second and use the cached data for this calculation.

src/providers/solscan.rs
```diff
@@ -0,0 +1,97 @@
...
+        address: String,
+        _params: BalanceQueryParams,
+        http_client: reqwest::Client,
+    ) -> RpcResult<BalanceResponseBody> {
+        let mut url = Url::parse(ACCOUNT_TOKENS_URL).map_err(|_| RpcError::BalanceParseURLError)?;
+        url.query_pairs_mut().append_pair("account", &address);
+
+        let response = http_client
+            .get(url)
+            .header("Content-Type", "application/json")
```
chris13524, line 58: Since this is a GET request isn't this header useless?

src/providers/solscan.rs
chris13524: Makes sense. I'd put a TODO in the code for that then

src/utils/crypto.rs
```diff
@@ -32,8 +32,13 @@ const ENSIP11_MAINNET_COIN_TYPE: u32 = 60;
 static CAIP_CHAIN_ID_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(r"[-a-zA-Z0-9]{1,32}").expect("Failed to initialize regexp for the chain ID format")
 });
-static CAIP_ADDRESS_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"[-a-zA-Z0-9]{1,63}").expect("Failed to initialize regexp for the address format")
+static CAIP_ETH_ADDRESS_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"0x[a-fA-F0-9]{40}")
```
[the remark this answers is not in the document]
chris13524: This doesn't seem right, I think your version was correct: https://walletconnect.slack.com/archives/C03T29N5DSL/p1724346824677309

chris13524 (approved):
LGTM!

src/utils/crypto.rs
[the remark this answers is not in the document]
geekbrother: Thanks for raising this 🙌 I was sure it's 40 and didn't check the CAIP spec. Now we are know 😅

geekbrother:
Removing `namespace` argument in favor of using `chainId`. The [full Slack context](https://walletconnect.slack.com/archives/C03SCF66K2T/p1724246068730469?thread_ts=1722872189.697099&cid=C03SCF66K2T).

src/providers/solscan.rs
geekbrother: Makes sense. Thanks for catching this.

PR #741, opened 2024-08-22
fix(rpc): passing through a `bad request` error from the RPC provider

### LINKED ISSUE #740: feat(rpc): proper handling of the provider unsuccessful responses
The proper handling and proper response to the user should be implemented in case of an unsuccessful provider error.

Currently, we are trying to call the next provider if the provider's RPC call responds with a not-success error code. 
This means that users with the wrong JSON-RPC request will get `HTTP 503` and `chain x is temporarily unavailable` instead of the proper error, like `HTTP 400`.

geekbrother:
# Description

This PR adds proper handling of the `HTTP 400 Bad Request` RPC provider response.
Currently, if the user sends a bad JSON-RPC request, the provider responds with `HTTP 400` which is considered a provider error, and we try the next provider in the retry list. Every provider fails and that results in `HTTP 503 Provider x unavailable` error instead of stopping retrying and passing the `HTTP 400 Bad request` error to the user with the error details.

Resolves #740

## How Has This Been Tested?

* A new integration test was created.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -109,12 +109,7 @@ pub async fn rpc_call(
         .await;
 
         match response {
-            // Passing the response through to the user if it's a bad request
-            // to preserve the JSON-RPC error message
-            Ok(response)
-                if response.status() == http::StatusCode::BAD_REQUEST
-                    || !response.status().is_server_error() =>
-            {
+            Ok(response) if !response.status().is_server_error() => {
                 return Ok(response);
             }
             e => {
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -109,7 +109,12 @@ pub async fn rpc_call(
         .await;
 
         match response {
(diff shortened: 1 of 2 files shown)


src/handlers/proxy.rs
```diff
@@ -109,7 +109,12 @@ pub async fn rpc_call(
         .await;
 
         match response {
-            Ok(response) if !response.status().is_server_error() => {
+            // Passing the response through to the user if it's a bad request
+            // to preserve the JSON-RPC error message
+            Ok(response)
+                if response.status() == http::StatusCode::BAD_REQUEST
+                    || !response.status().is_server_error() =>
+            {
```
chris13524, line 117: I don't think we need this change. A bad request is not a server error already

geekbrother: Good catch. Removed it. Thanks a lot!

PR #742, opened 2024-08-23
feat(transactions): implementing Solana transactions history support

geekbrother:
# Description

This PR adds the SolScan as a Solana address transactions history provider. 
The provider uses the [v2 Account Transfers API](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/v2-account-transfer) and `chainId` query parameter to determine which namespace should be used.
We are also adding the SolScan V2 API key to support both V1 (for the address balance API) and V2 for this transactions history API.

## How Has This Been Tested?

* New integration tests were implemented.

## Due Diligence

* [ ] Breaking change


src/providers/solscan.rs
```diff
@@ -94,3 +154,85 @@ impl BalanceProvider for SolScanProvider {
...
+                    mined_at: f.time,
+                    nonce: 0,
+                    sent_from: f.from_address,
+                    sent_to: f.to_address,
+                    status: "confirmed".to_string(), // Balance changes are always confirmed
+                    application: None,
+                    chain: Some(SOLANA_MAINNET_CHAIN_ID.to_string()),
+                },
+                transfers: Some(vec![HistoryTransactionTransfer {
+                    fungible_info: None, // Todo: Add fungible info from saved tokens info list
```
geekbrother, line 220: Same as [here in Solana balance support PR](https://github.com/WalletConnect/blockchain-api/pull/739#discussion_r1724012278):

> We need to fetch current tokens pricing from the different endpoint. Doing this in the same request for each token would be excessive so there will be a follow-up PR where we will fetch the current tokens' pricing once per N second and use the cached data for this calculation.

This will be a follow-up PR.

nopestack (approved):
All good to me it seems. @chris13524 please take a look as well

src/providers/solscan.rs
```diff
@@ -94,3 +154,85 @@ impl BalanceProvider for SolScanProvider {
...
+    ) -> RpcResult<HistoryResponseBody> {
+        let mut url =
+            Url::parse(ACCOUNT_HISTORY_URL).map_err(|_| RpcError::BalanceParseURLError)?;
+        url.query_pairs_mut().append_pair("page_size", "100");
+        url.query_pairs_mut().append_pair("remove_spam", "true");
+        url.query_pairs_mut()
+            .append_pair("exclude_amount_zero", "true");
+        url.query_pairs_mut().append_pair("address", &address);
+        let page = params.cursor.unwrap_or("1".into());
+        url.query_pairs_mut().append_pair("page", &page);
```
chris13524, line 175: Good to use this `Url` builder instead of `format!()`

`append_pair()` automatically handles URL encoding so there's no risk of this string `page` parameter injecting more flags.

chris13524 (approved):
LGTM!

PR #747, opened 2024-08-26
fix(tests): fixing transactions history integration tests

geekbrother:
# Description

This PR fixes transaction history integration tests by making the following changes:
* Updating the address format checking function;
* Fixing the comparison in the tests;
* Removing the `next` cursor in case of empty data.

## How Has This Been Tested?

* Tested by running the integration test locally and passing tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -6,6 +6,7 @@ use {
     },
     alloy_primitives::Address,
     base64::prelude::*,
+    bs58,
     ethers::{
         abi::Token,
         core::{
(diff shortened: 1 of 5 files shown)


src/utils/crypto.rs
```diff
@@ -548,8 +549,21 @@ pub fn is_coin_type_supported(coin_type: u32) -> bool {
...
+            if !CAIP_ETH_ADDRESS_REGEX.is_match(address) {
+                return false;
+            }
+            H160::from_str(address).is_ok()
+        }
+        CaipNamespaces::Solana => {
+            if !CAIP_SOLANA_ADDRESS_REGEX.is_match(address) {
+                return false;
+            }
+            match bs58::decode(address).into_vec() {
```
geekbrother, line 562: It's better not to check just regexp but to try to decode the address and not pass it to the provider in case it's wrong.

chris13524 (approved):

PR #765, opened 2024-09-05
feat(providers): adding Solana fungible price support

geekbrother:
# Description

This PR adds Solana support for the [fungibles pricing endpoint](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-api#price) by using the [SolScan Token price](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/v2-token-price) API endpoint for the SOL native token and [SolScan Token Metadata](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/v2-token-meta) endpoint for the SPL tokens.

## How Has This Been Tested?

* A new [integration test was implemented](https://github.com/WalletConnect/blockchain-api/pull/765/files#diff-4493ff1a59d5bf63fc4e0b226e9908932f9c2ef71dafddf6ecf99f00c5739487).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/solscan.rs
+++ b/src/providers/solscan.rs
@@ -1,15 +1,20 @@
 use {
-    super::{BalanceProvider, HistoryProvider},
+    super::{
+        BalanceProvider, FungiblePriceProvider, HistoryProvider, PriceResponseBody,
+        SupportedCurrencies,
+    },
     crate::{
         error::{RpcError, RpcResult},
         handlers::{
             balance::{BalanceItem, BalanceQuantity, BalanceQueryParams, BalanceResponseBody},
+            fungible_price::FungiblePriceItem,
             history::{
                 HistoryQueryParams, HistoryResponseBody, HistoryTransaction,
                 HistoryTransactionMetadata, HistoryTransactionTransfer,
                 HistoryTransactionTransferQuantity,
             },
         },
+        utils::crypto::SOLANA_NATIVE_TOKEN_ADDRESS,
     },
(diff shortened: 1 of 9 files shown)


geekbrother:
As the further optimization caching of the `get_token_info` method response will be implemented as a follow-up to save on SolScan provider calls since we can cache the tokens metadata for 24 hours (I don't think metadata changes often) and for 1-2 minutes for the price calls.

src/storage/redis/mod.rs
```diff
@@ -89,6 +89,7 @@ impl Redis {
         })
     }
 
+    #[allow(dependency_on_unit_never_type_fallback)]
```
geekbrother, line 92: This is not related to this PR, it is temporary fix to make clippy happy (on the latest Rust updates) and should be fixed in https://github.com/WalletConnect/blockchain-api/issues/766

src/providers/solscan.rs
```diff
@@ -18,22 +23,131 @@ use {
     url::Url,
 };
 
+const SOLANA_SOL_TOKEN_ICON: &str =
+    "https://cdn.jsdelivr.net/gh/trustwallet/assets@master/blockchains/solana/info/logo.png";
```
geekbrother, line 27: Grabbed from the [Solana labs tokens list](https://github.com/solana-labs/token-list/blob/main/src/tokens/solana.tokenlist.json).

chris13524 (approved):
LGTM

PR #769, opened 2024-09-06
fix(solscan): fixing tokens amount in the transactions history

geekbrother:
# Description

This PR fixes the token amount in the transaction history for Solana.
Currently, the token amount is returned without applying decimals, this PR fixes the issue by applying `amount / 10f64.powf(token_decimals)`.

## How Has This Been Tested?

* Current integration tests and manual testing.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/solscan.rs
+++ b/src/providers/solscan.rs
@@ -200,6 +200,7 @@ struct HistoryResponseItem {
     pub from_address: String,
     pub to_address: String,
     pub token_address: String,
+    pub token_decimals: u8,
     pub amount: usize,
     pub flow: HistoryDirectionType,
     pub time: String,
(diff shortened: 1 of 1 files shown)


chris13524 (approved):

chris13524 (commented):
Actually won't this result in floating point errors? Perhaps we should use string building instead? I remember you did something similar in another PR

geekbrother:
Thanks for the comment!

> Actually won't this result in floating point errors?

I think it shouldn't: `item.amount as f64 / 10f64.powf(token_info.decimals as f64)` amount is `usize` and decimals is `u8`, so we are converting unsigned integers to float and divide them. The only thing here is dividing by zero if decimals is zero.

> I remember you did something similar in another PR

It's the same expression but moved to the variable since it's used twice now.

PR #770, opened 2024-09-07
feat(providers): respond HTTP 503 instead of 500 on provider non-success request

geekbrother:
# Description

This PR makes the following change for the Zerion and SolScan provider's error handling: in case the request to the provider was not a successful `!response.status().is_success()` we will respond `HTTP 503` Provider temporary unavailable, instead of responding with HTTP 500 Server error. This makes better sense in case if the provider responds with the 500 or 503s.

## How Has This Been Tested?

* Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -500,6 +500,38 @@ impl IntoResponse for RpcError {
                 )),
             )
                 .into_response(),
+            Self::TransactionProviderError => (
+                StatusCode::SERVICE_UNAVAILABLE,
+                Json(new_error_response(
+                    "".to_string(),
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -500,6 +500,38 @@ impl IntoResponse for RpcError {
                 )),
             )
                 .into_response(),
+            Self::TransactionProviderError => (
```
geekbrother, line 503: We are getting these types of errors if the request to the Zerion or SolScan was not 2xx.

nopestack (approved):

chris13524 (approved):

PR #771, opened 2024-09-09
feat(names): refactoring and extending support for name registration zones

geekbrother:
# Description

This PR makes the following changes to extend name registrations for multiple zones:

* Refactoring by moving all names-related utils and constants into [the `names` subdirectory](https://github.com/WalletConnect/blockchain-api/pull/771/files#diff-f0a1480d98d765fa191604d2ba06bf205b52c1786cb5692bfdfd58fbea99f229) instead of placing in handlers.
* Removing hardcoded single name zone in favor of the `names_allowed_zones` TFC variable and `RPC_PROXY_NAMES_ALLOWED_ZONES` environment variable which is a comma-separated list of supported name zones. This will allow to use of more than one name zone.
* Adding optional `zone` query parameter for name suggestions endpoint according to the [API spec proposal](https://github.com/WalletConnect/walletconnect-specs/pull/255).

## How Has This Been Tested?

* Current integration tests were updated.

## Due Diligence

* [ ] Breaking change


src/handlers/profile/mod.rs
```diff
@@ -13,8 +13,10 @@ pub mod reverse;
 pub mod suggestions;
 pub mod utils;
 
-/// List of allowed name zones
-pub const ALLOWED_ZONES: [&str; 1] = ["wcn.id"];
+/// List of allowed name zones for lookup
+pub const ALLOWED_ZONES: [&str; 2] = ["reown.id", "wcn.id"];
+/// Main zone for new names registration and suggestions
+pub const MAIN_ZONE: &str = ALLOWED_ZONES[0];
```
chris13524, line 19: Suggest this to improve explicitness of using reown.id as main zone:

```suggestion
const REOWN_ID: &str = "reown.id";
const WCN_ID: &str = "wcn.id";
/// List of allowed name zones for lookup
pub const ALLOWED_ZONES: [&str; 2] = [REOWN_ID, WCN_ID];
/// Main zone for new names registration and suggestions
pub const MAIN_ZONE: &str = REOWN_ID;
```

chris13524 (approved):

src/handlers/profile/mod.rs
geekbrother: Makes sense! Thanks for the comment.

chris13524 (approved):

PR #772, opened 2024-09-12
feat(monitoring): separating the alert for transactions list handler

geekbrother:
# Description

This PR separates the handler's high execution time alert for all handlers and for the transactions history endpoint because Solana history takes longer and should have a higher alerting threshold.

## How Has This Been Tested?

* Only manual testing.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/monitoring/panels/app/handlers_latency.libsonnet
+++ b/terraform/monitoring/panels/app/handlers_latency.libsonnet
@@ -32,19 +32,36 @@ local _configuration = defaults.configuration.timeseries
         alertCondition.new(
           evaluatorParams = [ 3000 ],
           evaluatorType   = 'gt',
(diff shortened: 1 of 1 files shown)


terraform/monitoring/panels/app/handlers_latency.libsonnet
```diff
@@ -32,19 +32,36 @@ local _configuration = defaults.configuration.timeseries
         alertCondition.new(
           evaluatorParams = [ 3000 ],
           evaluatorType   = 'gt',
-          operatorType    = 'or',
```
geekbrother, line 35: We don't need an operator here since we have just one evaluator param.

chris13524 (approved):

PR #773, opened 2024-09-13
feat(monitoring): adding proper metrics for all external requests to non-proxy providers

geekbrother:
# Description

This PR adds proper monitoring metrics for all external requests to non-proxy providers.
At the moment we are just monitoring latency and status codes for the Proxy providers and only monitoring handlers execution latency for non-proxy providers and don't monitor response codes from non-proxy providers at all. Some handlers use different non-proxy providers and execute not only one request to the provider. We need to implement proper monitoring for the non-proxy provider's endpoints.

This PR adds the following changes:
* Refactoring: aggregating repeated code parts into private methods,
* Adding request latency and response status codes by invoking the [add_latency_and_status_code](https://github.com/WalletConnect/blockchain-api/pull/773/files#diff-9f90e93b5696724fd8e4407c9db0bfd381d5468bcfbf2f09c17ac8510a85e104R414) metrics function for each of the following provider's external calls:
  * Zerion,
  * OneInch,
  * SolScan.

By adding the metrics for each request we can monitor the health of each of the non-proxy provider's endpoints and make an alert if it's not available or has a high latency.

Grafana panels for these providers was added in the follow-up https://github.com/WalletConnect/blockchain-api/pull/774

## How Has This Been Tested?

* Current integration tests.

## Due Diligence

* [x] Add Grafana panels for each non-proxy provider
* [ ] Breaking change

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -5,7 +5,6 @@ use {
         providers::{ProviderKind, RpcProvider},
         storage::irn::OperationType,
     },
-    hyper::http,
     sqlx::PgPool,
     std::time::{Duration, SystemTime},
     sysinfo::{
@@ -334,10 +333,13 @@ impl Metrics {
         )
     }
 
-    pub fn add_external_http_latency(&self, provider_kind: ProviderKind, latency: f64) {
+    pub fn add_external_http_latency(&self, provider_kind: ProviderKind, start: SystemTime) {
         self.http_external_latency_tracker.record(
             &otel::Context::new(),
-            latency,
+            start
+                .elapsed()
+                .unwrap_or(Duration::from_secs(0))
+                .as_secs_f64(),
             &[otel::KeyValue::new("provider", provider_kind.to_string())],
         )
     }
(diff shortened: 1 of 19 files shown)


src/metrics.rs
```diff
@@ -334,10 +333,13 @@ impl Metrics {
         )
     }
 
-    pub fn add_external_http_latency(&self, provider_kind: ProviderKind, latency: f64) {
+    pub fn add_external_http_latency(&self, provider_kind: ProviderKind, start: SystemTime) {
```
chris13524, line 336: I think `Instant` is better than `SystemTime` for measuring delays. Since it isn't affected by time going backwards.

src/providers/zerion.rs
```diff
@@ -14,34 +14,45 @@ use {
...
+            provider_kind: ProviderKind::Zerion,
             api_key,
             http_client,
         }
     }
+
+    async fn send_request(&self, url: Url) -> Result<reqwest::Response, reqwest::Error> {
+        self.http_client
+            .get(url)
+            .header("authorization", format!("Basic {}", self.api_key))
```
chris13524, line 52: I think there's a `.basic_authentication()` or `.authentication()` method here

chris13524 (approved):
Awesome work, thank you!

nopestack (approved):

PR #775, opened 2024-09-16
fix(OneInch): proper handling of the bad request responses

geekbrother:
# Description

This PR adds proper handling of the HTTP 400 Bad Request response from OneInch provider.
* Updating the error object schema to the up-to-date to extract the error description and pass it through along with the HTTP 400 response,
* Responding with the HTTP 400 Bad Request even if the OneInch error schema was changed (currently we are responding with the HTTP 500 which is confusing) to reflect that this is a bad argument related issue and not the server error.

## How Has This Been Tested?

* Current integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/one_inch.rs
+++ b/src/providers/one_inch.rs
@@ -88,18 +88,27 @@ impl OneInchProvider {
         );
 
         if !price_response.status().is_success() {
+            // Passing through error description for the error context
+            // if user parameter is invalid (got 400 status code from the provider)
+            if price_response.status() == reqwest::StatusCode::BAD_REQUEST {
+                let response_error = match price_response.json::<OneInchErrorResponse>().await {
+                    Ok(response_error) => response_error.description,
+                    Err(e) => {
+                        error!(
+                            "Error parsing OneInch HTTP 400 Bad Request error response {:?}",
+                            e
+                        );
+                        // Respond to the client with a generic error message and HTTP 400 anyway
+                        "Invalid parameter".to_string()
+                    }
+                };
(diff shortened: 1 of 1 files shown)


src/providers/one_inch.rs
```diff
@@ -88,18 +88,27 @@ impl OneInchProvider {
...
+                            e
+                        );
+                        // Respond to the client with a generic error message and HTTP 400 anyway
+                        "Invalid parameter".to_string()
+                    }
+                };
+                return Err(RpcError::ConversionInvalidParameter(response_error));
             }
+
+            error!(
```
geekbrother, line 108: Logging error only when it's not HTTP 400 bad argument response.

src/providers/one_inch.rs
```diff
@@ -88,18 +88,27 @@ impl OneInchProvider {
...
             if price_response.status() == reqwest::StatusCode::BAD_REQUEST {
-                let response_error = price_response.json::<OneInchErrorResponse>().await?;
-                return Err(RpcError::ConversionInvalidParameter(
-                    response_error.error.description,
-                ));
+                let response_error = match price_response.json::<OneInchErrorResponse>().await {
+                    Ok(response_error) => response_error.description,
+                    Err(e) => {
+                        error!(
+                            "Error parsing OneInch HTTP 400 Bad Request error response {:?}",
```
geekbrother, line 98: Logging error if the error schema was changed but responding with the HTTP 400 anyway.

nopestack (approved):

PR #778, opened 2024-09-17
fix(SolScan): using the correct SPL tokens amount in balance

geekbrother:
# Description

This PR fixes SPL tokens balance amounts by using the amount with decimals applied.
Currently, the amount without applied decimals is returned which caused the wrong amount in the UI.

## How Has This Been Tested?

* Current integration tests and manual querying of the Solana address with SPL tokens.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/solscan.rs
+++ b/src/providers/solscan.rs
@@ -220,7 +220,7 @@ struct TokensResponseItem {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 struct AmountItem {
-    pub amount: String,
+    pub ui_amount_string: String,
     pub decimals: u8,
 }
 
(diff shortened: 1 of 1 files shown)


src/providers/solscan.rs
```diff
@@ -220,7 +220,7 @@ struct TokensResponseItem {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 struct AmountItem {
-    pub amount: String,
+    pub ui_amount_string: String,
```
geekbrother, line 223: The `uiAmountString` is the field with applied decimals from the [SolScan response schema](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/account-tokens).

chris13524 (approved):

PR #779, opened 2024-09-17
feat(CoSign): response depends on the `version` query parameter

geekbrother:
# Description

This PR adds the response to the CoSign request depending on the `version` query parameter.
The default response is used when no `version` or `version = 0` is provided for backward compatibility. The updated response `version != 0` should be a signature without calling the wallet service.
We are still in the testing/mvp stage so the `version` will reflect the further stages if needed until the stable.

## How Has This Been Tested?

* Only manual testing.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -526,6 +526,7 @@ pub enum ChainId {
         serialize = "binance_smart_chain",
         serialize = "bsc"
     )]
+    BaseSepoliaTestnet = 84532,
     BinanceSmartChain = 56,
     Blast = 81032,
     Celo = 42220,
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -521,12 +521,13 @@ pub enum ChainId {
     Aurora = 1313161554,
     Avalanche = 43114,
     Base = 8453,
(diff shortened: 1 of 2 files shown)


src/utils/crypto.rs
```diff
@@ -526,6 +526,7 @@ pub enum ChainId {
         serialize = "binance_smart_chain",
         serialize = "bsc"
     )]
+    BaseSepoliaTestnet = 84532,
```
geekbrother, line 529: Adding Base Sepolia ChainID to the supported list.

src/utils/crypto.rs
```diff
@@ -521,6 +521,8 @@ pub enum ChainId {
     Aurora = 1313161554,
     Avalanche = 43114,
     Base = 8453,
+    #[strum(serialize = "base_sepolia_testnet", serialize = "base-sepolia-testnet")]
+    BaseSepoliaTestnet = 84532,
```
geekbrother, line 525: Adding Base Sepolia Testnet to the supported chains for the CoSign.

chris13524 (approved):

PR #782, opened 2024-09-19
feat(SolScan): implementing tokens metadata and price caching

geekbrother:
# Description

This PR implements the local caching for tokens metadata and price.
We are requesting tokens metadata and price on each balance and transaction history request. Since the token's metadata doesn't change often we can cache it in Redis and reuse the cached data instead of querying the provider every time. 
Also, the token's price can be cached to reuse in balance, history, and fungible price endpoints without querying the provider on every request.
This will significantly decrease the latency of our balance, history, and fungible price endpoints and decrease provider request count.

The different TTL for the metadata and pricing cache was applied, since metadata changes are rare and token price changes more often:
* Metadata cache TTL: 1 hour,
* Token price cache TTL: 5 minutes.

The cache response latency metric is also added.

## How Has This Been Tested?

* These changes are covered by the current integration tests for Solana.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -37,7 +37,7 @@ use {
         hash::Hash,
         sync::Arc,
     },
-    tracing::{debug, log::warn},
+    tracing::{debug, error, log::warn},
     wc::metrics::TaskMetrics,
 };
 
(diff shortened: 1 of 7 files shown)


src/providers/mod.rs
```diff
@@ -605,7 +630,7 @@ pub trait RateLimited {
 }
 
 #[async_trait]
-pub trait HistoryProvider: Send + Sync + Debug {
```
geekbrother, line 608: Removing `Debug` from requirements since it's not required and not implemented in the `deadpool_redis::Pool`.

chris13524 (approved):
LGTM

PR #789, opened 2024-09-23
feat: partial prepareCalls impl

chris13524:
# Description

Functional implementation of prepareCalls according to [these specs](https://www.notion.so/walletconnect/API-specs-for-Wallet-Services-0c4b70a397e2435691f9bb1f20bc652d?pvs=4). Just duping this all in this repo for now to get this working, we can refactor into yttrium in another PR. There are a few details remaining to be implemented, but these will also be in a follow-up PR:
- Validation of isSafe and other things, matching the previous JS code (TODOs in code)
- More specific error types (TODOs in code)

Also updates copyright.

Depends on https://github.com/reown-com/yttrium/pull/28

Depends on https://walletconnect.slack.com/archives/C03SCF66K2T/p1727098325949169

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/prepare_calls.rs
+++ b/src/handlers/wallet/prepare_calls.rs
@@ -2,8 +2,10 @@ use crate::analytics::MessageSource;
 use crate::{handlers::HANDLER_TASK_METRICS, state::AppState};
 use alloy::network::Ethereum;
 use alloy::primitives::aliases::U192;
-use alloy::primitives::{address, Address, Bytes, U256, U64};
+use alloy::primitives::{address, bytes, Address, Bytes, FixedBytes, U256, U64};
 use alloy::providers::ReqwestProvider;
+use alloy::sol_types::SolCall;
+use alloy::sol_types::SolValue;
 use axum::{
     extract::State,
     response::{IntoResponse, Response},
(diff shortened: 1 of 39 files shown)


src/handlers/wallet/prepare_calls.rs
```diff
@@ -0,0 +1,532 @@
...
+            request.from,
+            &entry_point_config.address(),
+            key_from_validator_address(validator_address),
+        )
+        .await
+        .map_err(|e| PrepareCallsError::InternalError(PrepareCallsInternalError::GetNonce(e)))?;
+
+        let pimlico_client = BundlerClient::new(BundlerConfig::new(
+            format!(
+                "https://rpc.walletconnect.com/v1/bundler?chainId={}&projectId={}&bundler=pimlico",
```
geekbrother, line 333: We can use [bundler_rpc_call](https://github.com/WalletConnect/blockchain-api/blob/8be3ca5b08dec2387ee2c2ffcb4b7ca739443bcb/src/handlers/bundler.rs#L62) instead of the direct URL request.

geekbrother (approved):

src/handlers/wallet/prepare_calls.rs
chris13524: I would like to use this; but I need to refactor the library to support this sort of arbitrary function call vs needing an HTTP request.

PR #795, opened 2024-09-27
feat(ca_orchestration): implementing chain agnostic orchestration check endpoint

geekbrother:
# Description

This PR implements the chain agnostic bridging check endpoint according to the [research draft doc](https://www.notion.so/walletconnect/CA-Orchestrator-API-Spec-542a6ff37f3c47dc92e8a65b8ed3208c?pvs=4#10b3a661771e80c694f0e4dd1bb623d0).

The endpoint processing the transaction object and making the following checks to return is the bridging required for the transaction:

* Checking the native token balance of the address and compare it to the transaction value. If the native token value >= transaction amount, then return `false` for the bridging.
* Checking the transaction `data` by decoding it to the ERC20 contract ABI and check is the function call is `transfer`. If it's not, return `false` for the bridging.
* Checking the address balances for each of the supported bridging assets and if the balance is more or equal then transaction value returns `true` for the bridging.
* Returning `false` if no bridging assets was found.

What's not included:

* Partners ABIs are out of scope in this PR and should be a follow-up.

## How Has This Been Tested?

* Manually at the moment, integration tests TBD.

## Due Diligence

* [ ] Breaking change


src/handlers/chain_agnostic/check.rs
```diff
@@ -0,0 +1,137 @@
...
+    let native_token_balance = get_erc20_balance(
+        &request_payload.transaction.chain_id,
+        H160::repeat_byte(0xee),
+        from_address,
+        &query_params.project_id.clone(),
+        MessageSource::ChainAgnosticCheck,
+    )
+    .await?;
+    let transfer_value_stripped = request_payload.transaction.value.trim_start_matches("0x");
+    let transfer_value = U256::from_dec_str(transfer_value_stripped)
```
chris13524, line 86: Should this be from_hex_str not decimal? Didn't look at docs though

src/handlers/chain_agnostic/check.rs
```diff
@@ -0,0 +1,137 @@
...
+    if decode_erc20_call_function_data(&transaction_data)? != Erc20FunctionType::Transfer {
+        error!("The transaction data is not a transfer function");
+        return Ok(Json(RequiresMultiChainResponse {
+            requires_multi_chain: false,
+        })
+        .into_response());
+    }
+
+    // Check the ERC20 tokens balance for each of supported assets
+    for (asset, chains) in BRIDGING_AVAILABLE_ASSETS.entries() {
```
chris13524, line 109: This is a LOT of looping. This will take a long time as well as exhaust a lot of API calls.

src/utils/crypto.rs
```diff
@@ -202,6 +205,49 @@ fn concat_128(a: [u8; 16], b: [u8; 16]) -> [u8; 32] {
...
+    Allowance,
+}
+
+/// Decodes ERC20 contract call function data and returns the function name.
+pub fn decode_erc20_call_function_data(
+    function_data: &[u8],
+) -> Result<Erc20FunctionType, CryptoUitlsError> {
+    // Get the 4 bytes function selector
+    let selector: [u8; 4] = function_data[0..4].try_into().map_err(|_| {
+        CryptoUitlsError::Erc20DecodeError("Function data is less then 4 bytes.".into())
```
chris13524, line 232: Should this error be caught by the handler function? If they aren't using ERC-20 transfer should it still require multi chain bridging?

chris13524 (approved):

src/handlers/chain_agnostic/check.rs
lukaisailovic: +1 on this. Now we have 3 chains and 1 asset for the demo purposes, however this is very much going to increase in the future in both ways

geekbrother: We can use our balance API (which is Zerion API actually) to get the available balance without JSON-RPC calls for each contract, but the issue is that the address balance in this case is not the latest (Zerion cache is about 2 minutes) and we are using the JSON-RPC call as well to fetch the latest balance for example for swaps. Not sure how this will impact the UX.

lukaisailovic: @geekbrother Balance call is extremely expensive. We should look into minimizing the amount of those.

src/utils/crypto.rs
geekbrother: That's another case: the function selector in the ERC20 contract is 4 bytes, so if the function data is less, that means that's not the ERC20 contract. The function type check is in the [handler here](https://github.com/reown-com/blockchain-api/pull/795/files#diff-b1dec76ab5ce6ca9fb9cec11fe044bd0adf93b2714ddb030c59dfced4ae4292bR100) and we are returning `false` for the bridging if that's not a transfer. I don't think we should check the balance if that's not a `transfer` cc @lukaisailovic 
The [web example also have this check](https://github.com/reown-com/web-examples/blob/9e94417090cfdd7e4bdc44ee181865530379fc23/advanced/wallets/react-wallet-v2/src/views/SessionSendTransactionModal.tsx#L51) btw.

lukaisailovic: @geekbrother The web code works only for ERC20 transfers and nothing else. Its meant to be used as a showcase

PR #801, opened 2024-10-03
feat(providers): adding `Scroll` chain support

geekbrother:
# Description

This PR adds support for the Scroll Mainnet `eip155: 534352 ` and Scroll Sepolia testnet `eip155: 534351` chains to the RPC.

## How Has This Been Tested?

* New integration tests.

## Due Diligence

* [ ] Breaking change


nopestack (approved):
Consider making these strings enum values instead, as some are repeated and prone to be mistyped when modifying this code.

PR #804, opened 2024-10-07
feat: wallet_getCallsStatus

chris13524:
# Description

Adds support for `wallet_getCallsStatus` to Wallet Service according to [the spec](https://www.notion.so/walletconnect/API-specs-for-Wallet-Services-0c4b70a397e2435691f9bb1f20bc652d?pvs=4#5c4f865af7b24308920b13bae94da5ef).

This adds a basic happy-path test case to test this method. It mocks the call ID input with a "normal" Safe execution which we already have code for. However, going forward we will have code that will be able to test the other 2 `wallet_*` endpoints and we will be able to update this test case to be more E2E.

Depends on https://github.com/reown-com/yttrium/pull/37

## How Has This Been Tested?

New test case

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/get_calls_status.rs
+++ b/src/handlers/wallet/get_calls_status.rs
@@ -315,6 +315,11 @@ mod tests {
             std::env::var("RPC_PROXY_POSTGRES_URI").unwrap()
         );
         let faucet = LocalSigner::random();
+        std::env::vars().for_each(|(k, v)| println!("{}: {}", k, v));
+        println!(
+            "RPC_PROXY_POSTGRES_URI b.2: {:?}",
+            std::env::var_os("RPC_PROXY_POSTGRES_URI").unwrap()
+        );
         println!(
             "RPC_PROXY_POSTGRES_URI b.2: {}",
             std::env::var("RPC_PROXY_POSTGRES_URI").unwrap()
--- a/src/handlers/wallet/get_calls_status.rs
+++ b/src/handlers/wallet/get_calls_status.rs
@@ -310,7 +310,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_get_calls_status() {
-        println!("RPC_PROXY_POSTGRES_URI: {}", std::env::var("RPC_PROXY_POSTGRES_URI").unwrap());
+        println!("RPC_PROXY_POSTGRES_URI -1: {}", std::env::var("RPC_PROXY_POSTGRES_URI").unwrap());
         // let anvil = Anvil::new().spawn();
         let config = Config::local();
         let faucet = anvil_faucet(config.endpoints.rpc.base_url.clone()).await;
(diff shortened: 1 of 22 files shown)


src/providers/pimlico.rs
```diff
@@ -31,28 +33,28 @@ impl BundlerOpsProvider for PimlicoProvider {
     async fn bundler_rpc_call(
         &self,
         chain_id: &str,
-        id: u64,
-        jsonrpc: &str,
+        id: Id,
+        jsonrpc: Arc<str>,
```
geekbrother, line 37: Nice catch 👍

src/test_helpers.rs
geekbrother, line 1: Can we move this to `/tests` directory?  Looks like it fits better there.🙏

geekbrother (approved):
Great work on tests 👍 Thanks.

src/handlers/wallet/get_calls_status.rs
```diff
@@ -0,0 +1,453 @@
...
+    #[error("Internal error")]
+    InternalError(GetCallsStatusInternalError),
+}
+
+#[derive(Error, Debug)]
+pub enum GetCallsStatusInternalError {}
+
+impl IntoResponse for GetCallsStatusError {
+    fn into_response(self) -> Response {
+        #[allow(unreachable_patterns)] // TODO remove
```
tomiir, line 77: is this ok?

src/handlers/wallet/get_calls_status.rs
```diff
@@ -0,0 +1,453 @@
...
+            chain_id,
+        },
+        false,
+    ));
+
+    let receipt = provider
+        .get_user_operation_receipt(request.0 .0.user_op_hash)
+        .await
+        .unwrap();
+    // TODO handle None as CallStatus::Pending
```
tomiir, line 135: I guess this is a leftover?

src/test_helpers.rs
chris13524: No because some of these helpers are used by unit tests inside `src`, and `src` doesn't have access to `tests`

src/handlers/wallet/get_calls_status.rs
chris13524: Yeah this triggers because there's only 1 error that exists right now, `InternalError`, so clippy is saying I don't need the catch-all `e => ...` case. But we'll need that later when we add more error variants

src/handlers/wallet/get_calls_status.rs
chris13524: This is some implementation work that's still required for best error handling. In the case where you submitted a user op but it hasn't been confirmed yet we need to return an appropriate response instead of panicking. This test case is captured below in `// TODO test case:`

Didn't implement in this PR as it's an edge case and wasn't sure we'd encounter it in the short-term, but I captured in this issue: https://linear.app/reown/issue/CR-553/wallet-getcallsstatus-handle-pending-case

src/providers/pimlico.rs
chris13524: Yeah doesn't hurt to be a little more generic and lean into alloy types. These types were required for something in this PR anyway

PR #807, opened 2024-10-08
feat(CoSigner): changing the get session context endpoint

geekbrother:
# Description

This PR changed the session permission get context endpoint to be `GET /v1/sessions/:address/getcontext?projectId={projectId}&pci={pci}` instead of `GET /v1/sessions/:address/:pci` to reflect that we are getting only the PCI context and shorting the URL.

## How Has This Been Tested?

* The current integration test was updated.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #809, opened 2024-10-11
feat(CoSigner): implementing `contract-call` and `native-token-transfer` permissions check

geekbrother:
# Description

This PR introduces the draft testing implementation for the following permissions check during the CoSign process:

* `contract-call` permission:
Extracting the contract call address from the bundler's execution operation and comparing it to the permissions allowed for the address call.

* `native-token-transfer` permission:
Extracted values from the bundler's execution batch are summarized and checked with the allowance.

Currently missing:
* Check for the period, start, and expiration,
* Tracking the already spent amount.

## How Has This Been Tested?

* Manual testing only, integration tests will be added before merging.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/permissions.rs
+++ b/src/utils/permissions.rs
@@ -1,0 +1,86 @@
+use {
+    crate::{
+        error::RpcError,
+        utils::sessions::{
+            extract_addresses_from_execution_batch, extract_values_from_execution_batch,
+        },
+    },
+    ethers::{
+        abi::Token,
+        types::{H160, U256},
+    },
+    serde::{Deserialize, Serialize},
+    serde_json::Value,
+    strum_macros::{Display, EnumIter, EnumString},
+    tracing::error,
+};
+
+/// Supported permission types
+#[derive(Clone, Copy, Debug, EnumString, EnumIter, Display, PartialEq)]
+#[strum(serialize_all = "kebab-case")]
+pub enum PermissionType {
+    ContractCall,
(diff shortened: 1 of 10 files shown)


nopestack (approved):
1st pass, looks good though some tests so we can verify what it does would be appreciated

src/utils/permissions.rs
```diff
@@ -0,0 +1,86 @@
...
+pub fn native_token_transfer_permission_check(
+    execution_batch: Vec<Token>,
+    native_token_transfer_permission_data: NativeTokenTransferPermissionData,
+) -> Result<(), RpcError> {
+    let execution_values = extract_values_from_execution_batch(execution_batch.clone())?;
+    let allowance = native_token_transfer_permission_data.allowance;
+    // summ execution values from the execution batch and check if it is less than or equal to the allowance
+    let sum: U256 = execution_values
+        .iter()
+        .fold(U256::zero(), |acc, &x| acc + x);
```
chris13524, line 73: Can't you use `sum()` here?

```suggestion
        .sum();
```

chris13524: Guess not :(

chris13524 (approved):
LGTM

src/utils/permissions.rs
geekbrother: Nope, `sum` is not supported, that's why I used that way ((

PR #812, opened 2024-10-17
fix(CoSigner): decoding the ABI calldata from Safe format, PCI expiration and revocation check.

geekbrother:
# Description

This PR adds the following changes:

* Decoding the calldata ABI with packed (one transaction) and unpacked formats to support the `Safe` format,
* Using the yttrium smart_accounts->safe implementation for decoding,
* Removing `ethers-rs` in favor of `alloy`,
* Adding CoSigner check if the permission was revoked,
* Adding CoSigner check if the permission was expired.

## How Has This Been Tested?

New unit tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/sessions.rs
+++ b/src/utils/sessions.rs
@@ -1,7 +1,6 @@
 use {
     crate::error::RpcError,
     alloy::{
-        dyn_abi::DynSolValue,
         primitives::{Address, Bytes, U256},
         sol,
         sol_types::{SolCall, SolType},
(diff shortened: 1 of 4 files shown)


src/utils/sessions.rs
```diff
@@ -1,158 +1,124 @@
...
-          "outputs": [],
-          "stateMutability": "payable"
-      }
-    ]
-    "#;
-
-    let execute_abi: Abi = serde_json::from_str(execute_abi_json)?;
-    let execute_function = execute_abi.function("execute").map_err(|e| {
-        RpcError::AbiDecodingError(format!("Failed to parse execute function: {}", e))
+    call_data_bytes: &[u8],
```
geekbrother, line 19: Will move this function to the yttrium library later so it will be in one place with encoding/decoding.

src/utils/sessions.rs
```diff
@@ -1,158 +1,124 @@
...
+        let value = U256::from_be_bytes::<32>(value_bytes_array);
 
-    if execution_batch.is_empty() {
-        return Err(RpcError::AbiDecodingError("executionBatch is empty".into()));
-    }
+        // callData (remaining bytes)
+        let call_data = execution_calldata_bytes[52..].to_vec();
 
-    Ok(execution_batch.clone())
+        let transaction = DynSolValue::Tuple(vec![
```
chris13524, line 81: I find it odd that we are using `DynSolValue` instead of a more well-defined custom type. Isn't this just used by `extract_addresses_from_execution_batch()` to extract addresses?

I would think using a struct like this would be better:

```rust
struct Execution {
  address: Address,
  value: U256,
  call_data: Bytes,
}
```

chris13524 (approved):

src/utils/sessions.rs
geekbrother: Good catch! Removed `DynSolValue`.

PR #826, opened 2024-11-04
feat: updating Cargo dependencies and yttrium library

geekbrother:
# Description

This PR updates Cargo dependencies. The reason for this update is to fetch the latest yttrium library. 
Fixes to the `get_calls_status` handler were added since the response for the bundler's receipt was changed in the yttrium.
The [Unicode-3](https://github.com/unicode-org/icu/blob/main/LICENSE) license was added to thee allowed list.

## How Has This Been Tested?

* Tested by current integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,9 +137,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.44"
+version = "0.1.46"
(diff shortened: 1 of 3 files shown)


chris13524 (approved):
LGTM

PR #829, opened 2024-11-05
fix(CA): fixing the nonce calculation, decreasing gas estimate, transactions chain ID fix

geekbrother:
# Description

This PR fixes the following issues:

* Proper chain ID passing when calculating the nonce to fix the `nonce is too high` error,
* Decreasing the default gas estimate from x4 to x2,
* Fixing chain ID passing to the transactions list.

## How Has This Been Tested?

* Updated integration test

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/chain_orchestrator.test.ts
+++ b/integration/chain_orchestrator.test.ts
@@ -19,8 +19,8 @@ describe('Chain abstraction orchestrator', () => {
   const amount_multiplier = 5; // +5% topup
   // How much needs to be topped up
   const amount_to_topup = (amount_to_send - usdc_funds_on_optimism) * (100 + amount_multiplier) / 100;
-  // Default gas esimation is default with 4x increase
-  const gas_estimate = "0xa69ac";
+  // Default gas esimation is default with 2x increase
+  const gas_estimate = "0x534d6";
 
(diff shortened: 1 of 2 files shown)


integration/chain_orchestrator.test.ts
```diff
@@ -19,8 +19,8 @@ describe('Chain abstraction orchestrator', () => {
   const amount_multiplier = 5; // +5% topup
   // How much needs to be topped up
   const amount_to_topup = (amount_to_send - usdc_funds_on_optimism) * (100 + amount_multiplier) / 100;
-  // Default gas esimation is default with 4x increase
-  const gas_estimate = "0xa69ac";
+  // Default gas esimation is default with 2x increase
```
chris13524, line 22: Why are we hardcoding this 2x thing? Don't we know the actual amount with `eth_estimateGas`?

chris13524 (approved):

integration/chain_orchestrator.test.ts
geekbrother: That's because we hardcoded the x2 [here](https://github.com/reown-com/blockchain-api/pull/829/files#diff-3abd56c745d060c84c275ce850fc9b0b31b61b601521e33258e38804eb20fc0bR205), will do this followup now, but we need to ship this asap.

PR #832, opened 2024-11-08
feat(CA): removing the `check` endpoint, updating the `route` and `status` responses schema

geekbrother:
# Description

This PR making the following changes to the chain orchestrator API endpoints:

* The `/check` endpoint was removed,
* The `/route` endpoint is insufficient balance and no route errors are now in the `error` schema,
* Removing the check for the initial transaction gas,
* Removing the initial transaction from the returned transactions list.

## How Has This Been Tested?

Integration tests were updated.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #833, opened 2024-11-11
fix(providers): adding proper error handling for the Pokt(Groove) provider

geekbrother:
# Description

This PR adds proper handling of the internal node error response for the Pokt/Groove provider.
When the node becomes unhealthy there are `HTTP 200 Ok` responses with different response schema: `request_id` and `error`. This PR adds proper handling of these cases as an HTTP 5xx error response depending on the error message.

## How Has This Been Tested?

* Tested manually on the unhealthy Arbitrum node responses.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/pokt.rs
+++ b/src/providers/pokt.rs
@@ -11,6 +11,7 @@ use {
     },
     hyper::{self, client::HttpConnector, Client, Method, StatusCode},
     hyper_tls::HttpsConnector,
+    serde::Deserialize,
     std::collections::HashMap,
     tracing::debug,
 };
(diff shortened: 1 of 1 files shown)


src/providers/pokt.rs
```diff
@@ -38,32 +45,8 @@ impl Provider for PoktProvider {
 
 #[async_trait]
 impl RateLimited for PoktProvider {
-    // async fn is_rate_limited(&self, response: &mut Response) -> bool
```
geekbrother, line 41: Removing this and adding the proper handling along with thee `-32004` code below.

chris13524 (approved):

PR #845, opened 2024-11-26
fix(fungibles): adding `address` field to the fungibles price response schema

geekbrother:
# Description

According to [the SPEC](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-api#success-response-body-6), this PR adds the missing `address` field to the fungibles price response.

## How Has This Been Tested?

Integration tests were updated.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/fungible_price.rs
+++ b/src/handlers/fungible_price.rs
@@ -30,6 +30,7 @@ pub struct PriceResponseBody {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct FungiblePriceItem {
+    pub address: String,
     pub name: String,
     pub symbol: String,
     pub icon_url: String,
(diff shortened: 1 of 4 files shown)


src/handlers/fungible_price.rs
```diff
@@ -30,6 +30,7 @@ pub struct PriceResponseBody {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct FungiblePriceItem {
+    pub address: String,
```
chris13524, line 33: It's a CAIP-10 not an address, so this is a little bit wrong. What if we called the field `caip10`? Another option could be `account_id` as that's what it is called in the [standard](https://chainagnostic.org/CAIPs/caip-10)

```suggestion
    pub caip10: String,
```

chris13524 (approved):

src/handlers/fungible_price.rs
geekbrother: Thanks for the comment. We have `address` as CAIP-10 in the request/response everywhere in the [SPEC](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-api#request-body-4). If we want to change this we need to change the `address` name in the request as well and this will be a breaking change.

PR #847, opened 2024-11-27
feat(CA): fine tuning the bridging amount estimation

geekbrother:
# Description

This PR adds a fine-tuning to the bridging amount estimation by using the following approach:
* Get the bridging quote from the original transfer amount top-up;
* Calculate the bridging fee from the quote final amount response;
* Update the transfer amount top-up on the previous quote fee;
* Multiply the result by 2% slippage;
* Get the new quote with the final top-up amount.

## How Has This Been Tested?

The integration test was updated.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM assuming we add that check that the second quote's to_amount is greater than or equal to the send amount

PR #850, opened 2024-11-29
feat(providers): removing SolScan API v1 support

geekbrother:
# Description

Due to the sunset of the SolScan API v1 support this PR removed the v1 accounts balance endpoint use in favor of the [v2 account tokens balance endpoint](https://pro-api.solscan.io/pro-api-docs/v2.0/reference/v2-account-token-accounts).

## How Has This Been Tested?

* Current changes must be non-breaking, the current integration tests must pass.

## Due Diligence

* [ ] Breaking change


src/providers/solscan.rs
```diff
@@ -31,16 +31,14 @@ use {
...
 const ACCOUNT_HISTORY_URL: &str = "https://pro-api.solscan.io/v2.0/account/transfer";
 const TOKEN_METADATA_URL: &str = "https://pro-api.solscan.io/v2.0/token/meta";
 const TOKEN_PRICE_URL: &str = "https://pro-api.solscan.io/v2.0/token/price";
 const ACCOUNT_DETAIL_URL: &str = "https://pro-api.solscan.io/v2.0/account/detail";
 
-const WSOL_TOKEN_ADDRESS: &str = "So11111111111111111111111111111111111111112";
-
 // Caching TTL paramters
-const METADATA_CACHE_TTL: u64 = 60 * 60; // 1 hour
+const METADATA_CACHE_TTL: u64 = 60 * 60 * 24; // 24 hours
```
geekbrother, line 41: Increasing the token's metadata caching TTL to 24 hours, since I don't think the token's icon can change too often and the token name and symbol are stable.

chris13524: Could go to 7 days IMO

chris13524 (approved):
LGTM

src/providers/solscan.rs
geekbrother: Thanks for the comment 🙏 I was thinking of the case when the token icon URL was changed. I believe we can survive a few hours of the broken image in this case when using the 24 hours. But we should force clean the cache if that's 7 days.

PR #853, opened 2024-12-04
fix(CA): adding an exclusion into the bridging assets search

geekbrother:
# Description

This PR fixes the bug with the flaky route responses (empty routes) even when funds and bridging are available by adding the initial asset exclusion from the assets searching for the bridging. This causes a `no bridging needed` response when the initial asset is found.

## How Has This Been Tested?

Tested multiple times by requesting and reproducing the flaky issue.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -128,6 +128,8 @@ async fn handler_internal(
         query_params.project_id.as_ref().to_string(),
         erc20_topup_value,
         from_address,
+        initial_transaction.chain_id.clone(),
+        to_address,
     )
     .await?
     else {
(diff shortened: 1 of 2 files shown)


src/handlers/chain_agnostic/route.rs
```diff
@@ -141,11 +143,6 @@ async fn handler_internal(
     let bridge_contract = bridging_asset.contract_address;
     let current_bridging_asset_balance = bridging_asset.current_balance;
 
-    // Skip bridging if that's the same chainId and contract address
```
geekbrother, line 144: Removing this check since we will use the exclusion above.

chris13524 (approved):

PR #856, opened 2024-12-04
fix(o11y): change chain availability to unavailability Grafana panel

geekbrother:
# Description

This PR fixes the ChainID availability Grafana panel to use the unavailability metric and adds an alert instead of calculating the availability percentage since there is an error because of no data for some chains for the `no_providers_for_chain_counter` if the chain is constantly available.

## How Has This Been Tested?

Tested manually by putting the updated query into the Grafana.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):

arein (approved):
Thank you! :)

PR #859, opened 2024-12-06
feat(CA): adding transaction simulation

geekbrother:
# Description

This PR adds the transaction simulation using the Tenderly simulation provider to the chain abstraction route endpoint to check for the ERC20 asset changes in case the initial transaction is not an ERC20 transfer.

## How Has This Been Tested?

Integration tests update TBD.

## Due Diligence

* [ ] Breaking change


src/handlers/chain_agnostic/mod.rs
```diff
@@ -145,3 +168,102 @@ pub async fn check_bridging_for_erc20_transfer(
...
+        let mut account_state = HashMap::new();
+        account_state.insert(
+            // Since we are using only USDC for the bridging,
+            // we can hardcode the storage slot for the contract which is 9
+            compute_simulation_storage_slot(transaction.from, 9),
+            compute_simulation_balance(99000000000),
+        );
+        for contract in assets_contracts {
+            state_overrides.insert(
+                Address::from_str(&contract).unwrap_or_default(),
```
chris13524, line 222: I think we should to be offensive here, not defensive? Anyway, why isn't `contract` already an `Address` which would avoid this unwrap?

```suggestion
                Address::from_str(&contract).unwrap(),
```

src/handlers/chain_agnostic/mod.rs
```diff
@@ -145,3 +168,102 @@ pub async fn check_bridging_for_erc20_transfer(
     }
     Ok(None)
 }
+
+/// Compute the simulation state override balance for a given balance
+pub fn compute_simulation_balance(balance: u128) -> Bytes {
+    let raw_bytes = balance.to_be_bytes();
+    let mut padded = [0u8; 32];
+    padded[32 - raw_bytes.len()..].copy_from_slice(&raw_bytes);
+    Bytes::from(padded.to_vec())
```
chris13524, line 177: I think you can do `Bytes::from(B128::from(balance))`?

src/handlers/chain_agnostic/mod.rs
```diff
@@ -145,3 +168,102 @@ pub async fn check_bridging_for_erc20_transfer(
...
+}
+
+/// Compute the storage slot for a given address and slot number to use in the
+/// simulation state overrides
+/// https://docs.tenderly.co/simulations/state-overrides#storage-slot-calculation
+pub fn compute_simulation_storage_slot(address: Address, slot_number: u64) -> Bytes {
```
chris13524, line 183: We shouldn't need to convert to/from hex in order to concatenate some bytes together

chris13524 (commented):
Should this PR set the `gas` field somewhere?

src/handlers/chain_agnostic/mod.rs
geekbrother: Good catch, thank you! Updated to use `Address` everywhere instead.

src/handlers/chain_agnostic/mod.rs
geekbrother: Updated to use B256 where applied 👍 Thanks!

src/handlers/chain_agnostic/mod.rs
geekbrother: Updated the implementation a bit, but we still need to contact two B256 and use this for the keccak.

geekbrother:
> Should this PR set the `gas` field somewhere?

This is a quite big PR and a lot of changes, let's merge this first and I'll add gas with the caching as a follow-up (don't want to make two monster stacked PRs 😅).

chris13524 (approved):

PR #869, opened 2024-12-13
feat(CA): implementing cached gas estimation for the initial transaction

geekbrother:
# Description

This PR implements the cached gas estimation for the initial transaction by using the following flow:

* If the transaction is an ERC20 transfer, check the cached gas used value for the ERC20 token contract and ChainID,
* If the cached value is found - update the initial transaction `gas` with the cached value,
* If the cached value was not found - simulate the initial transaction, save the resulting gas used value to the cache by using the token contract and ChainID, and update the initial transaction `gas` with the cached value.

The cache record TTL for the gas estimated value is 24 hours.

The slippage of 3% is added to the gas estimation to cover gas changes.

## How Has This Been Tested?

Manually the integration test was updated.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -43,6 +43,9 @@ use {
 // Using default with 6x increase
 const DEFAULT_GAS: i64 = 0x029a6b * 0x6;
 
+// Slippage for the gas estimation
+const ESTIMATED_GAS_SLIPPAGE: i8 = 3;
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct QuoteRoute {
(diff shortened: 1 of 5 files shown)


src/handlers/chain_agnostic/route.rs
```diff
@@ -112,14 +112,45 @@ async fn handler_internal(
...
+                            &initial_transaction,
+                        )
+                        .await?;
+                        state
+                            .providers
+                            .simulation_provider
+                            .set_cached_erc20_gas_estimation(
+                                &request_payload.transaction.chain_id,
+                                to_address,
+                                simulated_gas_used,
```
chris13524, line 139: Should we cache something that's a bit larger, e.g. 10%? I think gas consumed can vary slightly between transactions

chris13524 (approved):
This looks good to me, we can test with the initial gas fee changed. But I would like too see the gas fees estimated for route txns as well. [Slack conversation](https://reown-inc.slack.com/archives/C0816SK4877/p1734095912521149).

src/handlers/chain_agnostic/route.rs
geekbrother: Makes sense to add slippage, but I think 10% is a bit big, I think 2-3% should be ok. Thanks!

geekbrother: A slippage of 3% was added to the estimation.

PR #872, opened 2024-12-16
fix: refactor txn types

chris13524:
# Description

Depends on https://github.com/reown-com/yttrium/pull/98

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -76,7 +76,7 @@ async fn handler_internal(
         to: request_payload.transaction.to,
         value: request_payload.transaction.value,
         gas_limit: U64::from(DEFAULT_GAS),
-        data: request_payload.transaction.data.clone(),
+        input: request_payload.transaction.input.clone(),
         nonce: U64::ZERO,
         chain_id: request_payload.transaction.chain_id.clone(),
     };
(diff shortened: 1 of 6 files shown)


geekbrother:
Fixing to fit the latest schema changes and initial tx nonce calculation in 112f5b91b3c8d8e539ab02bb7095cbbd90262dc4

geekbrother (approved):
Thanks!

src/handlers/chain_agnostic/route.rs
```diff
@@ -72,11 +71,30 @@ async fn handler_internal(
         .validate_project_access_and_quota(query_params.project_id.as_ref())
         .await?;
 
-    let mut initial_transaction = request_payload.transaction.clone();
+    let mut initial_transaction = Transaction {
+        from: request_payload.transaction.from,
+        to: request_payload.transaction.to,
+        value: request_payload.transaction.value,
+        gas_limit: U64::from(DEFAULT_GAS),
```
chris13524, line 78: When no bridging is required we are returning this wrong `gas_limit`. We should avoid mutable variables here and only construct `Transaction` when all its fields are valid.

geekbrother: In this case we will simulate all transactions even if we know that there will be no bridging, like native token transfer.

chris13524: Yes indeed; I think we need to. Otherwise we should return another transaction type that doesn't have `gas_limit` set. Can't be returning unusable values since wallet expects Blockchain API to return correct gas limit

chris13524: Also we can charge them 1 RPC worth of credits for this; since they'd have to make the RPC call themselves anyway so this reduces latency for them

PR #883, opened 2025-01-02
feat(balances): adding balance providers weights and retrying

geekbrother:
# Description

This PR adds multiple providers, weights, and retries for balance providers by following the same provider traits that we have for RPC providers. Also, refactoring `provider` traits to `rpc_provider` allows for the expansion of more providers (balance form this PR, etc).

## How Has This Been Tested?

* Current integration tests for the balance are passing.

## Due Diligence

* [ ] Breaking change


src/analytics/balance_lookup_info.rs
```diff
@@ -1,14 +1,9 @@
...
-    serde::Serialize,
-    std::{sync::Arc, time::Duration},
-};
+use {parquet_derive::ParquetRecordWriter, serde::Serialize, std::sync::Arc};
 
 #[derive(Debug, Clone, Serialize, ParquetRecordWriter)]
 #[serde(rename_all = "camelCase")]
 pub struct BalanceLookupInfo {
     pub timestamp: chrono::NaiveDateTime,
-    pub latency_secs: f64,
```
geekbrother, line 11: We have per-providers latency, so this should be removed.

chris13524: Did data team already remove `latency_secs` from the schema? Otherwise this could cause null errors

chris13524 (approved):
LGTM! Thanks!!

PR #890, opened 2025-01-09
feat(providers): extending Dune provider to serve Solana balances

geekbrother:
# Description

This PR extends the Dune provider to serve Solana balance requests along with the Eip155.

## How Has This Been Tested?

Current Solana balance integration tests were passed.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/dune.rs
+++ b/src/providers/dune.rs
@@ -21,28 +21,32 @@ use {
     url::Url,
 };
 
+const DUNE_API_BASE_URL: &str = "https://api.dune.com/api";
+
 /// Native token icons, since Dune doesn't provide them yet
 /// TODO: Hardcoding icon urls temporarily until Dune provides them
 pub static NATIVE_TOKEN_ICONS: phf::Map<&'static str, &'static str> = phf_map! {
       // Ethereum
-      "ETH" => "https://assets.coingecko.com/coins/images/279/small/ethereum.png",
+      "ETH" => "https://cdn.jsdelivr.net/gh/trustwallet/assets@master/blockchains/ethereum/info/logo.png",
       // Polygon
-      "POL" => "https://assets.coingecko.com/coins/images/32440/standard/polygon.png",
+      "POL" => "https://cdn.jsdelivr.net/gh/trustwallet/assets@master/blockchains/polygon/info/logo.png",
       // xDAI
-      "XDAI" => "https://assets.coingecko.com/coins/images/662/standard/logo_square_simple_300px.png",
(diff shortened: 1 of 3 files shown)


src/providers/dune.rs
```diff
@@ -21,28 +21,32 @@ use {
...
 };
 
+const DUNE_API_BASE_URL: &str = "https://api.dune.com/api";
+
 /// Native token icons, since Dune doesn't provide them yet
 /// TODO: Hardcoding icon urls temporarily until Dune provides them
 pub static NATIVE_TOKEN_ICONS: phf::Map<&'static str, &'static str> = phf_map! {
       // Ethereum
-      "ETH" => "https://assets.coingecko.com/coins/images/279/small/ethereum.png",
+      "ETH" => "https://cdn.jsdelivr.net/gh/trustwallet/assets@master/blockchains/ethereum/info/logo.png",
```
geekbrother, line 30: Using the same TrustWallet icons across providers.

src/providers/dune.rs
```diff
@@ -101,19 +101,81 @@ impl BalanceProvider for DuneProvider {
...
-        let body = response.json::<DuneResponseBody>().await?;
+        Ok(response.json::<DuneBalanceResponseBody>().await?)
+    }
+
+    async fn get_solana_balance(
+        &self,
+        address: String,
+        metrics: Arc<Metrics>,
+    ) -> RpcResult<DuneBalanceResponseBody> {
+        let base = format!("{}/echo/beta/balances/svm/{}", DUNE_API_BASE_URL, &address);
```
chris13524, line 122: Possible injection attack with `address` here? Not sure if sanitized already or not

src/providers/dune.rs
```diff
@@ -132,13 +204,20 @@ impl BalanceProvider for DuneProvider {
                         }
                     },
                     symbol: symbol.clone(),
-                    chain_id: Some(caip2_chain_id),
+                    chain_id: Some(caip2_chain_id.clone()),
                     address: {
                         // Return None if the address is native for the native token
                         if f.address == "native" {
-                            None
+                            // UI expecting `None`` for the Eip155 and Solana's native
```
chris13524, line 211: This sounds more complex than it should be, hoping we can agree on a common format across namespaces (and I would prefer not using magic token addresses)

chris13524 (approved):
LGTM

src/providers/dune.rs
geekbrother: Good catch! But we checking it [here](https://github.com/reown-com/blockchain-api/blob/c549daae841a09744811c90da8377ee89b621380/src/handlers/balance.rs#L105).

PR #906, opened 2025-01-20
feat(balance): implementing tokens metadata caching between Zerion and Dune providers

geekbrother:
# Description

This PR implements tokens metadata shared cache between Zerion and Dune providers for the address balances endpoint. The reason to use a shared cache is that when using multiple providers there can be a difference in the token metadata (like different names or icons). 
By using a long-living (1 day) cache with the "first hit, first saved" rule we will use the same metadata for every provider and fix icons flipping on every request and keeping the metadata updated in case some of the providers will be added or removed.
Since the tokens metadata doesn't update often, keeping the 1 day caching seems reasonable.

## How Has This Been Tested?

* Current balance integration tests are passed.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/balance.rs
+++ b/src/handlers/balance.rs
@@ -4,6 +4,7 @@ use {
         analytics::{BalanceLookupInfo, MessageSource},
         error::RpcError,
         state::AppState,
+        storage::KeyValueStorage,
         utils::{crypto, network},
     },
     axum::{
@@ -14,13 +15,17 @@ use {
     ethers::{abi::Address, types::H160},
     hyper::HeaderMap,
     serde::{Deserialize, Serialize},
-    std::{net::SocketAddr, sync::Arc},
+    std::{net::SocketAddr, sync::Arc, time::Duration},
     tap::TapFallible,
     tracing::log::{debug, error},
     wc::future::FutureExt,
 };
 
+// Empty address for the contract address mimicking the Ethereum native token
+pub const H160_EMPTY_ADDRESS: H160 = H160::repeat_byte(0xee);
+
 const PROVIDER_MAX_CALLS: usize = 2;
+const METADATA_CACHE_TTL: Duration = Duration::from_secs(60 * 60 * 24 * 7); // 1 week
 
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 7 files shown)


geekbrother:
Decreasing the cache from 7 days to one day, since it's possible that the icon URL can change and we shouldn't wait 7 days for the cache invalidation.

src/providers/zerion.rs
```diff
@@ -475,34 +482,57 @@ impl BalanceProvider for ZerionProvider {
...
+                .implementations
+                .iter()
+                .find(|impl_| impl_.chain_id == chain_id_human)
+                .and_then(|impl_| impl_.address.clone());
+
+            let token_address_strict = token_address
+                .clone()
+                .unwrap_or_else(|| H160_EMPTY_ADDRESS.to_string());
+            let chain_id = crypto::ChainId::to_caip2(&chain_id_human);
+            let chain_id_strict = chain_id.clone().unwrap_or_else(|| "eip155:1".to_string());
```
chris13524, line 500: Why are we defaulting to mainnet here?

src/handlers/balance.rs
```diff
@@ -61,6 +66,46 @@ pub struct BalanceQuantity {
...
+}
+
+pub async fn set_cached_metadata(
+    cache: &Option<Arc<dyn KeyValueStorage<TokenMetadataCacheItem>>>,
+    caip10_token_address: &str,
+    item: &TokenMetadataCacheItem,
+) {
+    if let Some(cache) = cache {
+        cache
+            .set(
```
chris13524, line 99: https://www.notion.so/walletconnect/Setting-cache-should-be-async-error-free-15e3a661771e806c81fec019bd0ba506?pvs=4

chris13524 (commented):
Some preliminary comments while I still review

src/handlers/balance.rs
geekbrother: Totally! Good catch, thanks! Updated.

src/providers/zerion.rs
geekbrother: That's a good point on this. Suppose we can't convert the human-readable chain name to CAIP-2 and push it to the cache with the default `eip155:1` it will not be used in the Dune later since it's probably not a mainnet and we should skip caching for the unknown token chain. I have updated this part to skip caching in this case in https://github.com/reown-com/blockchain-api/pull/906/commits/bd56a25ed6decd67ae5f6380ee380bc34cbbb53a
Thanks!

chris13524 (approved):
LGTM

PR #907, opened 2025-01-22
fix(Dune): checking for the cached metadata first

geekbrother:
# Description

This PR changes when to call for the cached metadata in the Dune balance provider implementation. Currently, if metadata is not provided the asset is skipped as spam and then we check for the cached item. This PR changed this to check the cached asset metadata first to show the asset even if it's not present in the Dune response.

## How Has This Been Tested?

* Current balance integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/dune.rs
+++ b/src/providers/dune.rs
@@ -182,13 +182,6 @@ impl BalanceProvider for DuneProvider {
 
         let mut balances_vec = Vec::new();
         for f in balance_response.balances {
-            // Skip if missing required fields as a possible spam token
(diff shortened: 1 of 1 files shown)


src/providers/dune.rs
```diff
@@ -182,13 +182,6 @@ impl BalanceProvider for DuneProvider {
 
         let mut balances_vec = Vec::new();
         for f in balance_response.balances {
-            // Skip if missing required fields as a possible spam token
```
geekbrother, line 185: We are moving this continuation after the cached entry check.

nopestack (approved):

PR #909, opened 2025-01-23
feat(IRN): updating irn client to `wcn_replication` and switching to the mainnet

geekbrother:
# Description

This PR updates the IRN client to the latest `wcn_replication`, switching to the mainnet. 
Internal methods were refactored to use `Strings` as keys and `Bytes (Vec<u8>)` as values to be in sync with the client. 
We are using the Serde serialization for the values data. Hence, it makes sense to serialize/deserialize directly to bytes instead of strings. `hfields` method was refactored to be `hscan` that returns a vector of tuples with `(field_name, value)` pairs.
The IRN node environment variable was updated to be `nodes` including a list of WCN nodes in a MultiAddr format.

## How Has This Been Tested?

* Tested on the local IRN cluster using `test_irn_client_set_get_del()` and `test_irn_client_hashmap()` cargo tests which are included in the implementation.

## Merging policy 🚧

* [x] Update TFC `irn_nodes` with actual nodes list,
* [x] Add PeerID to allow-list to connect to nodes.

### DIFF
--- a/src/handlers/sessions/list.rs
+++ b/src/handlers/sessions/list.rs
@@ -80,7 +80,7 @@ async fn handler_internal(
         .add_irn_latency(irn_call_start, OperationType::Hscan);
 
     let mut result_pcis: Vec<Pci> = Vec::new();
-    for (pci, entity) in pcis {
+    for (_, entity) in pcis {
         let storage_permissions_item = serde_json::from_slice::<StoragePermissionsItem>(&entity)?;
 
         // Get project data
@@ -96,7 +96,7 @@ async fn handler_internal(
                 url: None,
                 icon_url: None,
             },
-            pci,
+            pci: storage_permissions_item.pci,
             expiry: storage_permissions_item.expiry,
             created_at: storage_permissions_item.created_at,
             permissions: storage_permissions_item.permissions,
--- a/src/handlers/sessions/list.rs
+++ b/src/handlers/sessions/list.rs
@@ -15,6 +15,9 @@ use {
     wc::future::FutureExt,
 };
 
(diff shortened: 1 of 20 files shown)


src/handlers/sessions/list.rs
```diff
@@ -69,24 +72,16 @@ async fn handler_internal(
 
     // get all permission control identifiers for the address
     let irn_call_start = SystemTime::now();
-    let pcis = irn_client.hfields(address.clone()).await?;
+    let (pcis, _) = irn_client
+        .hscan(address.clone(), MAX_PCIS_COUNT, None)
```
heilhead, line 76: One thing to be aware of when working with WCN (same as it was in the legacy version, actually) is key collisions. Since maps support prefix scan, you need to be sure that different types of maps can't have overlap in keys. Also that maps and strings can't overlap. Otherwise `hscan` can potentially return seemingly broken data, or overwrite values set by the strings methods. This is an unlikely scenario, but something to be aware of.

There's two ways to prevent it:
- Prefix all keys describing their semantic types;
- Use a different namespace for every type of data stored.

In the relay we opted for [the prefix approach](https://github.com/reown-com/rs-relay/blob/master/src/storage/mod.rs#L163-L176).

heilhead (approved):

chris13524 (approved):
LGTM

geekbrother:
@chris13524 I'm removing the IRN image pull and up for the Mock bundler test in 2c8c8a96d402aff55e3c6be0aa4a1bc761b083b6 We are using the IRN nowhere in the mock test and not passing the `irn_nodes` variables in this case. So it doesn't make sense to spin it up here.

PR #910, opened 2025-01-24
feat(CA): adding cached bridging transaction gas estimation

geekbrother:
# Description

This PR adds the gas estimation for the bridging transaction by using the transaction simulation provider instead of the hardcoded value.

* The gas estimation is cached using the chainId and destination address composite key.
* The slippage for the estimated gas (to cover gas spikes between cache invalidation) is raised from 3% to 10%.
* The caching TTL is decreased from 24 hours to 4 hours for all gas caching (including ERC20 transfers) to cover gas spikes.

## How Has This Been Tested?

* Manually tested and partly covered by the current CA integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -52,7 +52,7 @@ use {
 const DEFAULT_GAS: i64 = 0x029a6b * 0x9;
 
 // Slippage for the gas estimation
-const ESTIMATED_GAS_SLIPPAGE: i8 = 3;
+const ESTIMATED_GAS_SLIPPAGE: i8 = 10;
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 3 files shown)


chris13524 (changes requested):
Looks like this only implements estimation for the transfer transaction, but not the approval transaction. We need it there too

geekbrother:
> Looks like this only implements estimation for the transfer transaction, but not the approval transaction. We need it there too (all transactions!)

I believe for the approval transaction estimation we can use `eth_estimateGas` instead of the simulation provider (since even bundled simulations are considered as separate calls). And I assume that will be faster. cc @lukaisailovic also

chris13524 (approved):
Approving this since it's technically better than current so should be good to merge, but it doesn't complete the [Linear issue](https://linear.app/reown/issue/RES-153/implementing-the-cached-gas-estimation-for-erc20)

> I believe for the approval transaction estimation we can use eth_estimateGas instead of the simulation provider

@geekbrother Yeah I think this should be fine! We don't need to modify the storage

PR #912, opened 2025-01-27
feat(IRN): removing Irn VPC peering

geekbrother:
# Description

This PR removes IRN VPC peering, since the IRN/WCN nodes are currently in a public network.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/res_network.tf
+++ b/terraform/res_network.tf
@@ -38,20 +38,6 @@ module "vpc" {
   vpc_flow_log_tags         = module.this.tags
 }
 
-resource "aws_vpc_peering_connection" "irn" {
-  vpc_id        = module.vpc.vpc_id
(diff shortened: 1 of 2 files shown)


xDarksome (approved):
I think you'll still need to remove it manually, that was the limitation of the tf resource iirc

geekbrother:
> I think you'll still need to remove it manually, that was the limitation of the tf resource iirc

Looks like the TF Plan succeeded, so maybe it will go through.

xDarksome:
> > I think you'll still need to remove it manually, that was the limitation of the tf resource iirc
> 
> Looks like the TF Plan succeeded, so maybe it will go through.

It will, but the peering will still be there

PR #916, opened 2025-01-28
feat(CA): adding cached gas estimation for the approval transaction

geekbrother:
# Description

This PR adds gas estimation for the approval transaction using the `eth_estimateGas` RPC call and caches the result to reuse it. The gas set/get cache method was extended to use the ERC20 function as an additional optional key parameter since the ERC20 `transfer` gas estimate and ERC20 `approval` functions will have different gas estimations but the same contract address.

## How Has This Been Tested?

* Current integration test: schema check only.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #919, opened 2025-01-29
fix(CA): using the bundled simulation for allowance and bridging transactions

geekbrother:
# Description

This PR changes the approach of the gas estimation from using the `eth_estimateGas` for the approval transaction and simulation for the bridging transaction to the bundled simulation for the approval and bridging transaction together.

* The bundled simulation provider trait was added and the implementation for the Tenderly provider;
* Flow of the estimation changed to simulate both transactions in a bundle;
* The proper check for the transaction simulation failure was added;
* The proper error handling for the simulation provider error and the simulation failure added;
* The estimated gas slippage decreased to 20% to cover the volatility between the estimation and the transaction execution.

## How Has This Been Tested?

The current integration test for the gas estimation shouldn't be zero which is the default if the simulation was not fulfilled.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -247,6 +247,12 @@ pub enum RpcError {
 
     #[error("Bridging final amount is less then expected")]
     BridgingFinalAmountLess,
+
+    #[error("Simulation provider unavailable")]
+    SimulationProviderUnavailable,
+
+    #[error("Simulation failed: {0}")]
+    SimulationFailed(String),
 }
 
 impl IntoResponse for RpcError {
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -247,6 +247,12 @@ pub enum RpcError {
 
     #[error("Bridging final amount is less then expected")]
     BridgingFinalAmountLess,
+
+    #[error("Simulation provider unavailable")]
+    SimulationProviderUnavailable,
+
+    #[error("Simulation failed: {0}")]
+    SimulationFailed(String),
```
geekbrother, line 255: Adding this specific error will result in HTTP 500 and an alert for us.

llbartekll (approved):

jakubuid (approved):
Reading more and more Rust code is fun :D

PR #925, opened 2025-01-31
fix(CA): comparing simulated transaction by the input

geekbrother:
# Description

This PR changes the simulated bridging transactions comparison by the data input instead of by the nonce, since this causes an error on the correct transaction because of the different nonce getting parameters are used in our implementation and in the Tenderly. It's much safer to check transactions using input data.

## How Has This Been Tested?

Manually tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -363,7 +363,7 @@ async fn handler_internal(
         );
         return Err(RpcError::BridgingFinalAmountLess);
     }
-    let final_bridging_fee = bridging_amount - required_topup_amount;
+    let final_bridging_fee = required_topup_amount - bridging_amount;
 
     // Build bridging transaction
     let bridge_tx = state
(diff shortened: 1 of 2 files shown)


src/providers/tenderly.rs
```diff
@@ -93,6 +93,7 @@ pub struct TokenInfo {
 pub enum TokenStandard {
     Erc20,
     Erc721,
+    NativeCurrency,
```
geekbrother, line 96: Adding an additional Token parameter that Tenderly responds with.

nopestack (approved):

src/handlers/chain_agnostic/route.rs
```diff
@@ -363,7 +363,7 @@ async fn handler_internal(
         );
         return Err(RpcError::BridgingFinalAmountLess);
     }
-    let final_bridging_fee = bridging_amount - required_topup_amount;
+    let final_bridging_fee = required_topup_amount - bridging_amount;
```
geekbrother, line 366: Fixing the wrong bridging fee calculation.

PR #926, opened 2025-01-31
feat(o11y): implementing metrics and Grafana panels for chain abstraction

geekbrother:
# Description

This PR implements metrics and Grafana o11y panels for the Chain Abstraction.

The following metrics and panels were created:
* Gas estimation histogram with by chain observability panel,
* `Insufficient funds` responses counter and observability panel,
* `No bridging needed` responses counter with by type observability panel,
* `No routes found` responses counter with by route observability panel.

When we get some data to make alert threshold adjustments, appropriate alerts will be added.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -20,6 +20,20 @@ use {
     },
 };
 
+#[derive(strum_macros::Display)]
+pub enum ChainAbstractionTransactionType {
+    Transfer,
+    Approve,
+    Bridge,
+}
+
+#[derive(strum_macros::Display)]
+pub enum ChainAbstractionNoBridgingNeededType {
+    NativeTokenTransfer,
+    AssetNotSupported,
+    SufficientFunds,
+}
+
 #[derive(Debug)]
 pub struct Metrics {
     pub rpc_call_counter: Counter<u64>,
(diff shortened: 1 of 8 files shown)


src/metrics.rs
```diff
@@ -642,6 +686,43 @@ impl Metrics {
...
+                otel::KeyValue::new("tx_type", tx_type.to_string()),
+            ],
+        );
+    }
+
+    pub fn add_ca_no_routes_found(&self, route: String) {
+        self.ca_no_routes_found_counter.add(
+            &otel::Context::new(),
+            1,
+            &[otel::KeyValue::new("route", route)],
```
chris13524, line 709: What is cardinality of this? https://stackoverflow.com/questions/79200910/how-much-cardinality-can-i-have-in-my-prometheus-metrics

chris13524 (approved):

src/metrics.rs
geekbrother: Thanks for pointing out this! I'm aware of this Prometheus limitation (bottleneck). 
Currently, we have not too much and I think it will not expand significantly. In total: we (will)have 4 assets (USD, USDC, USDS, ETH) for 3 chains (Arb, Op, Base) it's 24 total bridging routes. I believe we will have a few routes which are not available.

PR #930, opened 2025-02-04
feat(balance): adding the balance response caching

geekbrother:
# Description

This PR adds the balance response caching for 10 seconds TTL except if the `forceUpdate` parameter was provided where we need to query for the latest balance.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/balance.rs
+++ b/src/handlers/balance.rs
@@ -26,6 +26,7 @@ pub const H160_EMPTY_ADDRESS: H160 = H160::repeat_byte(0xee);
 
 const PROVIDER_MAX_CALLS: usize = 2;
 const METADATA_CACHE_TTL: Duration = Duration::from_secs(60 * 60 * 24); // 1 day
+const BALANCE_CACHE_TTL: Duration = Duration::from_secs(10); // 10 seconds
 
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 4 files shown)


arein (approved):
do we need some math on the impact of Redis?
We're at 20% usage now and this might increase it significantly

But I guess it just doesn't use the cache if Redis fails

geekbrother:
> do we need some math on the impact of Redis? We're at 20% usage now and this might increase it significantly
> 
> But I guess it just doesn't use the cache if Redis fails

That's a good point, thanks! Let's play safe and increase the cache to `t4g.medium` from `small` and I will observe the usage and downgrade it after.
Increasing to the `t4g.medium` in https://github.com/reown-com/blockchain-api/pull/930/commits/7975195a090bb98cadff67d3a8af7bb986fd414b

geekbrother:
> > do we need some math on the impact of Redis? We're at 20% usage now and this might increase it significantly
> > But I guess it just doesn't use the cache if Redis fails
> 
> That's a good point, thanks @arein ! Let's play safe and increase the cache to `t4g.medium` from `small` and I will observe the usage and downgrade it after. Increasing to the `t4g.medium` in [7975195](https://github.com/reown-com/blockchain-api/commit/7975195a090bb98cadff67d3a8af7bb986fd414b)

I have made clean calculations, and we shouldn't increase the cache node type to handle the caching. Reverting in https://github.com/reown-com/blockchain-api/pull/930/commits/d41c8d620f0f73bfbbdd7ac980401605cee8fd00

PR #939, opened 2025-02-06
fix: capitalizing the metadata token name, multiply gas estimation from 1Inch

geekbrother:
# Description

This PR makes a few small fixes:
* Making the token name based on the chain name (native token) capitalized;
* Applying the slippage (+10%) to the 1Inch gas estimation to cover differences for SA.

## How Has This Been Tested?

Current integration tests and manual testing.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -21,3 +21,34 @@ pub fn generate_random_string(len: usize) -> String {
         .take(len)
         .collect()
 }
+
+pub fn capitalize_first_letter(s: &str) -> String {
+    let mut c = s.chars();
+    match c.next() {
+        None => String::new(),
+        Some(first) => {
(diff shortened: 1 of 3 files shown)


nopestack (approved):

src/utils/mod.rs
```diff
@@ -21,3 +21,34 @@ pub fn generate_random_string(len: usize) -> String {
...
+        let input = "";
+        let expected = "";
+        assert_eq!(capitalize_first_letter(input), expected);
+
+        let input = "rust";
+        let expected = "Rust";
+        assert_eq!(capitalize_first_letter(input), expected);
+
+        let input = "rust world";
+        let expected = "Rust world";
```
nopestack, line 51: probably a good idea to later revisit to make it capitalize the first letter of every word in the string

PR #945, opened 2025-02-07
feat: adding SDK info `sv` and `st` query parameters for analytics

geekbrother:
# Description

This PR adds the following SDK info query parameters: 
* `sv` - version,
* `st` - type.

These parameters append to the analytics where the analytics are already applied.

Also, adding the `sv` query parameter to the balances endpoint as a workaround for the deprecated `x-sdk-version` header parameter, keeping it backward compatible.

## How Has This Been Tested?

The integration test for the balance endpoint was updated to check the `sv` query parameter and backward compatibility with the `x-sdk-version` header.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/balance.test.ts
+++ b/integration/balance.test.ts
@@ -14,12 +14,7 @@ describe('Account balance', () => {
 
   it('fulfilled balance Ethereum address', async () => {
     let resp: any = await httpClient.get(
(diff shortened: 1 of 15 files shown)


chris13524 (approved):
👌

PR #948, opened 2025-02-10
feat(o11y): adding an alert for providers availability

geekbrother:
# Description

This PR adds an alert for providers (RPC/Non-RPC) availability. The availability is calculated as an `HTTP 200` response percentage and other response codes. The alert will fire if the provider availability is less than 90%.

## How Has This Been Tested?

Manually tested.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #951, opened 2025-02-11
fix: sessionId hardcoded list

chris13524:
# Description

Fixes the crash that the ID was lowercase instead of uppercase.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -3,6 +3,7 @@ use {
     crate::{
         analytics::MessageInfo,
         error::RpcError,
+        providers::ProviderKind,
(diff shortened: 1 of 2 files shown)


nopestack (approved):

nopestack (commented):
maybe also a good idea to lowercase them on this match jic: 

https://github.com/reown-com/blockchain-api/blob/2a32718e030c9d9aaa2c084ece251b64d1fe232d/src/providers/mod.rs#L697

ganchoradkov (approved):

PR #960, opened 2025-02-19
fix: hardcode origin header for registry API

chris13524:
# Description

Resolves # (issue)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -50,7 +50,11 @@ impl Registry {
                 ));
             };
 
-            let client = RegistryHttpClient::new(api_url, api_auth_token)?;
+            let client = RegistryHttpClient::new(
+                api_url,
+                api_auth_token,
+                "https://rpc-service.walletconnect.org",
+            )?;
 
             let cache_addr = cfg_storage.project_data_redis_addr();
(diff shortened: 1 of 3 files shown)


geekbrother (approved):

src/project/mod.rs
```diff
@@ -50,7 +50,11 @@ impl Registry {
                 ));
             };
 
-            let client = RegistryHttpClient::new(api_url, api_auth_token)?;
+            let client = RegistryHttpClient::new(
+                api_url,
```
bkrem, line 54: Wouldn't it be simpler to just use the `api_url` in each of these implementations and strip it down to only the hostname, which is the true `origin` value anyways?

chris13524: That approach would work for setting the `Host` header, but not `Origin`

bkrem: Ok maybe I'm misunderstanding what the actual values for `api_url` would be here

chris13524: > strip it down to only the hostname, which is the true origin value anyways?

Also origin is not just the hostname: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin

PR #961, opened 2025-02-19
feat: self provider

chris13524:
# Description

Creates a "self provider" for alloy and upgrades various usages of this. This new approach avoids doing HTTP requests to the service itself, and instead makes a direct function call to the proxy handler.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -9,7 +9,7 @@ use {
     },
     axum::{
         body::Bytes,
-        extract::{ConnectInfo, MatchedPath, Query, State},
+        extract::{ConnectInfo, Query, State},
         response::Response,
     },
     hyper::{http, HeaderMap},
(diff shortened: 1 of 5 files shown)


src/handlers/proxy.rs
```diff
@@ -49,7 +48,6 @@ async fn handler_internal(
     State(state): State<Arc<AppState>>,
     ConnectInfo(addr): ConnectInfo<SocketAddr>,
     Query(query_params): Query<RpcQueryParams>,
-    _path: MatchedPath,
```
chris13524, line 52: This made implementing the self provider slightly harder as I needed to pass a path. This path was unused anyway, so I opted to remove it.

geekbrother (approved):
LGTM! Thanks a lot 🙏

PR #964, opened 2025-02-20
fix(CA): adding different storage slots per contracts and enabling USDT on Arbitrum

geekbrother:
# Description

This PR adds different contract storage slots support, adding a correct `balanceOf` storage slot for the USDT Arbitrum contract to fix the simulation error and enable the USDT Arbitrum chain abstraction support.

## How Has This Been Tested?

* A new integration test was created to cover such cases.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/assets.rs
+++ b/src/handlers/chain_agnostic/assets.rs
@@ -9,8 +9,8 @@ pub struct AssetMetadata {
 
 /// Asset simulation parameters to override the asset's balance state
 pub struct SimulationParams {
-    /// Asset contract balance storage slot number
-    pub balance_storage_slot: u64,
+    /// Asset contract balance storage slot number per chain
+    pub balance_storage_slots: &'static phf::Map<&'static str, u64>,
     /// Balance override for the asset
     pub balance: u128,
 }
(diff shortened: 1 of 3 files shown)


src/handlers/chain_agnostic/assets.rs
```diff
@@ -9,8 +9,8 @@ pub struct AssetMetadata {
 
 /// Asset simulation parameters to override the asset's balance state
 pub struct SimulationParams {
-    /// Asset contract balance storage slot number
-    pub balance_storage_slot: u64,
+    /// Asset contract balance storage slot number per chain
+    pub balance_storage_slots: &'static phf::Map<&'static str, u64>,
```
geekbrother, line 13: Make storage slots a map, because they can be different (the difference is not only per chain, but should be per contract).

src/handlers/chain_agnostic/assets.rs
```diff
@@ -55,7 +59,11 @@ pub static BRIDGING_ASSETS: phf::Map<&'static str, AssetEntry> = phf_map! {
             decimals: 6,
         },
         simulation: SimulationParams {
-            balance_storage_slot: 0,
+            // Must be in sync with the `USDT_CONTRACTS` from above
+            balance_storage_slots: &phf_map! {
+                "eip155:10" => 0u64,
```
chris13524, line 64: Are we missing Base here?

```suggestion
                "eip155:8453" => 0u64,
```

chris13524 (approved):
I think missing Base but otherwise LGTM! 🙏

src/handlers/chain_agnostic/assets.rs
geekbrother: unfortunately, Bungee supports only USDC on Base

Umar93azizi (commented):
Let's see

Umar93azizi (commented):
Let's see

PR #975, opened 2025-02-27
feat(Tenderly): properly handling of the native currency token standard

geekbrother:
# Description

This PR fixes the `NativeCurrency` token standard name in the Tenderly simulation result. The Tenderly response for this type is `NativeCurrency` and not in a screaming case like `ERC20`. Adding an exception for this type to avoid an internal server error. Also, the contract address is optional for this type.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/mod.rs
+++ b/src/handlers/chain_agnostic/mod.rs
@@ -298,10 +298,14 @@ pub async fn get_assets_changes_from_simulation(
         if asset_changed.asset_type.clone() == AssetChangeType::Transfer
             && asset_changed.token_info.standard.clone() == TokenStandard::Erc20
             && asset_changed.to.is_some()
+            && asset_changed.token_info.contract_address.is_some()
         {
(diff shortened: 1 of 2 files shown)


src/handlers/chain_agnostic/mod.rs
```diff
@@ -298,10 +298,14 @@ pub async fn get_assets_changes_from_simulation(
...
             && asset_changed.to.is_some()
+            && asset_changed.token_info.contract_address.is_some()
         {
             asset_changes.push(Erc20AssetChange {
                 chain_id: transaction.chain_id.clone(),
-                asset_contract: asset_changed.token_info.contract_address,
+                asset_contract: asset_changed
+                    .token_info
+                    .contract_address
+                    .unwrap_or_default(),
```
chris13524, line 308: We shouldn't use `unwrap_or_default()` here as it suppresses errors. Can we use `if let Some(..)`, or else offensive programming techniques?

PR #976, opened 2025-02-28
feat(CA): improving error responses

geekbrother:
# Description

This PR improves error responses with the `reason` field containing the detailed error messages for any of the CA error responses.

## How Has This Been Tested?

Not tested.

## Due diligence before merging

* [x] Update Yttrium to the rev after merging https://github.com/reown-com/yttrium/pull/147

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -320,10 +320,10 @@ async fn handler_internal(
         state.metrics.add_ca_insufficient_funds();
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::InsufficientFunds,
-            reason: Some(format!(
+            reason: format!(
                 "No supported assets with at least {} amount were found in the address {}",
                 erc20_topup_value, from_address
-            )),
+            ),
         }))
         .into_response());
     };
@@ -358,14 +358,14 @@ async fn handler_internal(
             ));
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::NoRoutesAvailable,
-            reason: Some(format!(
+            reason: format!(
                 "No routes were found from {}:{} to {}:{} for an initial amount {}",
                 bridge_chain_id.clone(),
                 bridge_contract,
                 initial_tx_chain_id.clone(),
                 asset_transfer_contract,
                 erc20_topup_value
-            )),
+            ),
         }))
         .into_response());
     };
@@ -391,7 +391,7 @@ async fn handler_internal(
         state.metrics.add_ca_insufficient_funds();
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::InsufficientFunds,
-            reason: Some(error_reason),
+            reason: error_reason,
         }))
         .into_response());
     }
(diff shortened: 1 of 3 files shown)


src/handlers/chain_agnostic/route.rs
```diff
@@ -323,6 +320,10 @@ async fn handler_internal(
         state.metrics.add_ca_insufficient_funds();
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::InsufficientFunds,
+            reason: Some(format!(
+                "No supported assets with at least {} amount were found in the address {}",
```
geekbrother, line 324: No amount in the wallet.

src/handlers/chain_agnostic/route.rs
```diff
@@ -352,11 +353,19 @@ async fn handler_internal(
...
                 bridge_chain_id.clone(),
                 bridge_contract.to_string(),
-                request_payload.transaction.chain_id.clone(),
+                initial_tx_chain_id.clone(),
                 asset_transfer_contract.to_string(),
             ));
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::NoRoutesAvailable,
+            reason: Some(format!(
+                "No routes were found from {}:{} to {}:{} for an initial amount {}",
```
geekbrother, line 362: No routes found.

src/handlers/chain_agnostic/route.rs
```diff
@@ -374,13 +383,15 @@ async fn handler_internal(
         / U256::from(100))
         + required_topup_amount;
     if current_bridging_asset_balance < required_topup_amount {
-        error!(
-            "The current bridging asset balance on {} is {} less than the required topup amount:{}. The bridging fee is:{}",
-            from_address, current_bridging_asset_balance, required_topup_amount, bridging_fee
+        let error_reason = format!(
+            "The current bridging asset balance on {} is {} less than the required topup amount:{}",
```
geekbrother, line 387: Insufficient funds.

src/handlers/chain_agnostic/route.rs
```diff
@@ -405,11 +416,19 @@ async fn handler_internal(
...
                 bridge_chain_id.clone(),
                 bridge_contract.to_string(),
-                request_payload.transaction.chain_id.clone(),
+                initial_tx_chain_id.clone(),
                 asset_transfer_contract.to_string(),
             ));
         return Ok(Json(PrepareResponse::Error(PrepareResponseError {
             error: BridgingError::NoRoutesAvailable,
+            reason: Some(format!(
+                "No routes were found from {}:{} to {}:{} for updated (fee included) amount: {}",
```
geekbrother, line 425: No routes found on the second pass (with fees included).

chris13524 (commented):
Good start but these errors are not user-friendly enough IMO. These errors will be shown directly to users and should clearly explain what the problem is and how the user can resolve the error.

Amounts should be formatted

chris13524 (approved):

src/handlers/chain_agnostic/route.rs
jakubuid: Not sure how it works but will `{}` be replaced with the real values?

jakubuid (approved):

src/handlers/chain_agnostic/route.rs
jakubuid: How would the final message look with the proper values?

geekbrother: > Not sure how it works but will `{}` be replaced with the real values?

Yes, that's correct.

> How would the final message look with the proper values?

It will look like: `"No supported assets with at least 105000 amount were found in the address 0xeeeeeeeeeee"`

PR #981, opened 2025-03-06
feat(CA): adding USDS asset support

geekbrother:
# Description

This PR adds USDS (DAI) asset support for the chain abstraction on Optimism and Arbitrum chains.
Since the bridging fee for DAI is very low, increasing the bridging fee slippage from 75% to 200% will avoid bridging transactions failing in case of fee fluctuation.

## How Has This Been Tested?

* A new integration test was created.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/mod.rs
+++ b/src/handlers/chain_agnostic/mod.rs
@@ -13,7 +13,7 @@ use {
     assets::{SimulationParams, BRIDGING_ASSETS},
     ethers::{types::H160 as EthersH160, utils::keccak256},
     serde::{Deserialize, Serialize},
-    std::{collections::HashMap, str::FromStr, sync::Arc},
+    std::{cmp::Ordering, collections::HashMap, str::FromStr, sync::Arc},
     tracing::debug,
     yttrium::chain_abstraction::api::Transaction,
 };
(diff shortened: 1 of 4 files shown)


src/handlers/chain_agnostic/mod.rs
```diff
@@ -23,7 +23,7 @@ pub mod route;
 pub mod status;
 
 /// How much to multiply the bridging fee amount to cover bridging fee volatility
-pub const BRIDGING_FEE_SLIPPAGE: i8 = 75; // 75%
+pub const BRIDGING_FEE_SLIPPAGE: i16 = 200; // 200%
```
geekbrother, line 26: Since the bridging fee for DAI is very low, increasing the slippage to avoid bridging transaction failing in case of the fee fluctuation.

nopestack (approved):

src/handlers/chain_agnostic/mod.rs
```diff
@@ -314,3 +318,49 @@ pub async fn get_assets_changes_from_simulation(
...
+/// Convert the amount between different decimals
+pub fn convert_amount(amount: U256, from_decimals: u8, to_decimals: u8) -> U256 {
+    match from_decimals.cmp(&to_decimals) {
+        Ordering::Equal => amount,
+        Ordering::Greater => {
+            // Reducing decimals: divide by 10^(from_decimals - to_decimals)
+            let diff = from_decimals - to_decimals;
+            let exp = U256::from(diff as u64);
+            let factor = U256::from(10).pow(exp);
+            amount / factor
```
chris13524, line 331: The result here will be truncated causing the result to be less than the actual value which is bad for comparisons. The smallest-decimals number should be increased to match the largest one, so that precision isn't lost.

PR #989, opened 2025-03-11
feat(Providers): adding Berachain Mainnet support

geekbrother:
# Description

This PR adds Berachain Mainnet `eip155:80094` RPC support by adding the Berachain native mainnet RPC endpoint.

## How Has This Been Tested?

The integration test for the Berachain provider was updated and passed.

## Due Diligence

* [ ] Breaking change


arein (approved):
Should we also configure for publicnode?

chris13524 (approved):

geekbrother:
> Should we also configure for publicnode?

I'll check for all providers that we have to add it where possible as a follow-up.

PR #993, opened 2025-03-17
fix(CA): proper nonce calculation for the same chain swaps

geekbrother:
# Description

This PR fixes the nonce calculation for the initial transaction for cases where there is only a swap on the same chain performing.

Also, handling the negative bridging fee on USDs swaps considering it as a zero fee.

## How Has This Been Tested?

The integration test was updated to check this case.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
Clean fix, thanks!

PR #994, opened 2025-03-18
fix: allow simple requests with JSON

chris13524:
# Description

Adds a `SimpleRequestJson` extractor which is the same as the normal `Json` extractor except that it's compatible with any Content-Type. It does this by modifying the request to have the `application/json` content type and then passing it to the normal `Json` extractor.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
LGTM, Thanks a lot ❤️

PR #996, opened 2025-03-19
fix(Prometheus): adding a check that the Prometheus server is ready in a constructor

geekbrother:
# Description

This PR adds a check that the Prometheus server is ready in a constructor for better error messaging.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2187,6 +2187,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bip39"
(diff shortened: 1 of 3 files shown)


Cargo.lock
geekbrother, line 1: Looks like we forgot to update the `Cargo.lock` from the last commit.

nopestack (approved):

PR #998, opened 2025-03-19
feat: pay with exchange

lukaisailovic:
# Description

Adding the ability to fetch exchanges and then return pay urls for each exchange.

Tests are missing since we're not final on the implementation, but I think this part it good enough to merge.

Note: 
Added openssl crate in order to match RSA signatures in the format Binance requested. Tried with the `RSA` and `Sha2` crates first, but couldn't get the signature to match. This requires the system to have `openssl` support which I think is fine since IRN already has this crate installed. Sometimes the env is bugging out for me locally and I have to reload zsh config.

Resolves # (issue)

## Due Diligence

* [ ] Breaking change


src/handlers/wallet/handler.rs
```diff
@@ -155,6 +165,8 @@ impl Error {
             Error::MethodNotFound => -32601,
             Error::InvalidParams(_) => -32602,
             Error::Internal(_) => -32000,
+            Error::GetExchanges(_) => -32603,
+            Error::GetUrl(_) => -32604,
```
chris13524, line 169: As per the [JSON-RPC spec](https://web.archive.org/web/20250201021958/https://www.jsonrpc.org/specification#error_object) "The error codes from and including -32768 to -32000 are reserved"

These should be -5 and -6 similar to the method specific codes at the top of this match statement

src/handlers/wallet/handler.rs
```diff
@@ -226,6 +238,40 @@ async fn handle_rpc(
...
+        PAY_GET_EXCHANGES => serde_json::to_value(
+            &get_exchanges::handler(
+                state,
+                connect_info,
+                headers,
+                Query(get_exchanges::QueryParams {
+                    sdk_info: query.sdk_info,
+                }),
+                Json(
+                    serde_json::from_value::<get_exchanges::GetExchangesRequest>(params)
```
chris13524, line 250: The type here should be inferable

```suggestion
                    serde_json::from_value(params)
```

src/handlers/wallet/handler.rs
```diff
@@ -226,6 +238,40 @@ async fn handle_rpc(
...
+        PAY_GET_EXCHANGE_URL => serde_json::to_value(
+            &get_exchange_url::handler(
+                state,
+                connect_info,
+                headers,
+                Query(get_exchange_url::QueryParams {
+                    sdk_info: query.sdk_info,
+                }),
+                Json(
+                    serde_json::from_value::<get_exchange_url::GeneratePayUrlRequest>(params)
```
chris13524, line 267: Same here

```suggestion
                    serde_json::from_value(params)
```

src/handlers/wallet/handler.rs
lukaisailovic: Other codes have a todo, do you think I should add -5 and -6 or something unique but out of range of the specified errors?

src/handlers/wallet/exchanges/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+            id: self.id().to_string(),
+            name: self.name().to_string(),
+            image_url: self.image_url().map(|s| s.to_string()),
+        }
+    }
+}
+
+// Define static instances for each exchange to avoid repeated instantiation
+const BINANCE: BinanceExchange = BinanceExchange;
+const COINBASE: CoinbaseExchange = CoinbaseExchange;
```
chris13524, line 36: I wouldn't do this. Initialization is zero-cost in this case anyway since struct is empty. If we want to hold onto something between requests it should go inside `State` but we don't need that here (currently).

src/handlers/wallet/get_exchanges.rs
```diff
@@ -0,0 +1,96 @@
...
+    let mut exchanges = get_supported_exchanges();
+
+    match (&request.include_only, &request.exclude) {
+        (Some(_), Some(_)) => {
+            return Err(GetExchangesError::ValidationError(
+                "includeOnly and exclude are mutually exclusive".to_string(),
+            ));
+        }
+        (Some(include_only), None) => {
+            exchanges.retain(|exchange| include_only.contains(&exchange.id));
```
chris13524, line 84: I would use `filter` here instead of mutating `exchanges`

src/handlers/wallet/get_exchanges.rs
```diff
@@ -0,0 +1,96 @@
...
+    serde::{Deserialize, Serialize},
+    std::{net::SocketAddr, sync::Arc},
+    thiserror::Error,
+    wc::future::FutureExt,
+};
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetExchangesRequest {
+    pub page: u32,
```
chris13524, line 21: Use `usize` so you don't need to truncate the vec length into the u32 later

```suggestion
    pub page: usize,
```

src/handlers/wallet/get_exchanges.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub include_only: Option<Vec<String>>,
+    pub exclude: Option<Vec<String>>,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetExchangesResponse {
+    pub total: u32,
```
chris13524, line 29: Use `usize` so you don't need to truncate the vec length into the u32 later

```suggestion
    pub total: usize,
```

src/handlers/wallet/handler.rs
chris13524: Use -5 and -6. The TODO is to add specific error codes for all the error variants inside of each of those, vs just the 1 error code for all. Maybe can remove the TODO even. The full error should be in the `message` field already.

src/handlers/wallet/exchanges/mod.rs
chris13524: It will become even more important to not do this when `get_buy_url()` makes HTTP requests. The HTTP client will need to be re-used from the global HTTP client, and that can't be done statically.

lukaisailovic: Fixed!

src/handlers/wallet/get_exchange_url.rs
```diff
@@ -0,0 +1,140 @@
...
+    tracing::debug,
+    wc::future::FutureExt,
+};
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GeneratePayUrlRequest {
+    pub exchange_id: String,
+    pub asset: String,
+    pub amount: String,
```
chris13524, line 25: I'd use `U256` for for easy hex parsing numbers and bounds checking. Also `Bytes` when no bounds required.

```suggestion
    pub amount: alloy::primitives::U256,
```

PR #1000, opened 2025-03-20
chore: add Caip19Asset and Caip2ChainId

lukaisailovic:
# Description

Adding `Caip2ChainId` and `Caip19Asset`

Note: Didn't change the current implementation of `ChainId` and its usage

Introduced a new regex that matches the CAIP2 spec exactly, which is not the case with `CAIP_CHAIN_ID_REGEX`

Resolves # (issue)

## How Has This Been Tested?

Automated test coverage

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/utils/crypto.rs
+++ b/src/utils/crypto.rs
@@ -807,7 +807,7 @@ pub enum CaipNamespaces {
 
 /// A struct representing a CAIP-2 Chain ID with format:
 /// `{namespace}:{reference}`
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Caip2ChainId {
     namespace: String,
     reference: String,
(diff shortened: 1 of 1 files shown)


geekbrother:
@lukaisailovic Curious, about where `Caip19Asset` and `Caip2ChainId` will be used? Since there is no usage in this PR.

lukaisailovic:
@geekbrother I need it for #998. Decided to do it in separate PR so its easier to review and handle on its own

src/utils/crypto.rs
```diff
@@ -779,6 +805,207 @@ pub enum CaipNamespaces {
...
+            asset_namespace: asset_namespace.to_string(),
+            asset_reference: asset_reference.to_string(),
+            token_id: token_id.map(ToString::to_string),
+        })
+    }
+
+    /// Parse a CAIP-19 asset ID string
+    pub fn parse(asset_id: &str) -> Result<Self, CryptoUitlsError> {
+        let parts: Vec<&str> = asset_id.splitn(2, '/').collect();
+        if parts.len() != 2 {
```
chris13524, line 934: This makes heavy use of manual length checks and indexing with magic numbers which could panic if implemented incorrectly.

To guarantee no panics or subtle bugs I prefer to structure my code to avoid the chance of panics (from wrong indexing) entirely. For example this is how I would parse a CAIP-10:

```
let caip10_parts = a.splitn(3, ":");
let namespace = caip10_parts.next().ok_or(RpcError::InvalidParameter(
    "The account is not a valid CAIP-10 address: missing namespace".to_string(),
))?;
let reference = caip10_parts.next().ok_or(RpcError::InvalidParameter(
    "The account is not a valid CAIP-10 address: missing reference".to_string(),
))?;
let address = caip10_parts.next().ok_or(RpcError::InvalidParameter(
    "The account is not a valid CAIP-10 address: missing address".to_string(),
))?;
```

chris13524 (approved):

src/utils/crypto.rs
lukaisailovic: @chris13524 Very nice suggestion! I like the approach.

Updated both CAIP-2 and CAIP-19 implementations. 


LMK if I can merge this!

chris13524: LGTM!

PR #1004, opened 2025-03-25
fix(prometheus): making prometheus url optional instead of one time health check

geekbrother:
# Description

This PR makes the Prometheus URL config parameter optional and disables querying Prometheus if it's not provided, instead of the one-time health check during the bootstrap.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -218,31 +218,20 @@ pub struct ProviderRepository {
 
 impl ProviderRepository {
     #[allow(clippy::new_without_default)]
-    pub async fn new(config: &ProvidersConfig) -> Self {
-        let prometheus_client = {
-            let prometheus_query_url = config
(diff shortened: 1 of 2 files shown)


src/providers/mod.rs
```diff
@@ -623,30 +612,33 @@ impl ProviderRepository {
...
         debug!("Updating weights");
 
+        let Some(prometheus_client) = &self.prometheus_client else {
+            debug!("Prometheus client not configured, skipping weight update");
+            return;
+        };
+
         let Ok(header_value) = HeaderValue::from_str(&self.prometheus_workspace_header) else {
-            warn!(
+            error!(
```
geekbrother, line 621: It must be an error since if the Prometheus URL was passed and we can't update, something is wrong.

nopestack (approved):

PR #1017, opened 2025-04-07
feat: middlewares reordering, writing metrics async

geekbrother:
# Description

This PR reorders middlewares to execute **all middlewares** except `latency and status codes` middleware **only on existing routes** by using [route_layer](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.route_layer) instead of [layer](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.layer) on all routes (including not existing).
Also, changes to write metrics in the `latency and status codes` middleware async instead of sync which can lead to blocking or latency increase for responses.

## How Has This Been Tested?

Tested locally.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -157,14 +157,17 @@ pub async fn status_latency_metrics_middleware<B>(
     let response = next.run(req).await;
     let request_latency = request_started.elapsed();
 
-    // Record metrics
-    state
-        .metrics
-        .add_http_call(response.status().as_u16(), uri.clone());
-    state.metrics.add_http_latency(
-        response.status().as_u16(),
-        uri,
-        request_latency.as_secs_f64(),
-    );
+    // Record metrics async
+    let state_clone = state.clone();
+    let uri_clone = uri.clone();
+    let status = response.status().as_u16();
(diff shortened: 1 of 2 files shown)


nopestack (approved):

src/handlers/mod.rs
```diff
@@ -157,14 +157,17 @@ pub async fn status_latency_metrics_middleware<B>(
...
-        response.status().as_u16(),
-        uri,
-        request_latency.as_secs_f64(),
-    );
+    // Record metrics async
+    let state_clone = state.clone();
+    let uri_clone = uri.clone();
+    let status = response.status().as_u16();
+    let latency_secs = request_latency.as_secs_f64();
+    tokio::spawn(async move {
```
chris13524, line 165: There are no `await` points in this task so running it as a task doesn't really do much. I think metrics recording is very very quick anyway, so doesn't need to be done async.

chris13524 (approved):
Switch to `route_layer` makes sense to me. No need to run all this validation for non-existent routes.

But I don't think that the async metric recording is better than how it was before.

PR #1019, opened 2025-04-07
fix(metrics): using the matched path instead of the URI in status and latency metrics

geekbrother:
# Description

This PR fixes a metrics overflow bug when using the URI as a latency and status metrics key that results into a metrics overflow due to use of the `/route/:address` routes. The fix is to use the matched path extractor instead.

## How Has This Been Tested?

Tested locally: a middleware using the matched path instead of the URI.

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
LGTM

PR #1022, opened 2025-04-08
fix: prepare use op

chris13524:
# Description

Fixes the prepare flow to use `eth_estimateUserOperationGas` as well as fully support the new 7677 paymaster spec. The full flow was not tested, but this is better than previously.

[Slack conversation](https://reown-inc.slack.com/archives/C06G97QSSUX/p1744113153391959?thread_ts=1744017462.368659&cid=C06G97QSSUX)

## How Has This Been Tested?

Somewhat manually

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/prepare_calls.rs
+++ b/src/handlers/wallet/prepare_calls.rs
@@ -4,6 +4,8 @@ use crate::handlers::sessions::get::{
     get_session_context, GetSessionContextError, InternalGetSessionContextError,
 };
 use crate::handlers::wallet::types::SignatureRequestType;
+use crate::utils::erc4337::BundlerRpcClient;
+use crate::utils::erc7677::{PaymasterRpcClient, PmGetPaymasterDataParams};
 use crate::{handlers::HANDLER_TASK_METRICS, state::AppState};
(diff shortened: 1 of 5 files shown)


lukaisailovic (approved):

src/handlers/wallet/prepare_calls.rs
```diff
@@ -252,15 +260,15 @@ async fn handler_internal(
...
-            .parse()
-            .unwrap(),
-        ));
+        let bundler_url = format!(
+            "https://rpc.walletconnect.com/v1/bundler?chainId={}&projectId={}&bundler=pimlico",
+            chain_id.caip2_identifier(),
+            project_id,
+        )
+        .parse::<Url>()
+        .unwrap();
```
geekbrother, line 269: Can we use `expect` here or handle this?

geekbrother (approved):
LGTM with a comment 👍

src/handlers/wallet/prepare_calls.rs
chris13524: Switched to expect

PR #1025, opened 2025-04-08
feat: LiFi feature flag

chris13524:
# Description

Adds a query param allowing clients to opt-in to using LiFi for EVM bridging transactions.

Resolves WK-637
Resolves WK-617

Depends on https://github.com/reown-com/yttrium/pull/160

## How Has This Been Tested?

New tests (wip)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -879,33 +879,6 @@ async fn handler_internal(
             )
         }
         bridge_contract => {
-            // let mut solana_accounts = Vec::with_capacity(request_payload.accounts.len());
-            // for (chain_id, account) in request_payload
-            //     .accounts
-            //     .iter()
-            //     .filter_map(|a| a.strip_prefix("solana:"))
-            //     .filter_map(|a| a.split_once(":"))
-            // // skip any malformed CAIP-10 addresses without 3 colons
-            // {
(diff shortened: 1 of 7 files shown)


src/handlers/chain_agnostic/route.rs
```diff
@@ -857,51 +856,43 @@ async fn handler_internal(
...
+
+            // let mainnet_solana_accounts = solana_accounts
+            //     .iter()
+            //     .filter(|(chain_id, _)| chain_id == &solana_mainnet)
+            //     .map(|(_, account)| account)
+            //     .collect::<Vec<_>>();
+
+            // let _found_account = mainnet_solana_accounts
+            //     .iter()
+            //     .find(|a| bridging_asset.account.as_solana() == Some(a))
```
chris13524, line 884: I'm not sure why I wrote all this code. I have `bridging_asset.account` right here which should be the same as `found_account`?

nopestack: why keep it?

chris13524: Might remove

geekbrother: Looks like it's duplicating the logic. If tests pass after removing let's remove it.

geekbrother (approved):

PR #1031, opened 2025-04-10
feat(CA): enabling ERC20 transfer gas estimation caching

geekbrother:
# Description

This PR enables the gas estimation caching for the ERC20 transfer CA initial transaction. The cache TTL decreased to 30 minutes, and the gas estimation slippage is x5 to cover the volatility.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -339,9 +339,7 @@ impl ProviderRepository {
             config.tenderly_api_key.clone(),
             config.tenderly_account_id.clone(),
             config.tenderly_project_id.clone(),
-            // Todo: Temporary disabling the gas estimation caching
-            // redis_pool.clone(),
-            None,
+            redis_pool.clone(),
         ));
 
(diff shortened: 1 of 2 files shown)


nopestack (approved):

src/providers/mod.rs
```diff
@@ -339,9 +339,7 @@ impl ProviderRepository {
             config.tenderly_api_key.clone(),
             config.tenderly_account_id.clone(),
             config.tenderly_project_id.clone(),
-            // Todo: Temporary disabling the gas estimation caching
-            // redis_pool.clone(),
-            None,
+            redis_pool.clone(),
```
nopestack, line 342: why was it disabled in the first place?

geekbrother: We want to test without caching first to check for a proper TTL and slippage to make sure it's not a caching issue.

PR #1034, opened 2025-04-11
fix: route performance & refactor

chris13524:
# Description

- Optimize nonce fetching to be done in-parallel to other operations, improving performance
- Refactor initial transaction handling to defer construction to as late as possible, keeping mutations to a minimum (only `nonce` is mutable, not the entire `initial_transaction`)
- Refactor `get_assets_changes_from_simulation()` to avoid coupling on `Transaction` type
- Improve overall handler clarity by removing multiple local variables which were just copies of other variables or inputs

## How Has This Been Tested?

Existing test suite & manually with Yttrium tests

## Due Diligence

* [ ] Breaking change


src/handlers/chain_agnostic/route.rs
```diff
@@ -727,15 +748,10 @@ async fn handler_internal(
...
-            let mut current_nonce = get_nonce(
-                from_address,
-                &provider_pool.get_provider(
-                    format!("eip155:{}", bridge_tx.chain_id),
-                    MessageSource::ChainAgnosticCheck,
-                ),
-            )
-            .await?;
+            let mut bridge_account_nonce = bridge_account_nonce
+                .expect("bridge_account_nonce should've been initiated")
```
geekbrother, line 752: Would be good to handle this without expect?

geekbrother (approved):
LGTM, with a minor comment. Thanks!

src/handlers/chain_agnostic/route.rs
chris13524: No because this is an invariant that should always hold true

chris13524: Alternatively I suppose we could introduce a new class of error variants e.g. `RouteHandlerError::Internal(InternalRouteHandlerError::Invariant(FatalRouteHandlerError::BridgeAccountNonce)`. This would allow us to classify these types of errors specifically as invariants and they would be handled somewhat more gracefully.

However I'm not sure in the case of Blockchain API this would be any better than just panicking? My understanding of `axum` is that it will catch panics and classify them as a 500 anyway without affecting the other requests the server is processing. This should be captured by all of our current logging and alerting. Reading what is idiomatic use of panics in Rust is programmer error, vs things you expect to happen at some point e.g. network error. Perhaps considering lock poisoning, it's a "better safe than sorry" situation? How far do we go? For example replacing `arr[i]` with `arr.get(i)`?

I could perhaps see an argument for FFI/WASM/embedded contexts where you do panic=abort for binary size reasons which causes your debuggability of the panic to become very difficult.

Curious where and why do we draw this line?

PR #1044, opened 2025-04-24
fix: some internal wallet RPC errors were not being logged

chris13524:
# Description

Generalizes what #813 tried to do for all errors.

This introduces an OOP-based way to determine if an error is internal or not, self-reported by the error itself.

[Slack conversation](https://reown-inc.slack.com/archives/C03SCF66K2T/p1745484302400629?thread_ts=1744378976.546939&cid=C03SCF66K2T)

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/get_assets.rs
+++ b/src/handlers/wallet/get_assets.rs
@@ -36,21 +36,11 @@ pub enum GetAssetsErrorInternalError {
     GetBalance(RpcError),
 }
 
-impl IntoResponse for GetAssetsError {
-    fn into_response(self) -> Response {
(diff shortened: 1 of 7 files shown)


src/handlers/wallet/get_assets.rs
```diff
@@ -36,21 +36,11 @@ pub enum GetAssetsErrorInternalError {
     GetBalance(RpcError),
 }
 
-impl IntoResponse for GetAssetsError {
```
chris13524, line 39: I think this was some old code. Seems to work fine without.

lukaisailovic (approved):

PR #1053, opened 2025-05-01
fix: nonce manager

chris13524:
# Description

Fix failing nonce-related check: https://github.com/reown-com/blockchain-api/actions/runs/14777402543/job/41496849695#step:5:162

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/chain_agnostic/route.rs
+++ b/src/handlers/chain_agnostic/route.rs
@@ -2,8 +2,8 @@ use {
     super::{
         super::HANDLER_TASK_METRICS, assets::NATIVE_TOKEN_ADDRESS,
         check_bridging_for_erc20_transfer, convert_amount, find_supported_bridging_asset,
-        get_assets_changes_from_simulation, BridgingStatus, StorageBridgingItem,
-        BRIDGING_FEE_SLIPPAGE, STATUS_POLLING_INTERVAL,
+        get_assets_changes_from_simulation, nonce_manager::NonceManager, BridgingStatus,
+        StorageBridgingItem, BRIDGING_FEE_SLIPPAGE, STATUS_POLLING_INTERVAL,
     },
     crate::{
         analytics::{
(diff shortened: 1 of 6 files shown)


src/handlers/chain_agnostic/route.rs
```diff
@@ -439,13 +439,11 @@ async fn handler_internal(
...
 
-    let from_account_nonce = tokio::task::spawn({
-        let provider = provider_pool.get_provider(
-            request_payload.transaction.chain_id.clone(),
-            MessageSource::ChainAgnosticCheck,
-        );
-        async move { get_nonce(request_payload.transaction.from, &provider).await }
-    });
+    let mut nonce_manager = NonceManager::new(provider_pool.clone());
+    nonce_manager.initialize_nonce(
```
geekbrother, line 443: What is the reason for introducing two steps for a nonce here: initialize, get_nonce? 
Looks like we are always and everywhere waiting for a `.get_nonce.await?` to execute the future, so it's a sync operation in total.
I believe the initialization step can be included in the `get_nonce`, and we can simplify this.

geekbrother (approved):
It fixes the nonce issue. Thanks!

src/handlers/chain_agnostic/route.rs
chris13524: It's not a sync operation because the nonce fetching is spawned in a separate task [here](https://github.com/reown-com/blockchain-api/pull/1053/files#diff-a35adec44262faf0b989e1715c22ad3c81bbf45a21eecd778fd0c519fb2c175bR38).

By separating these two actions it forces you to reason about when to fetch the nonce, in order to do it in-parallel. If you don't properly implement the parallelization it will alert you to the error later.

src/handlers/chain_agnostic/route.rs
```diff
@@ -598,17 +601,7 @@ async fn handler_internal(
...
-            let provider = provider_pool
-                .get_provider(bridge_chain_id.clone(), MessageSource::ChainAgnosticCheck);
-            async move { get_nonce(request_payload.transaction.from, &provider).await }
-        }))
-    } else {
-        None
-    };
-
-    let mut from_account_nonce = from_account_nonce.await??;
+    nonce_manager.initialize_nonce(bridge_chain_id.clone(), request_payload.transaction.from);
```
chris13524, line 604: I think there might be an error here. We shouldn't initialize this if it's not an eip155 chain. I think this might give a runtime error?

geekbrother: Agree, we should handle an error here.

chris13524: Was fixed in [bd735d9](https://github.com/reown-com/blockchain-api/pull/1053/commits/bd735d9ec813d3cf3de5df24e7f8acce036a96cc)

lukaisailovic (approved):

PR #1056, opened 2025-05-06
fix(transactions): adding a filtering by `chainId`

geekbrother:
# Description

This PR adds a filtering by `chainId` by using an optional `chainId` parameter in the transactions history endpoint.

## How Has This Been Tested?

The new integration test was created.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/zerion.rs
+++ b/src/providers/zerion.rs
@@ -263,6 +263,18 @@ impl HistoryProvider for ZerionProvider {
             url.query_pairs_mut().append_pair("page[after]", &cursor);
         }
 
+        if let Some(chain_id) = params.chain_id {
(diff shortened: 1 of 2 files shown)


src/providers/zerion.rs
```diff
@@ -263,6 +263,18 @@ impl HistoryProvider for ZerionProvider {
             url.query_pairs_mut().append_pair("page[after]", &cursor);
         }
 
+        if let Some(chain_id) = params.chain_id {
+            let chain_name = if chain_id.contains(':') {
```
chris13524, line 267: This `if` statement seems redundant since `from_caip2` already returns a result type?

chris13524 (approved):
LGTM

PR #1057, opened 2025-05-07
chore: pay merchant name

lukaisailovic:
# Description
Fetch project name and provide it to Binance as a merchant name. If unavailable, use empty string
Screenshot 2025-05-07 at 09 29 49

Resolves # (issue)

## Due Diligence

* [ ] Breaking change


src/handlers/wallet/exchanges/binance.rs
```diff
@@ -19,6 +19,7 @@ pub struct BinanceExchange;
 
 const PRE_ORDER_PATH: &str = "/papi/v1/ramp/connect/buy/pre-order";
 const QUERY_ORDER_DETAILS_PATH: &str = "/papi/v1/ramp/connect/order";
+const DEFAULT_MERCHANT_NAME: &str = " ";
```
geekbrother, line 22: Do we really need to introduce an empty string const here? It looks like it's used only once, so maybe it makes sense just to construct an empty string where it's used.

geekbrother (approved):
Thanks!

src/handlers/wallet/exchanges/binance.rs
```diff
@@ -392,6 +395,21 @@ impl BinanceExchange {
...
+            .registry
+            .project_data(&params.project_id)
+            .await
+            .map_err(|e| {
+                debug!("Failed to get project data: {}", e);
+                ExchangeError::InternalError(format!("Failed to get project data: {}", e))
+            })?;
+        let project_name = if project.project_data.name.is_empty() {
+            debug!("Project name is empty, using default name");
+            DEFAULT_MERCHANT_NAME.to_string()
```
chris13524, line 408: I would call this "fallback" not "default". I assume this is just used temporary placeholder information in the UI while the registry API is down, not something that would be consumed and/or stored by another system?

chris13524 (approved):

src/handlers/wallet/exchanges/binance.rs
lukaisailovic: Its empty if you run it locally, so I left it. But agree on fallback

src/handlers/wallet/exchanges/binance.rs
lukaisailovic: I think its more readable, personal preference

PR #1058, opened 2025-05-07
fix: coinbase transaction status

lukaisailovic:
# Description

Fix Coinbase transaction status not properly parsed

Resolves # (issue)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/exchanges/coinbase.rs
+++ b/src/handlers/wallet/exchanges/coinbase.rs
@@ -246,14 +246,7 @@ impl CoinbaseExchange {
             )
             .await?;
 
-  
-        
-        let response_text = res.text().await.map_err(|e| {
-            debug!("Error reading response text: {:?}", e);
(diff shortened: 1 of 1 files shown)


geekbrother (approved):
Would love to see a basic schema check integration test [here](https://github.com/reown-com/blockchain-api/tree/master/integration) 🙏 So we can catch such errors in CI/CD.

chris13524 (approved):
LGTM. Assuming all these options really exist independently like this. If these are multiple separate possible response variants or many of these options are associated with one another you should consider separate structs for each variant and put them in a (serde) flattened enum or nested flattened struct.

lukaisailovic:
@chris13524 Unfortunately they do. And its completely different than their docs which is why it was broken initially. I couldn't find exactly when is each option present (I assume they change based on the state) so I just put everything as optional as its not super important to us

arein (approved):

PR #1073, opened 2025-05-16
feat(RPC): adding Solana RPC WebSocket support

geekbrother:
# Description

This PR adds Solana Mainnet and Devnet RPC WebSocket support to the `/ws` endpoint.
Syndica RPC provider is used as a WebSocket Solana RPC provider.

## How Has This Been Tested?

Tested manually by subscribing to the `rootSubscribe` method.

## Due Diligence

* [ ] Breaking change


src/env/syndica.rs
```diff
@@ -55,3 +57,26 @@ fn default_supported_chains() -> HashMap<String, (String, Weight)> {
...
     ])
 }
+
+fn default_ws_supported_chains() -> HashMap<String, (String, Weight)> {
+    // Keep in-sync with SUPPORTED_CHAINS.md
+
+    HashMap::from([
+        // Solana Mainnet
+        (
+            "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp".into(),
```
chris13524, line 67: Maybe we should move all these CAIP-2 IDs into a constants file somewhere? Then we can just reference that instead of having trust and duplicate magic strings everywhere.

chris13524 (approved):
LGTM. Thanks!!

src/env/syndica.rs
geekbrother: Good idea! I believe we can extend the current [ChainId](https://github.com/reown-com/blockchain-api/blob/11c5a20e124e89900f5d4a78b5941dff81626075/src/utils/crypto.rs#L705) with this.

PR #1075, opened 2025-05-16
feat: exchange projectid validation

lukaisailovic:
# Description

Added an allowlist for pay with exchange feature. Current implementation is based on what we already have in the code (balance denylist) and naively checks the projectId.

If we see the need, we can also implement the check for allowed origins in the future.

Resolves PAY-29

## Due Diligence

* [ ] Breaking change


src/handlers/wallet/exchanges/mod.rs
```diff
@@ -157,3 +162,25 @@ pub fn get_supported_exchanges(asset: Option<String>) -> Result<Vec<Exchange>, E
...
+    let allowed_project_ids = state
+        .config
+        .exchanges
+        .allowed_project_ids
+        .as_ref()
+        .ok_or_else(|| ExchangeError::FeatureNotEnabled("Feature is not enabled".to_string()))?;
+
+    debug!("allowed_project_ids: {:?}", allowed_project_ids);
+
+    if !allowed_project_ids.contains(&project_id.to_string()) {
```
chris13524, line 179: https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

Instead of calling `to_string()` here, `project_id` param should be `String` to allow caller to decide when to perform the memory allocation. In-fact, you are already passing it as a `&String` but it's being deref coerced into the `&str`, resulting in a redundant allocation.

chris13524 (approved):

PR #1076, opened 2025-05-16
feat: handle custom bundler URL

chris13524:
# Description

[Design doc](https://www.notion.so/walletconnect/1Pager-WalletKit-Util-for-WalletService-1d83a661771e8093b014e2594f947c4e?pvs=4#1f53a661771e80cc935bd7156166ea19)

Resolves https://linear.app/reown/issue/CR-701/blockchain-api-bundler-endpoint-accept-bundler-url-as-bundler-name

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/bundler.rs
+++ b/src/handlers/bundler.rs
@@ -4,7 +4,10 @@ use {
         error::RpcError,
         providers::SupportedBundlerOps,
         state::AppState,
-        utils::{crypto::disassemble_caip2, simple_request_json::SimpleRequestJson},
+        utils::{
+            crypto::{self, disassemble_caip2},
+            simple_request_json::SimpleRequestJson,
+        },
     },
     alloy::rpc::json_rpc::Id,
     axum::{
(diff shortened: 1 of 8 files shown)


src/handlers/bundler.rs
```diff
@@ -70,6 +74,79 @@ async fn handler_internal(
...
+                .map_err(RpcError::UnsupportedBundlerNameUrlParseError)?;
+            if url.scheme() != "https" {
+                return Err(RpcError::UnsupportedBundlerName(format!(
+                    "must be https://, got {}",
+                    url.scheme()
+                )));
+            }
+            let domain = url.domain();
+            if let Some(domain) = domain {
+                if domain.ends_with("localhost")
```
geekbrother, line 102: Do we really need this `localhosts` tests? Does it make sense to return an error only for this case, or any cases where localhost can be provided?

geekbrother (approved):
Thanks!

src/handlers/bundler.rs
chris13524: https://www.notion.so/walletconnect/1Pager-WalletKit-Util-for-WalletService-1d83a661771e8093b014e2594f947c4e?pvs=4#1e33a661771e804595eaf911291bbb6d

PR #1077, opened 2025-05-20
fix(Meld): proper handling of the bad request response

geekbrother:
# Description

This PR adds handling of the "bad request" Meld provider response, to handle both `HTTP 400` and `HTTP 422`, since the production endpoint uses `HTTP 422` and the sandbox uses `HTTP 400`.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/meld.rs
+++ b/src/providers/meld.rs
@@ -12,6 +12,7 @@ use {
         Metrics,
     },
     async_trait::async_trait,
+    reqwest::StatusCode,
     serde::{Deserialize, Serialize},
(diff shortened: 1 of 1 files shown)


nopestack (approved):

src/providers/meld.rs
```diff
@@ -12,6 +12,7 @@ use {
         Metrics,
     },
     async_trait::async_trait,
+    reqwest::StatusCode,
```
ignaciosantise, line 15: is this a typo? `reqwest`?

PR #1078, opened 2025-05-21
feat(providers): adding Sui chain RPC support

geekbrother:
# Description

This PR adds Sui chain RPC support for the mainnet, testnet and devnet using the following RPC providers:
* Publicnode,
* Grove/Pokt,
* Native RPC endpoint.

## How Has This Been Tested?

New provider tests were implemented.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/http/sui.rs
+++ b/tests/functional/http/sui.rs
@@ -1,0 +1,17 @@
+use {
+    super::check_if_rpc_is_responding_correctly_for_sui, crate::context::ServerContext,
+    rpc_proxy::providers::ProviderKind, test_context::test_context,
+};
+
+#[test_context(ServerContext)]
+#[tokio::test]
+#[ignore]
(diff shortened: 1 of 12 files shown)


tests/functional/http/sui.rs
```diff
@@ -0,0 +1,17 @@
+use {
+    super::check_if_rpc_is_responding_correctly_for_sui, crate::context::ServerContext,
+    rpc_proxy::providers::ProviderKind, test_context::test_context,
+};
+
+#[test_context(ServerContext)]
+#[tokio::test]
+#[ignore]
```
nopestack, line 8: why ignore though?

geekbrother: We run provider tests only when the provider code is changed to speed up the deployment. [This workflow handles this](https://github.com/reown-com/blockchain-api/blob/master/.github/workflows/sub-providers.yml).

PR #1096, opened 2025-06-03
fix: coinbase order status

lukaisailovic:
# Description

Check both if Coinbase returned something and if its an empty string (they also returned that)

Resolves # (issue)

## Due Diligence

* [ ] Breaking change


chris13524 (approved):
I'd use `map` and `unwrap_or` but this is fine :D

lukaisailovic:
@chris13524 Updated the impl. Clippy suggested a better approach so went with it

PR #1107, opened 2025-06-05
fix(balance): using the minimal priority providers only for a failover

geekbrother:
# Description

This PR changes the weighted balance providers selection algorithm to use balance providers with the `Minimal` priority only for a failover.

This should temporarily fix the balance providers' tokens list inconsistency.
The [Slack thread](https://reown-inc.slack.com/archives/C03SCF66K2T/p1748842766848099).

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -457,33 +457,33 @@ impl ProviderRepository {
             return Err(RpcError::UnsupportedChain(namespace.to_string()));
         }
 
-        let weights: Vec<_> = providers
-            .iter()
-            .map(|(_, weight)| weight.value())
-            .map(|w| w.max(1))
-            .collect();
-
         // Adding non-minimal priority providers and use providers with the minimal priority
         // only for a failover retrying (append them to the end of the list)
         let minimal_weight_value = Weight::new(Priority::Minimal)
             .expect("Failed to create a Minimal priority value")
             .value();
-        let non_minimal_weight_providers = weights
-            .iter()
-            .filter(|&x| *x > minimal_weight_value)
-            .count();
-        let keys = providers.keys().cloned().collect::<Vec<_>>();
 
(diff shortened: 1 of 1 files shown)


src/providers/mod.rs
```diff
@@ -501,6 +511,25 @@ impl ProviderRepository {
                             })
                     })
                     .collect::<Result<Vec<_>, _>>()?;
+
+                // Append minimal priority providers to the end of the list
+                // to use it only for a failover retrying
+                if non_minimal_weight_providers < max_providers {
```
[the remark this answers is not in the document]
chris13524: Maybe I'm wrong, but overall seems like this section could be simplified. Multiple pieces of math going on (this comparison and the subtraction inside) but I think we should be able to simplify to 1 subtraction operation and a comparison to zero or `providers_result.len()`

src/providers/mod.rs
```diff
@@ -457,23 +457,58 @@ impl ProviderRepository {
...
+        let keys: Vec<_> = high_priority_providers.iter().map(|(key, _)| key).collect();
+
+        // If no non-minimal providers are available, directly append minimal-priority providers
+        if non_minimal_weight_providers.is_empty() {
+            let minimal_weight_providers = minimal_weight_providers
+                .into_iter()
+                .take(max_providers)
+                .collect::<Vec<_>>();
+            return Ok(minimal_weight_providers);
+        }
```
chris13524, line 500: Why do we need this extra case up here? Shouldn't the minimal-provider filling you you added below cover this?

geekbrother: This special case is if all providers have minimal weights. In this case, no `non_minimal_weight_providers` and no weights distribution will be created below, so no providers will be returned at all.

PR #1121, opened 2025-06-23
feat(Meld): getting quotes for all possible payment types

geekbrother:
# Description

This PR adds support for obtaining Meld quotes for all possible payment types.
By default, the Meld quotes response has quotes only for the bank card payment type if no payment type was provided.
We want to get all available quotes for all available payment types by getting all available payment types for the country, and then querying in parallel the Meld quotes endpoint for each payment method. 
The aggregated response quotes will be returned.

## How Has This Been Tested?

The quotes endpoint integration test was updated to check that the quotes response has multiple payment types.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ yttrium = { git = "https://github.com/reown-com/yttrium.git", rev = "5e1b5f4", f
 
 # Async
 async-trait = "0.1.82"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1.45", features = ["full"] }
 
 # Web
 hyper = "0.14"
--- a/src/providers/meld.rs
+++ b/src/providers/meld.rs
@@ -12,9 +12,11 @@ use {
         Metrics,
     },
     async_trait::async_trait,
+    futures_util::future::join_all,
     reqwest::StatusCode,
     serde::{Deserialize, Serialize},
     std::{sync::Arc, time::SystemTime},
+    tokio::task,
     tracing::log::error,
     url::Url,
 };
@@ -86,6 +88,14 @@ pub struct MeldQuotesResponse {
 pub struct MeldErrorResponse {
     pub code: String,
     pub message: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct PaymentMethod {
+    pub name: String,
+    pub payment_method: String,
+    pub payment_type: String,
 }
 
 #[async_trait]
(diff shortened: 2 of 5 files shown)


src/providers/meld.rs
```diff
@@ -304,55 +314,122 @@ impl OnRampMultiProvider for MeldProvider {
...
-                        // Respond to the client with a generic error message and HTTP 400 anyway
-                        MeldErrorResponse {
-                            code: "BAD_REQUEST".to_string(),
-                            message: "Invalid parameter".to_string(),
-                        }
+        // Get quotes for each payment type in parallel and aggregate results
+        // otherwise only the card payment is provided in quotes if there are no
+        // payment type was provided to the request, but we want to get all
+        // available quotes for all payment types.
+        let mut handles = Vec::new();
```
nopestack, line 346: may be a good idea to cap the amount of handles you're planning to have and also consider   using FuturesUnordered/Ordered or JoinSet as they're meant for this and are more ergonomic

nopestack (changes requested):
please address comments in PR

src/providers/meld.rs
geekbrother: Good catch, thanks a lot! 
It's time to update Tokio and use the `JoinSet` instead 👍

Cargo.toml
```diff
@@ -16,7 +16,7 @@ yttrium = { git = "https://github.com/reown-com/yttrium.git", rev = "5e1b5f4", f
 
 # Async
 async-trait = "0.1.82"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1.45", features = ["full"] }
```
geekbrother, line 19: Stick to the recent version to use the `JoinSet`.

src/providers/meld.rs
```diff
@@ -304,55 +382,76 @@ impl OnRampMultiProvider for MeldProvider {
...
-                "Error on Meld get quotes. Status is not OK: {:?}",
-                response.status(),
-            );
-            return Err(RpcError::OnRampProviderError);
+        // We are not expecting more than 20 payment types, we should stop
+        // if we have more than 20 payment types.
+        if payment_types.len() > 20 {
+            return Err(RpcError::ConversionInvalidParameter(
+                "Too many payment types".to_string(),
+            ));
```
chris13524, line 415: Should this be a 5xx error? Seems like it would happen from an internal API giving too many results?

src/providers/meld.rs
```diff
@@ -304,55 +382,76 @@ impl OnRampMultiProvider for MeldProvider {
...
+        while let Some(result) = join_set.join_next().await {
+            match result {
+                Ok(Ok(quotes_response)) => quotes.extend(quotes_response),
+                Ok(Err(e)) => return Err(e),
+                Err(e) => {
+                    error!("Meld parallel quotes tasks join error: {:?}", e);
+                    return Err(RpcError::OnRampProviderError);
+                }
+            }
+        }
```
chris13524, line 453: Should we consider partial-functionality/response if one of these quotes fails instead of failing the whole request?

src/providers/meld.rs
```diff
@@ -66,6 +67,75 @@ impl MeldProvider {
...
+        api_key: String,
+    ) -> RpcResult<Vec<QuotesResponse>> {
+        params.payment_method_type = Some(payment_type);
+
+        let latency_start = SystemTime::now();
+        let response = http_client
+            .post(url)
+            .json(&params)
+            .header("Meld-Version", API_VERSION)
+            .header("Authorization", format!("BASIC {}", api_key))
```
chris13524, line 87: I think you could use `basic_auth(api_key)` instead

chris13524 (approved):

src/providers/meld.rs
geekbrother: Unfortunately, not 🥲 `basic_auth()` is designed for username/password authentication, not API key authentication, it expects the password also in this case and doesn't work with an empty password provided.

src/providers/meld.rs
geekbrother: Makes sense, since we have an alert for each `HTTP 500` to catch this if it happens. I've changed to use a separate error type that results in an `HTTP 500`. Thank you 🙌

PR #1132, opened 2025-07-01
feat: record metrics for available chain

chris13524:
# Description

It seems we don't have a metric for if a chain is unavailable. We have `no_providers_for_chain_counter` but this only counts chain un-availability but not availability. This enables us to have a proper % availability for each chain.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


geekbrother (approved):
Thanks 🙏

PR #1153, opened 2025-07-17
feat(proxy): adding an error check only when the response is success and debugging message

geekbrother:
# Description

This PR adds a response deserialization check only when the response is successful, because there is no reason to check the schema and the error in case of unsuccessful responses. Adding error message logging for unhandled error codes.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/proxy.rs
+++ b/src/handlers/proxy.rs
@@ -160,7 +160,10 @@ pub async fn rpc_call(
         .await;
 
         match response {
-            Ok(response) if !response.status().is_server_error() => {
+            Ok(response)
+                if (!response.status().is_server_error()
+                    && response.status() != http::StatusCode::TOO_MANY_REQUESTS) =>
+            {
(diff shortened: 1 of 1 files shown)


src/handlers/proxy.rs
```diff
@@ -173,32 +173,37 @@ pub async fn rpc_call(
...
-                                    error.code,
-                                );
-                                state
-                                    .metrics
-                                    .add_rpc_call_retries(i as u64, chain_id.clone());
-                                continue;
+                // Check the JSON-RPC response schema and possible internal error codes
+                // if the status is success and error is present
+                if status.is_success() {
+                    match serde_json::from_slice::<jsonrpc::Response>(&body_bytes) {
```
nopestack, line 179: nit: i'd avoid using a match here to avoid nesting

PR #1159, opened 2025-07-19
chore: fix some minor issues in comments

jishudashen:
# Description

fix some minor issues in comments

Resolves # (issue)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/migrations/20231209213851_init-names.sql
+++ b/migrations/20231209213851_init-names.sql
@@ -10,6 +10,6 @@ CREATE TABLE names (
   -- We are using the hstore as a key-value for the extensible attributes list
   attributes hstore,
 
-  -- Check for the standartized name format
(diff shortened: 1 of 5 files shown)


migrations/20231209213851_init-names.sql
```diff
@@ -10,6 +10,6 @@ CREATE TABLE names (
   -- We are using the hstore as a key-value for the extensible attributes list
   attributes hstore,
 
-  -- Check for the standartized name format
+  -- Check for the standardized name format
```
chris13524, line 13: Cannot change migrations as this would result in a runtime error

chris13524:
@jishudashen thank you for your PR, but at this time we cannot accept these types of contributions for security reasons 🙏

PR #1163, opened 2025-07-21
feat(Meld): adding default providers and exclude providers parameters

geekbrother:
# Description

This PR adds a default providers list for the `/v1/onramp/providers/properties` and `excludeProviders` query parameter to provide a providers exclusion list.

## How Has This Been Tested?

Tested manually.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/providers/meld.rs
+++ b/src/providers/meld.rs
@@ -23,6 +23,21 @@ use {
 const API_VERSION: &str = "2023-12-19";
 const DEFAULT_CATEGORY: &str = "CRYPTO_ONRAMP";
 const DEFAULT_SESSION_TYPE: &str = "BUY";
+const DEFAULT_PROVIDERS_LIST: &[&str] = &[
+    "BINANCECONNECT",
+    "BANXA",
+    "BLOCKCHAINDOTCOM",
+    "BTCDIRECT",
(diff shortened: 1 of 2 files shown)


src/providers/meld.rs
```diff
@@ -263,6 +278,25 @@ impl OnRampMultiProvider for MeldProvider {
         url.query_pairs_mut()
             .append_pair("categories", DEFAULT_CATEGORY);
 
+        // Exclude provider from DEFAULT_PROVIDERS_LIST if exclude_providers is provided
+        // and the type is not `countries-defaults`
+        if params.r#type != PropertyType::CountriesDefaults {
+            let providers_list = if let Some(exclude_providers) = params.exclude_providers {
+                let exclude_set: std::collections::HashSet<&str> =
```
chris13524, line 285: Perhaps slightly more readable, instead of filtering, would be to subtract one set from the other via set operation

chris13524 (approved):
LGTM

src/providers/meld.rs
geekbrother: CoPilot will probably complain that this is not super efficient 😅

PR #1169, opened 2025-07-24
feat: test exchange

lukaisailovic:
# Description

Add test exchange so that customers can interact with AppKit Pay without the need to use Coinbase or Binance. Its available on Sepolia and Base Sepolia networks when native ETH is provided.

Resolves # (issue)

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/exchanges/test_exchange.rs
+++ b/src/handlers/wallet/exchanges/test_exchange.rs
@@ -1,0 +1,95 @@
+use {
+    crate::handlers::wallet::exchanges::{
+        ExchangeProvider, ExchangeError, GetBuyUrlParams, GetBuyStatusParams, GetBuyStatusResponse, BuyTransactionStatus,
+    },
+    crate::utils::crypto::Caip19Asset,
+    once_cell::sync::Lazy,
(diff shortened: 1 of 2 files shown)


src/handlers/wallet/exchanges/test_exchange.rs
```diff
@@ -0,0 +1,106 @@
...
+    std::sync::Arc,
+};
+
+pub struct TestExchange;
+
+const TEST_EXCHANGE_URL: &str = "https://appkit-pay-test-exchange.reown.com";
+
+static CAIP_19_SUPPORTED_ASSETS: Lazy<Vec<Caip19Asset>> = Lazy::new(|| {
+    vec![
+        Caip19Asset::parse("eip155:11155111/slip44:60").unwrap(),
```
geekbrother, line 20: `Expect` would be better here to catch errors in case of a panic.
```suggestion
Caip19Asset::parse("eip155:11155111/slip44:60").expect("Failed to parse CAIP-19 assets list"),
```

geekbrother (approved):
LGTM, Thanks! 🙏

PR #1182, opened 2025-08-06
feat: makes deployment rollouts smooth

geekbrother:
# Description

This PR adds the following changes to make deployment rollouts and ECS cluster size changes smooth:
* Graceful shutdown was implemented to not drop the current connections during the sigterm,
* `health_check_grace_period_seconds` is set to allow the service to start properly before routing the traffic to it,
*  `minimum_healthy_percent` for the ECS cluster is set to 50% and `maximum_percent` allows up to 200% the desired tasks during deployment.

## How Has This Been Tested?

* The graceful shutdown was tested locally,
* The cluster changes weren't tested and will be tested on the staging during the deployment.

## Due Diligence

* [ ] Breaking change


nopestack (approved):

src/lib.rs
```diff
@@ -465,28 +481,76 @@ pub async fn bootstrap(config: Config) -> RpcResult<()> {
...
+    tokio::select! {
+        result = futures_util::future::select_all(services) => {
+            if let Err(e) = result.0 {
+                warn!("Server error: {e:?}");
+            }
+        }
+        _ = shutdown_signal() => {
+            info!("Graceful shutdown initiated, allowing services to complete current work...");
+            // Give services a moment to finish current requests
+            tokio::time::sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
geekbrother: I believe it doesn't make sense here since if the waiting time was changed, something was changed in the code, so it will require rebuilding and probably we will not touch this constant.

terraform/ecs/cluster.tf
```diff
@@ -263,6 +264,14 @@ resource "aws_ecs_service" "app_service" {
   # Wait for the service deployment to succeed
   wait_for_steady_state = true
 
+  # Grace period for health checks during startup
+  # This prevents ELB health checks from failing during container initialization
+  health_check_grace_period_seconds = 60
+
+  # Deployment configuration for smooth rolling updates
```
chris13524, line 271: Why reduce to 50%? Default is 100%. Isn't it better to have all of the replicas we need to handle the scale?

terraform/ecs/cluster.tf
```diff
@@ -263,6 +264,14 @@ resource "aws_ecs_service" "app_service" {
...
+  deployment_minimum_healthy_percent = 50  # Keep at least 50% of tasks healthy during deployment
+  deployment_maximum_percent         = 200 # Allow up to 200% of desired tasks during deployment
```
chris13524, line 273: This is the default value: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DeploymentConfiguration.html#ECS-Type-DeploymentConfiguration-maximumPercent

PR #1211, opened 2025-08-28
feat: pos transaction builder

lukaisailovic:
# Description

- Add POS transaction RPC methods `reown_pos_buildTransaction` and `reown_pos_checkTransaction`
- Implement EVM transaction builder with support for native ETH and ERC20 transfers
- Add CAIP validation for asset, sender, and recipient addresses with proper namespace/chain ID matching
- Extensible `TransactionBuilder` trait for multi-chain support 
- Add transaction status checking. Implementation is querying transaction confirmation status via receipt lookup
- Added WalletBuildPosTx and WalletSendPosTx message sources

Resolves # (issue)

## How Has This Been Tested?
Manually
Current version of `alloy` that we use doesn't support mocked transport unfortunately. I'll be upgrading `alloy` across the implementation and implementing mocked transport testing in the follow up pr.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/wallet/pos/evm.rs
+++ b/src/handlers/wallet/pos/evm.rs
@@ -1,7 +1,7 @@
 use {
     super::{
         BuildPosTxError, BuildTransactionParams, BuildTransactionResult, TransactionBuilder,
-        TransactionRpc,
+        TransactionId, TransactionRpc, TransactionStatus,
     },
     crate::{
         analytics::MessageSource,
@@ -9,7 +9,7 @@ use {
         utils::crypto::{disassemble_caip10, Caip19Asset, Caip2ChainId, CaipNamespaces},
     },
     alloy::{
-        primitives::{utils::parse_units, Address, U256},
+        primitives::{utils::parse_units, Address, TxHash, U256},
         providers::{Provider, ProviderBuilder},
         rpc::types::TransactionRequest,
         sol,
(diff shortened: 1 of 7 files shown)


geekbrother:
@lukaisailovic can we do some basic integration test with pre-created items? Same as we do in [CA with pre-created wallets](https://github.com/reown-com/blockchain-api/blob/master/integration/chain_orchestrator.test.ts#L54).

src/handlers/wallet/pos/evm.rs
```diff
@@ -0,0 +1,334 @@
...
+        sol,
+    },
+    async_trait::async_trait,
+    axum::extract::State,
+    std::{convert::TryFrom, sync::Arc},
+    strum_macros::EnumString,
+    tracing::debug,
+};
+
+const NATIVE_GAS_LIMIT: u64 = 21_000;
```
geekbrother, line 24: Shouldn't this limit be provided for the transaction somehow by the client/pos? In case the gas limit is too low/high, but the client wants to pay for the gas more/less?

lukaisailovic: This is the maximum limit. It shouldn't be provided by the client. We forward this request to the wallet who can override the values if it needs to

lukaisailovic: Also this is the max known limit **for native transfers only**. Native transfer (no data) on EVM cannot exceed this limit.

PR #1249, opened 2025-09-24
feat(CORS): populate CORS headers for the `/json-rpc` exchanges endpoint

geekbrother:
# Description

This PR populates CORS headers for the `/json-rpc` exchanges endpoint.

Per-project CORS driven by the `projectId` query param and the project’s registry data. 
Preflight and actual responses now set CORS headers only if the request `Origin` matches the project’s `allowed_origins`.

### How it works:

* OPTIONS → json_rpc_preflight
* POST → json_rpc_with_dynamic_cors

Both handlers read `projectId` query param, fetch project data and:

* Allow if `Origin` matches any `allowed_origins` (case-insensitive exact match) or a hostname entry in allowed_origins.
* Preflight returns `204` with `Access-Control-Allow-*` headers when allowed, otherwise `403`.
* `POST` adds `Access-Control-Allow-Origin` and related headers to the normal response when allowed. Otherwise, it returns the normal response *without CORS headers* (browser will block).

## How Has This Been Tested?

Two integration tests were implemented: CORS header allow and deny (no header).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/json_rpc/handler.rs
+++ b/src/handlers/json_rpc/handler.rs
@@ -57,6 +57,164 @@ pub async fn handler(
         .await
 }
 
+// Wrapper that adds dynamic CORS headers based on project registry data
+pub async fn json_rpc_with_dynamic_cors(
+    state: State<Arc<AppState>>,
+    connect_info: ConnectInfo<SocketAddr>,
+    headers: HeaderMap,
+    query: Query<WalletQueryParams>,
+    SimpleRequestJson(request_payload): SimpleRequestJson<JsonRpcRequest>,
+) -> Response {
+    let mut response = handler(
+        state.clone(),
+        connect_info,
+        headers.clone(),
+        query.clone(),
+        SimpleRequestJson(request_payload),
+    )
+    .await;
+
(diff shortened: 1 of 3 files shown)


src/handlers/json_rpc/handler.rs
```diff
@@ -57,6 +57,170 @@ pub async fn handler(
...
+    if let Some(list) = get_project_allowed_origins(state.0.clone(), &query.project_id).await {
+        insert_allowed_origins_debug_header(&mut response, &list);
+    }
+
+    if let Some(origin) = headers
+        .get(hyper::header::ORIGIN)
+        .and_then(|v| v.to_str().ok())
+    {
+        if is_origin_allowed_for_project(state.0.clone(), &query.project_id, origin).await {
+            insert_cors_headers(&mut response, origin);
```
lukaisailovic, line 87: Should we also make sure to always include localhost? Otherwise people wont be able to develop, and you can't add it in Cloud. We do this for secure site

lukaisailovic (approved):
Tested locally, seems to be working fine

PR #1254, opened 2025-09-26
fix(CD): ignoring the flaky test

geekbrother:
# Description

This PR adds ignoring the flaky names count test until the fix. 
We are already cross-checking names in [separate integration tests](https://github.com/reown-com/blockchain-api/blob/master/integration/names.test.ts).

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/database.rs
+++ b/tests/functional/database.rs
@@ -348,6 +348,8 @@ async fn insert_delete_two_addresses() {
     assert!(delete_result.is_ok(), "Deleting name should succeed");
 }
 
+/// This test is ignored because because of the flakiness until the fix
+/// We are also testing names in `/integration/names.test.ts`
 #[tokio::test]
 #[ignore]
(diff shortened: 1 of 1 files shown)


tests/functional/database.rs
```diff
@@ -349,6 +349,7 @@ async fn insert_delete_two_addresses() {
 }
 
 #[tokio::test]
+#[ignore]
```
arein, line 352: @geekbrother maybe add a comment here that this is redundant is it's covered in https://github.com/reown-com/blockchain-api/blob/master/integration/names.test.ts

arein (approved):

tests/functional/database.rs
geekbrother: Makes sense, thank you! Added.

